PR #12, opened 2020-07-21
Add encrypted Souce Labs credentials

Vitalii-Misechko:
Added encrypted Souce Labs credentials according [these steps](https://github.com/BrightspaceUI/guide/wiki/Testing#cross-browser-testing-with-sauce-labs)

FYI @rohitvinnakota @hyehayes @johngwilkinson @MykolaGalian


.travis.yml
```diff
@@ -17,4 +17,4 @@ after_success:
 env:
   global:
   - SAUCE_USERNAME: Desire2Learn
-
+  - secure: gcOFFoC7sDNbaWPuEPHsMFz2G6BcVRe7vr/Qm8nvo+z6g0lSL9uu4nITvkmC35EMB7t3w5oXUw4wc7O6rbdnXstFEdD+D2kpTJKU/WTl35U91Tyj3HapbBsYQGNCZqrJlkdpW3PB0xR8pr2RKn9Y2K/Ha3AApESyiCQgcHigfBw+7p6sh3entm9x+5/IlscspAQ2xGhnIVhE8iq7exf72f10ctbYkY74hrtvEGAGZDKH28MMBDgPN0OK0eEdAFWIKtGtih1ny/vrMpiywjJjyhhEPq2GNkHU6v4N3pddMuXwjiRvR7u+VzHIo7LCZ0gfScw+/34u3BQkEVFHfG8MOE2/4csgHqAvsbIjcN+h9ZHIFw2gVLl/eLG1PVd5lRwtD0VwnAeX4E3QiCSwDg9FsIDpnm8xRFdQD7h5bBZcNs+gNq3dxjYvnI1iGQ0qeSwYyOYwY81/H4z4qZP/n1pv75Pk6IGHk1aiUT0fVbjQqh8fb6Jnlw/5EaW1rESmmMWVXxKm7hINfHgfQBDj0TQvWRtuX/GyHzIDQ1IpyA2/GG/YrweSwnkz+MCoLn0HxTOiVGC7fvbxK5GqtpKfmvauIFDfjscKADZ2eKGZhg8FpTE4lkYkprMBSF15F5XYoD8LZtx3mxH+An2PrPPsbhBc0eJ963UkfylNuW9ex6MMv5o=
```
anhill-D2L, line 20: Is it common practice to have creds stored in a github repo?

anhill-D2L: Looks like it is: https://search.d2l.dev/search?full=secure&defs=&refs=&path=.travis.yml&hist=&type=&xrd=&si=full&searchall=true&si=full

Vitalii-Misechko: Yeah....this secret value is useless outside this repo

bemailloux:
Do we have our own Sauce account for Draco? Or are we just using someone else's account?

Vitalii-Misechko:
> Do we have our own Sauce account for Draco? Or are we just using someone else's account?

No we don't have account for Draco, unfortunately. We were provided with credential to parent account (`Desire2Learn`).

.travis.yml
neverendingqs: You can add it via Travis CI's UI as well: https://travis-ci.com/github/Brightspace/insights-engagement-dashboard/settings

This makes it easier to manage because you don't need to use the CLI to do the encryption.

neverendingqs, line 20: It's common to leave a comment stating what the environment variable name is. That makes it easier to manage in the future (e.g. need to rotate the secret), although I would recommend adding it via the UI instead for that same reason.

Vitalii-Misechko: Added. Thanks. @neverendingqs, does this mean that this PR should be declined then?

Vitalii-Misechko:
Added `SAUCE_ACCESS_KEY` via travis UI

PR #15, opened 2020-07-27
Us117429/temp simple role filter

bemailloux:
Quality notes
* Testing
  * Unit tests for roles model class and insights-role-filter component
  * Manual testing
    * It loads the roles from the LMS using the `GET /d2l/api/lp/(version)/roles/` endpoint
    * It orders the roles by role name (and role id if multiple roles have the same name)
    * Every time a role is selected or unselected in the filter, it fires an event that the dashboard can listen for
* Security
  * Should only load roles that the dashboard user can manage (based on the notes in the valence docs)
* Devops
  * No error handling. 404 causes the filter to load with no entries inside
  * It uses a temporary `simple-filter` component that will eventually be replaced with [`facet-filter-sort`](https://github.com/BrightspaceUI/facet-filter-sort) once the latest version of that component is available in BSI
* Perf: 
  * Depends on valence API perf
  * Sorting roles by name may be an expensive operation. I haven't tested with lots of roles
* UX/Docs
  * Added a11y unit test, did not test a11y manually
  * No localization support yet

Screenshot
image

FYI @hyehayes @rohitvinnakota @anhill-D2L @MykolaGalian 

--------------------------

*Pre-review discussion*
Looking to have a discussion around these changes before I submit this for actual review. Here are the questions I'm hoping to answer:
* I made `rolesData` a property of `engagement-dashboard` so that whenever it was updated, it would re-render the roles filter. 
  * Is that the right approach? 
  * Or should I convert roles to a mobxLitElement so that triggering a re-render is handled by mobx? 
  * Or is there some other way of triggering a re-render that I'm not aware of?
* `roles.js` is similar to our I/O components in SSR, so it's tricky to test.
  * Does this component need testing? (I think so, looking for other opinions)
  * If it does, how should I go about that?
    * My thoughts: I think mocking the `fetch` API is not a great idea. I was thinking the karma test server could probably serve up test data, but I'm not sure how to configure it to do that.


components/simple-filter.js
```diff
@@ -0,0 +1,79 @@
...
+
+import '@brightspace-ui/core/components/dropdown/dropdown-content';
+import '@brightspace-ui/core/components/dropdown/dropdown';
+import '@brightspace-ui/core/components/inputs/input-checkbox';
+
+import {html, LitElement} from 'lit-element';
+import {selectStyles} from '@brightspace-ui/core/components/inputs/input-select-styles';
+
+/**
+ * @fires d2l-simple-filter-selected - event.detail contains {string} filterName, {string} id, and {boolean} selected
```
johngwilkinson, line 11: do you know anything about this `@` notation? It's new to me.

components/simple-filter.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+class SimpleFilter extends LitElement {
+    static get styles() {
+        return [selectStyles];
+    }
+
+    static get properties() {
+        return {
+            name: {type: String, attribute: true},
+            data: {type: Array, attribute: false}
```
johngwilkinson, line 21: what goes in data? (comment)

engagement-dashboard.js
```diff
@@ -68,12 +73,30 @@ class EngagementDashboard extends LocalizeMixin(LitElement) {
 		});
 
 		this.histogram = new Histogram({id: 'engagement-demo-chart', field: 'n', title: 'distribution of n'}, this._data);
+
+		this.roleData = [];
```
johngwilkinson, line 77: Instead of putting roleData on the dashboard, I'd like to see it encapsulated in a component that passes it into your simple-filter component. Let's say it's called `d2l-insights-role-filter`; this class would render `<d2l-insights-role-filter></d2l-insights-role-filter>`. Then a the `RoleFilter` class would call the API and render the `<d2l-simple-filter name="Roles" .data="${this.roleData}"></d2l-simple-filter>`. That way `d2l-insights-role-filter` is can be dropped into any dashboard.

components/simple-filter.js
```diff
@@ -0,0 +1,79 @@
...
+        this.name = '';
+
+        this.addEventListener('d2l-simple-filter-selected-internal', this._handleElementSelected);
+    }
+
+    _handleElementSelected(event) {
+        event.stopPropagation();
+        this.dispatchEvent(new CustomEvent('d2l-simple-filter-selected', {
+            bubbles: true,
+            composed: true,
```
johngwilkinson, line 38: https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-beware-composed-events

Not saying this is wrong, but is it needed, and why?

components/simple-filter.js
```diff
@@ -0,0 +1,79 @@
...
+        };
+    }
+
+    constructor() {
+        super();
+
+        this.data = [];
+        this.name = '';
+
+        this.addEventListener('d2l-simple-filter-selected-internal', this._handleElementSelected);
```
johngwilkinson, line 31: Probably not the right place to hook this up:  see https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-beware-composed-events

engagement-dashboard.js
johngwilkinson: Another possibility, which is perhaps more html-y: RoleFilter renders something like 
```
<d2l-insights-simple-filter>
  ${roles.map(r => html`<d2l-insights-simple-filter-item name=${r.name}>`)}
</d2l-insights-simple-filter>
```
Where simple filter has a slot for putting role items.

I'm right now trying to figure out whether it's better for the tree filter to have the base component be a simple thing with slots, or a thing that takes a nested object representing the tree and draws the whole thing.

Right now I'm leaning toward "don't pass objects when an html structure could carry the same info".

johngwilkinson: Hmm (from the lit-element docs):
> Also, children can be styled from outside the shadow tree, so you should regard your ::slotted() styles as default styles that can be overridden.

This makes me think slots are not the right construct for rendering complex objects, and passing in "options" so the sub-component can do the rendering is the way to go.

components/simple-filter.js
```diff
@@ -0,0 +1,79 @@
+// just a list of checkboxes inside a dropdown to act as a temporary basic filter
+
+import '@brightspace-ui/core/components/dropdown/dropdown-content';
+import '@brightspace-ui/core/components/dropdown/dropdown';
+import '@brightspace-ui/core/components/inputs/input-checkbox';
```
johngwilkinson, line 5: I'm likely going to use this element. I don't see anything in your package-json for it. Isn't something needed?

components/simple-filter.js
bemailloux: Thanks, "repackaging" the events felt really wrong but I didn't know you could just listen for events in the template like that

components/simple-filter.js
bemailloux: Everything is part of core, so no extra imports needed

engagement-dashboard.js
bemailloux: This makes sense, will do

components/simple-filter.js
bemailloux: It's JSDoc notation, but I don't think I'm technically using it correctly: https://jsdoc.app/tags-fires.html

Copied from here: https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-checkbox.js

bemailloux:
@johngwilkinson any thoughts on the questions in the PR description?

johngwilkinson:
> @johngwilkinson any thoughts on the ~questions~ other question about testing in the PR description? (I think the property one is handled by creating that `d2l-insights-role-filter` element)

I don't really know, but perhaps with a little refactoring you can unit test things. There's also the visual diff stuff which should help for top-level testing (along with some canned data - which might necessitate adding a property to the dashboard element to pass in data?)

model/roles.js
```diff
@@ -0,0 +1,42 @@
+class Roles {
```
bemailloux, line 1: Missed out on a chance to call this class `RoleModel` :)

components/simple-filter.js
bemailloux: Wow, using the `@` notation even binds `this` to the component, instead of to the target or something else

bemailloux:
@johngwilkinson I added a `d2l-insights-role-filter` element and dependency-injected the Roles model into it. Not sure if DI-ing the roles via the role-filter's properties was the correct way to do it though

johngwilkinson:
> @johngwilkinson I added a `d2l-insights-role-filter` element and dependency-injected the Roles model into it. Not sure if DI-ing the roles via the role-filter's properties was the correct way to do it though

I'd suggest the Roles object should not be referenced outside of the role-filter component, which should create it itself.

bemailloux:
> the Roles object should not be referenced outside of the role-filter component,

Doesn't the top level at least need a reference to (or maybe its own copy of) the list of selected role ids, so it can apply the filter?

bemailloux:
> which should create it itself.

Also, having `d2l-insights-role-filter` have a direct dependency on `Roles` makes it harder to test. Maybe that's ok though, if we want that component and the data model to be linked together, as long as `Roles` can be correctly set up for a test environment

johngwilkinson:
> > the Roles object should not be referenced outside of the role-filter component,
> 
> Doesn't the top level at least need a reference to (or maybe its own copy of) the list of selected role ids, so it can apply the filter?

I would expect it would listen for a change event and simply call a `.getSelected()` method (or even just read a `.selected` property/getter) on the role filter when it gets one. I'm sure there are other ways, but ideally adding a role filter is as easy as
```
import 'role-filter.js'
...
render() { html`<d2l-insights-role-filter @change="${this._onChange}"></d2l-insights-role-filter>`; }

_onChange(e) {
  this.data.selectedRoles = e.target.selected;
}
```

bemailloux:
> this.data.selectedRoles = e.target.selected;

Riiiiiight I forgot that the target was included in the event. Thanks

bemailloux:
@johngwilkinson @Vitalii-Misechko this is now ready for review

johngwilkinson:
>@johngwilkinson @Vitalii-Misechko this is now ready for review

We should really bring in some UI experts to review this. I'd suggest the Daves, but it's possible Andrew has other names that the Gaudi manager would like us to use.

bemailloux:
> We should really bring in some UI experts to review this.

Hey @svanherk, here's one of the places where the `facet-filter-sort` component will be used. Would you be willing to review this PR as well (or volunteer someone else from Gaudi)?

svanherk:
> Hey @svanherk, here's one of the places where the facet-filter-sort component will be used. Would you be willing to review this PR as well (or volunteer someone else from Gaudi)?

Sure I can take a look!  The other person I would have volunteered is Dave B who's on vacation this week :)

model/roles.js
```diff
@@ -0,0 +1,44 @@
+const rolesEndpoint = '/d2l/api/lp/1.23/roles/';
+
+class Roles {
+	async fetchRolesFromLms() {
+		const response = await fetch(rolesEndpoint);
```
johngwilkinson, line 5: I'm loving this API.

test/components/insights-role-filter.test.js
```diff
@@ -0,0 +1,106 @@
+import '../../components/insights-role-filter';
+
+import {expect, fixture, html, oneEvent} from '@open-wc/testing';
+import fetchMock from 'fetch-mock/esm/client';
+import {rolesEndpoint} from '../../model/roles';
```
johngwilkinson, line 5: This may seem like DRY, but it means that if I accidentally break the roles endpoint in the shipping code, this test will still pass - so I usually duplicate this sort of constant in my tests.

components/insights-role-filter.js
```diff
@@ -0,0 +1,48 @@
...
+import Roles from '../model/roles';
+
+/**
+ * @fires role-selections-updated
+ */
+class InsightsRoleFilter extends LitElement {
+
+	static get properties() {
+		return {
+			roleData: {type: Array, attribute: false}
```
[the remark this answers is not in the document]
bemailloux: This has to be a prop for the element to re-render when the roles are loaded from the LMS. I imagine there's a way to make the element directly track a property of the roles model instead, but I wasn't sure how to do that.

test/engagement-dashboard.test.js
```diff
@@ -5,7 +5,9 @@ import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-help
 describe('d2l-insights-engagement-dashboard', () => {
 
 	describe('accessibility', () => {
-		it('should pass all axe tests', async() => {
+		it('should pass all axe tests', async function() {
+			this.timeout(3000);
+
 			const el = await fixture(html`<d2l-insights-engagement-dashboard></d2l-insights-engagement-dashboard>`);
```
johngwilkinson, line 11: Why doesn't this need the roles API mocked, too? I'm guessing the role-filter just fails to load? (aside: I'll likely grab the fetch-mock approach to mock the main query for this test once I get it hooked up to the ou filter)

components/insights-role-filter.js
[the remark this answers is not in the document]
johngwilkinson: mobx would do that, but it seems overkill here.

hmmm... does it need to re-render? Maybe when the roles are loaded, you just update the property on the simple-filter. I _think_ that's legit.

components/insights-role-filter.js
```diff
@@ -0,0 +1,48 @@
...
+			<d2l-simple-filter
+				@item-selected="${this._updateFilterSelections}"
+				name="${this.name}"
+				.data="${this.roleData}">
+			</d2l-simple-filter>
+		`;
+	}
+
+	_updateFilterSelections(event) {
+		this.roles.setSelectedState(event.detail.itemId, event.detail.selected);
```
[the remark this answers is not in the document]
bemailloux: The problem I ran into with implementing #2 was visibility: how do some elements "disappear" when they're filtered out (for example when someone applies a search parameter)? I think the solution to that might be "add/remove a CSS style that toggles visibility on an element" but I didn't play around enough with that idea to get it working. Now that I'm a bit more familiar with how this all works though, it might be worth trying.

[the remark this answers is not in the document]
johngwilkinson: Ah, yes, I ran into that with the ou filter too (when nodes are collapsed). I needed to hide rather than remove parts of my tree so that state persisted in the dom (since I was aiming for approach 2).

Overall, I don't feel too strongly. I suspect relying on the dom as the source of truth will be simpler in the long run for fairly self-contained components like this, but I don't know for sure.

components/insights-role-filter.js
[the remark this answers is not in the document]
bemailloux: I could try that, but I think then I'd have to handle a possible race between the element first rendering (can't get the simple filter if it hasn't rendered for the first time yet) and the roles getting loaded from the LMS. Both of those things need to be completed before I can update the property. 

Initial idea of how that might look:
```
constructor() {
        ...
	this.roles.fetchRolesFromLms().then(() => {
                if (this.shadowRoot.querySelector('d2l-simple-filter'))     // then attach the data to the simple filter
                else this.roleData = this.roles.getRoleDataForFilter();      // store it so we can use it when rendering is done
	});
}

firstUpdated() {
        if (this.roleData) {
               this.shadowRoot.querySelector('d2l-simple-filter').data = this.roleData;
        }
}
```
Seems like more work than just having it be a prop on the insights-role-filter

components/simple-filter.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @fires item-selected - event.detail contains {string} filterName, {string} id, and {boolean} selected
+ */
+class SimpleFilter extends LitElement {
+
+	static get properties() {
+		return {
+			name: {type: String, attribute: true},
+			data: {type: Array, attribute: false}
```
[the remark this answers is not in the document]
bemailloux: This syntax allows for type hinting in webstorm (and probably intelliJ):
```
/**
 * @property {string} name
 * @property {{id: string, displayName: string}[]} data
 * @fires item-selected - event.detail contains {string} filterName, {string} id, and {boolean} selected
 */
class SimpleFilter extends LitElement {
```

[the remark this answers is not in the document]
johngwilkinson: Oh cool!

engagement-dashboard.js
```diff
@@ -70,10 +73,19 @@ class EngagementDashboard extends LocalizeMixin(LitElement) {
 		this.histogram = new Histogram({id: 'engagement-demo-chart', field: 'n', title: 'distribution of n'}, this._data);
 	}
 
+	_handleRoleSelectionsUpdated(event) {
+		event.stopPropagation();
```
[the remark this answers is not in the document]
bemailloux: Just a habit - stops elements higher up from handling events, which comes in handy when you're calling your events all the same thing like "change", "selected", etc. Not really necessary here

test/engagement-dashboard.test.js
bemailloux: This is what's rendered when I run es-dev-server, and the roles endpoint returns 404:
image

I expect the same thing is happening in the dashboard test, so it doesn't have any issues passing constructor and accessibility tests.

components/insights-role-filter.js
```diff
@@ -0,0 +1,48 @@
...
+		});
+	}
+
+	get selected() {
+		return this.roles.getSelectedRoleIds();
+	}
+
+	render() {
+		return html`
+			<d2l-simple-filter
```
svanherk, line 35: Will this eventually use the `d2l-filter-dropdown-*` components directly and `d2l-simple-filter` can be deleted?

components/insights-role-filter.js
```diff
@@ -0,0 +1,48 @@
...
+				@item-selected="${this._updateFilterSelections}"
+				name="${this.name}"
+				.data="${this.roleData}">
+			</d2l-simple-filter>
+		`;
+	}
+
+	_updateFilterSelections(event) {
+		this.roles.setSelectedState(event.detail.itemId, event.detail.selected);
+		this.dispatchEvent(new Event('role-selections-updated'));
```
svanherk, line 45: We usually namespace events with the component name to avoid collisions, and to make it easier to see where the event came from. Something like `d2l-insights-role-filter-selections-updated`

components/insights-role-filter.js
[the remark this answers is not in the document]
svanherk: If this property is only used internally by this component (and shouldn't be set by consumers of this component), we mark it as private by adding an underscore to the name (`_roleData`).  This doesn't actually make it private to the consumer, but it's widely accepted and helps people determine the public interface for your component if the docs are lacking or out of date.

components/insights-role-filter.js
```diff
@@ -0,0 +1,48 @@
...
+	static get properties() {
+		return {
+			roleData: {type: Array, attribute: false}
+		};
+	}
+
+	constructor() {
+		super();
+
+		this.name = 'Roles';
```
svanherk, line 20: Seems like everything here is internal/private?

components/insights-role-filter.js
bemailloux: Yup, that's the plan

components/insights-role-filter.js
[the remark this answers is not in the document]
svanherk: I find #1 is more common because with Lit, `render` should be a function of your properties.  You don't want to manipulate the DOM directly outside of the `render` function, and I should be able to look at the state of all the properties (in this case `_roleData`) and know exactly what the DOM will look like.

However, I do find this setup a little confusing - it wasn't clear to me that this function was going to manipulate `_roleData`.  Do you expect to use `roles.js` somewhere else?  It feels like it should just be pulled into this file, or be responsible for returning the initial data but then this file does the manipulations?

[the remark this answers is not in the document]
bemailloux: > Do you expect to use roles.js somewhere else?

Not planning to use `roles.js` anywhere else as far as I know. 

> or be responsible for returning the initial data but then this file does the manipulations?

This makes sense, I guess the element itself should be responsible for what happens when an item is selected, instead of an arbitrary logic class

bemailloux:
Made some changes:
- The `d2l-insights-role-filter` now handles its own selection logic, rather than deferring to the `Roles` class
- `Roles` class is now renamed to `Lms`, since all it does is fetch things now
- Applied "private" naming convention

bemailloux:
Additional testing note: verified working in chrome, ff, edgium, edge legacy

components/insights-role-filter.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	constructor() {
+		super();
+
+		this._name = 'Roles';
+		this._roleData = [];
+
+		const lms = new Lms();
+		lms.fetchRoles().then(this._setRoleData.bind(this));
```
svanherk, line 25: Maybe consider moving this to `connectedCallback` or something that happens a bit later than the `constructor` - technically the constructor could run but then the element is never actually appended to the DOM.  If you want to make sure this only runs once, you can check if `_roleData ` is empty or not

components/insights-role-filter.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	get selected() {
+		return this._roleData.filter(role => role.selected).map(role => role.Identifier);
```
svanherk, line 29: Any reason this is exposed instead of just sending the result along in the `d2l-insights-role-filter-change` event's `detail`?  Then you can avoid looking at the event's `target` and worrying about event retargeting with shadow DOM if things start getting nested

components/simple-filter.js
```diff
@@ -0,0 +1,64 @@
...
+					`)}
+				</d2l-dropdown-content>
+
+			</d2l-dropdown>
+		`;
+	}
+
+	_handleElementSelected(event) {
+		// propagate the event one level up, since it can't cross the shadow DOM boundary
+		this.dispatchEvent(new CustomEvent('item-selected', {
```
svanherk, line 55: Minor: I know this whole file is temporary, but consider namespacing this event as well if it's going to stick around for a little while

package.json
```diff
@@ -34,13 +34,14 @@
...
     "karma-sauce-launcher": "^4",
-    "lit-analyzer": "^1",
+    "lit-analyzer": "^1.2.0",
     "puppeteer": "^5"
   },
   "dependencies": {
     "@adobe/lit-mobx": "0.0.4",
     "@brightspace-ui/core": "^1.55.1",
     "@webcomponents/webcomponentsjs": "^2",
+    "fetch-mock": "^9.10.5",
```
svanherk, line 44: Can this be moved to `devDependencies`?  That way BSI doesn't need to install it

components/insights-role-filter.js
bemailloux: Not sure I follow - you're saying that Lit might try multiple times to construct this component but fail to append it to the DOM on the first try? Which means we would have requested data from the LMS multiple times instead of just once?

components/insights-role-filter.js
bemailloux: I had a concern about perf when recalculating the `selected` list every time the event was fired. That is what we're doing right now (and it doesn't really matter for roles since the roles list is pretty small), but for other components like the upcoming OU tree filter, where recalculating the selected list is expensive, we'll probably only want to do it when we're ready to apply the filters.

I guess what we _really_ want is for the element to return the `selected` list when the filter is closed.

johngwilkinson: Keep in mind the ou filter won't be using this code anyway. For your perf concern, are you worried about the cost of reading the state from the DOM, or the cost of making an API call based on the new values? My impression is that we are to update the data whenever the filter changes, not just when it is closed - but I like the when-closed idea. Maybe run that by Kevin?

If we are not waiting for close, I think we'll want to debounce the API call. Not sure if reading a few values (or a few hundred) from the DOM is an issue, but I could be wrong.

components/insights-role-filter.js
johngwilkinson: minor: since es6 I generally prefer arrow functions to `.bind` (`lms.fetchRoles().then(data => this._setRoleData(data))`)

components/insights-role-filter.js
bemailloux: > For your perf concern, are you worried about the cost of reading the state from the DOM, or the cost of making an API call based on the new values?

I had been worried about reading the state from the DOM every time (not so much for roles but for maybe OU or semester filters), but now that you mention it trying to hit the API every single time something is selected is not a great thing to do either. I thought I heard somewhere that we only wanted to apply filters when a filter was closed, but I'll check with Kevin.

johngwilkinson:
Just noticed: I don't see any unit tests for simple-filter.

bemailloux:
>Just noticed: I don't see any unit tests for simple-filter.

That's intentional, since it's only a temporary component that we'll throw away later. I could add them in if you think they're valuable

components/insights-role-filter.js
bemailloux: Oh yeah, Kevin's not here. Should we proceed with the assumption that every selection causes filters to be applied, and see how bad the perf gets? Seems like a relatively easy change to go to the "apply on close" approach

johngwilkinson:
> > Just noticed: I don't see any unit tests for simple-filter.
> 
> That's intentional, since it's only a temporary component that we'll throw away later. I could add them in if you think they're valuable

Temporary solutions last the longest.

components/insights-role-filter.js
johngwilkinson: I'd say pick whichever one for now - if apply on close is easy, then maybe start there, since it's safer - and let Megan know.

components/insights-role-filter.js
svanherk: No, I kinda threw two comments into one - we tend to delay work if we can to not hold up first paint.  If you didn't want it in `connectedCallback` _because_ that can be called multiple times, then you could make sure to only run it if you haven't already.

@dbatiste When you're back, do you have thoughts here?

components/insights-role-filter.js
svanherk: So I'm currently making a lot of changes to this file, but you can use the `d2l-hm-filter` for inspiration... it currently supports both.  Depending on whether you want changes applied immediately, or sent in a batch when the filter closes, it will listen to a different event from `d2l-filter-dropdown`: https://github.com/BrightspaceHypermediaComponents/common/blob/master/components/d2l-hm-filter/d2l-hm-filter.js#L87-L91

bemailloux: Regarding this comment: 

> Then you can avoid looking at the event's target and worrying about event retargeting with shadow DOM if things start getting nested

If this element is used inside a lit element, wouldn't any events have to be retargeted anyway, assuming I use `composed: false` in a CustomEvent? Since events can't cross the shadow DOM boundary?

svanherk: Ah good point, since this event isn't `composed` you won't be able to catch it higher up when the event has been retargeted anyways.  I think we still tend to use the event's `detail` to avoid this if it ever changed in the future, but it's not a problem with how you have it currently.  Discouraging composed and bubbling events at D2L is fairly new, so you'll see them usually set to true all over the place

bemailloux: Will keep as is for now, to give consumers better control over when the selected list is calculated. Might change it later when we have a better idea of what those consumers will look like

bemailloux:
Hoping to merge today. Are there any remaining concerns other than moving data loading to `connectedCallback` (which I'll probably do in a separate PR)?

test/components/insights-role-filter.test.js
```diff
@@ -0,0 +1,130 @@
...
+			fetchMock.get(rolesEndpoint, mockLmsResponseData);
+
+			const expectedFilterData = [
+				{ id: '1', displayName: 'role' },
+				{ id: '2', displayName: 'Role' },
+				{ id: '3', displayName: 'Role' },
+				{ id: '4', displayName: 'ZZZRole' }
+			];
+			const el = await fixture(html`<d2l-insights-role-filter></d2l-insights-role-filter>`);
+			await new Promise(resolve => setTimeout(resolve, 500));
```
johngwilkinson, line 84: Why the long pause?

bemailloux: Required for the element for finish rendering, for some reason. I tried using `await el.updateComplete` but my tests failed with that for some reason.

johngwilkinson: Because the first render happens before the mock fetch completes, I suppose. But a half-second pause here and elsewhere seems like a lot. I wonder if a 1ms wait (to give up the thread to fetch) followed by `await el.updateComplete` would be reliable.

bemailloux: Good idea, I'll try a couple of combinations to try to cut this waiting time down

bemailloux: Uh oh! Ran into something kind of unexpected with `fetchMock`: if a previous test uses fetchMock, and doesn't reset fetchMock after, it'll apply its behaviour to any subsequent tests.

So for example `engagement-dashboard.js` was loading with some role data defined in `insights-role-filter.js`, simply because `insights-role-filter.js` ran first.

My thoughts on what we can do about this:
- Best practice: clean up your fetchMock at the end of a test suite if you're using it (easiest path but we'll probably forget and break the tests). I'll probably go with this for the time being
- Move away from fetchMock

Anyone have any other thoughts?

bemailloux: > I wonder if a 1ms wait (to give up the thread to fetch)

By the way, `setTimeout(0)` seems to be used in the wild for exactly this reason

bemailloux:
John reminded me that localization is a thing. I'll handle that in a separate PR (along with screen reader testing)

svanherk:
I'm dealing with a lot of certfix scrambling today, but I don't think there was anything else from me anyways!

bemailloux:
> I'm dealing with a lot of certfix scrambling today, but I don't think there was anything else from me anyways!

Thanks for taking the time to review!

PR #19, opened 2020-07-31
Us117429/l10n and a11y

bemailloux:
Sets up serge to translate our repo, and adds localization for the role filter. (More localization will be required once we shift to `facet-filter-sort`.) 

This is basically a copy of how BrightspaceUI/core does localization.

Quality notes
* Testing
  * Localization
    * English renders in English
    * Other languages
      * Setup: I created a temporary `fr.js` file (which I later deleted) and put a French langterm in there. Then I set my local LMS to French
      * If `fr.js` exists and contains the langterm, then it uses the term in `fr.js`
      * If `fr.js` doesn't exist, then it defaults to English
      * ~~If `fr.js` exists but doesn't contain the langterm, then it defaults to empty string. I think this case could happen if we add a new langterm but the file isn't updated until _after_ cert, so (TODO) we want to handle this case in a smarter way.~~
      * **UPDATE** If `fr.js` exists but doesn't contain the langterm, it defaults to English
  * [x] screen reader testing (leftover from previous PR)
* Security: N/A
* Perf:
  * We always load in the English langterms file, even if the LMS is in a different language
* Devops
  * [x] add this repo to https://git.dev.d2l/projects/RE/repos/serge-localize
    * PR here: https://git.dev.d2l/projects/RE/repos/serge-localize/pull-requests/132/diff
* UX/Docs: N/A

FYI @anhill-D2L @rohitvinnakota @hyehayes @MykolaGalian

### DIFF
--- a/components/insights-role-filter.js
+++ b/components/insights-role-filter.js
@@ -2,12 +2,13 @@ import './simple-filter';
 
 import {html, LitElement} from 'lit-element';
 import Lms from '../model/lms';
+import {Localizer} from '../locales/localizer';
 
 /**
  * @property {{id: string, displayName: string}[]} _filterData
  * @fires d2l-insights-role-filter-change - detail includes the list of selected role ids
  */
-class InsightsRoleFilter extends LitElement {
+class InsightsRoleFilter extends Localizer(LitElement) {
 
 	static get properties() {
 		return {
@@ -18,7 +19,6 @@ class InsightsRoleFilter extends LitElement {
 	constructor() {
 		super();
 
-		this._name = 'Roles';
 		this._roleData = [];
 
 		const lms = new Lms();
@@ -60,7 +60,7 @@ class InsightsRoleFilter extends LitElement {
 		return html`
 			<d2l-simple-filter
 				@d2l-simple-filter-selected="${this._updateFilterSelections}"
-				name="${this._name}"
+				name="${this.localize('components.insights-role-filter.name')}"
 				.data="${this._filterData}">
 			</d2l-simple-filter>
 		`;
--- a/components/insights-role-filter.js
+++ b/components/insights-role-filter.js
@@ -5,9 +5,9 @@ import Lms from '../model/lms';
 import {Localizer} from '../locales/localizer';
 
 const demoData = [
-	{ id: '500', displayName: 'Administrator' },
-	{ id: '600', displayName: 'Instructor' },
-	{ id: '700', displayName: 'Student' }
+	{ Identifier: '500', DisplayName: 'Administrator' },
+	{ Identifier: '600', DisplayName: 'Instructor' },
+	{ Identifier: '700', DisplayName: 'Student' }
 ];
 
 /**
@@ -26,7 +26,9 @@ class InsightsRoleFilter extends Localizer(LitElement) {
 	constructor() {
 		super();
 
+		this.isDemo = false;
 		this._roleData = [];
+		this._filterData = [];
 
 		const lms = new Lms();
 		lms.fetchRoles().then(data => this._setRoleData(data));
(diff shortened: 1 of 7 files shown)


locales/localizer.js
```diff
@@ -0,0 +1,70 @@
+// copied from BrightspaceUI/core's LocalizeCoreElement, with modifications
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin';
+
+export const Localizer = superclass => class extends LocalizeMixin(superclass) {
+
+	static async getLocalizeResources(langs) {
+		let translations;
+		try {
+			for await (const lang of langs) {
+				switch (lang) {
```
bemailloux, line 10: `switch/case` and explicitly listing the file paths is required for BSI's rollup to actually load the files in. Trying to use a variable-based dynamic import doesn't work. See https://d2l.slack.com/archives/C0PHG3QB0/p1596050453348900

components/insights-role-filter.js
```diff
@@ -18,7 +19,6 @@ class InsightsRoleFilter extends LitElement {
 	constructor() {
 		super();
 
-		this._name = 'Roles';
```
bemailloux, line 21: I tried to set the localized name here, but that failed, because the localized resources aren't loaded by the time the constructor runs

johngwilkinson:
For testing, you tested using an LMS and local BSI?

bemailloux:
> For testing, you tested using an LMS and local BSI?

Yes.

test/locales/localizer.test.js
```diff
@@ -0,0 +1,15 @@
+import {expect} from '@open-wc/testing';
+import {Localizer} from '../../locales/localizer';
+
+class MockLocalizedObjectSuperclass {}
+class MockLocalizedObject extends Localizer(MockLocalizedObjectSuperclass) {}
```
bemailloux, line 5: Hacky way to test a mixin. Happy to change this if anyone knows of a simpler way

johngwilkinson:
> > For testing, you tested using an LMS and local BSI?
> 
> Yes.

Cool, thanks. Once you've merged the latest from master (sorry, I created a couple of conflicts for you), can you make sure that the demo page (`npm start`) still works, too? I find it super useful for developing, and it was broken when I merged the initial role-filter stuff into my branch - I added demo data to fix.

bemailloux:
> can you make sure that the demo page (npm start) still works, too? I find it super useful for developing, and it was broken when I merged the initial role-filter stuff into my branch - I added demo data to fix.

Sure thing, I also noticed you added demo data to the role filter, thanks

bemailloux:
Conflict resolution should work, but I need to run a full_all to know for sure (I don't have the engagement dashboard backend yet)

bemailloux:
A few notes on screen reader testing
* Everything in the filter is keyboard accessible
* On focus, the facet filter sort says "Open roles filter menu button collapsed submenu". "Open roles filter" is the element's aria-label, not sure if the rest is expected or not
* Other elements' behaviour is pretty good, checkboxes read the role name and checked status when focused, and say "checked" or "unchecked" when an action is taken
* Pressing "Esc" will close the filter, but if the previous action was checking one of the boxes, then it doesn't say the filter was closed.

anhill-D2L:
> A few notes on screen reader testing
> ...

Do we have guidelines or a checklist of what to look for/check when doing accessibility testing?

bemailloux:
> > A few notes on screen reader testing
> > ...
> 
> Do we have guidelines or a checklist of what to look for/check when doing accessibility testing?

This is the main static a11y resource I'm aware of (which, admittedly, I didn't look at before doing my testing): https://docs.dev.d2l/index.php/Making_Your_Web_Page_Accessible_-_Top_12

Other than that I don't know if there are any like "checklist" type docs

johngwilkinson:
Umm, we shouldn't be saying "NA" for the "UX" part of our quality notes when doing front-end work. :) Probably the note to expect there is "approved by Kevin".

bemailloux:
> Umm, we shouldn't be saying "NA" for the "UX" part of our quality notes when doing front-end work. :) Probably the note to expect there is "approved by Kevin".

Oops, I forgot to check with either Design or Product that defaulting to English was what we wanted :/

PR #21, opened 2020-08-06
Us117440/users table

bemailloux:
Adds the table with the list of users who are currently in view.

NB: This PR does _not_ cover:
- actually filtering the users (US118248)
- sorting options that are different from the default (ASC by lastFirstName) (not in a story AFAIK)
- paging (US118254)

Quality notes
* Testing
  * Test cases
    * [x] It shows all users who are in the current view
    * [x] It displays the number of users in view
    * [x] Responsive
  * Browsers
    * [x] Chrome
    * [x] FF
    * [x] Edgium
    * [x] Edge Legacy
  * Unit tests added
* Security: N/A
* Perf: 
  * I think mobx is smart about when it recalculates the list of users (that is, it doesn't do it all the time). It re-sorts the list every time it has to recalculate, so recalculating too often will be slow.
* Devops: N/A
* UX/Docs
  * Localization: The table handles right-to-left (RTL) languages correctly

FYI @johngwilkinson @MykolaGalian @rohitvinnakota @hyehayes


bemailloux:
### Screenshots

####Desktop
image

...after scrolling down...

image


#### Responsive
image

...after scrolling down...

image

#### RTL
image

anhill-D2L:
> ### Screenshots
> ...

It's nice that these look just like what you see when you're searching for users in the LMS

bemailloux:
Screen reader and keyboard nav testing notes (Windows / Chrome Desktop / NVDA)
* Added an accessible title to the table
* Fully navigable using the arrow keys. Using Tab does not focus on the table because it is not an input element
* transcript: "Table with 359 rows and 1 columns. User Details. *down arrow* Row 1, Column 1. Last name, first name. *down arrow* Row 2. Aadland, Beverly. *etc...* End of table. *down arrow* Total users: 359" 
* Getting to the bottom of the table is currently a pain with lots of users. The paging story should fix that.

bemailloux:
Last thing: for some reason on Edge Legacy the border radius property doesn't work. No idea why
image

components/users-table.js
```diff
@@ -0,0 +1,69 @@
+import './table.js';
+
+import {css, html} from 'lit-element';
+import {Localizer} from '../locales/localizer';
+import {MobxLitElement} from '@adobe/lit-mobx';
+
+/**
+ * At the moment the mobx data object is doing all the display logic,
```
Vitalii-Misechko, line 8: I'm curious how is `data` connected with `mobx`?

Vitalii-Misechko: Oh...I see `model/data.js`. I need to look at it closer.

styles/table.style.js
```diff
@@ -0,0 +1,35 @@
+import {css} from 'lit-element';
+
+// shamelessly copied with modifications from the BrightspaceUI/table repo (since that isn't compatible with LitElement)
+export const tableStyle = css`
```
Vitalii-Misechko, line 4: What is a reason to separate styles and `components/users-table.js`?

bemailloux: It's just organization. Originally the styles I copied over from BrightspaceUI/table were very large, so putting them in the main table.js file would have been very cluttered. After I found out I could delete a lot of the unused styles, I just never moved them back.

PR #22, opened 2020-08-06
US117427 [Engagement][Filters] Semester

Vitalii-Misechko:
[US117427](https://rally1.rallydev.com/#/detail/userstory/396954104020?fdp=true): [Engagement][Filters] Semester

Back-end API PR: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/14230/overview

FYI @johngwilkinson 

### Functional Testing
  * Test cases - [WIP]
    * [x] It shows sorted semesters by DateCreated descending and OrgUnitId descending for each test case bellow
    * [x] It shows the first page of semesters list 
    * [x] It loads the second page of semesters list after clicking `Load More` button
    * [x] It shows filtered first page of semesters list if search value is specified
    * [x] It shows filtered second page if search value is specified after clicking `Load More` button
    * [x] It shows the first page of semesters list if filter is cleared
    * [x] ~Responsive~
  * Browsers
    * [x] Chrome
    * [x] FF
    * [x] Edgium
    * [x] ~Edge Legacy~ Skip ( constantly hanging on my laptop 😞  )
  * Unit tests added
### UX and docs

image

image

### DIFF
--- a/components/semester-filter.js
+++ b/components/semester-filter.js
@@ -1,60 +1,8 @@
 import './simple-filter';
 
 import {html, LitElement} from 'lit-element';
+import FakeLms from '../model/fake-lms';
 import Lms from '../model/lms';
-
-async function testData(bookmark, search) {
-	let response = {
-		PagingInfo: {
-			Bookmark: '1326467654053_120127',
-			HasMoreItems: true
-		},
-		Items: [
-			{
-				orgUnitId: 10007,
-				orgUnitName: 'IPSIS Semester New'
-			},
-			{
-				orgUnitId: 121194,
-				orgUnitName: 'Fall Test Semester'
-			},
-			{
-				orgUnitId: 120127,
-				orgUnitName: 'IPSIS Test Semester 1'
-			}
-		]
-	};
-	const secondPage = {
-		PagingInfo: {
-			Bookmark: '1326467625687_120124',
-			HasMoreItems: false
-		},
-		Items: [
-			{
-				orgUnitId: 120126,
-				orgUnitName: 'IPSIS Test Semester 12'
-			},
-			{
-				orgUnitId: 120125,
-				orgUnitName: 'IPSIS Test Semester 123'
-			},
-			{
-				orgUnitId: 120124,
-				orgUnitName: 'IPSIS Test Semester 4'
-			}
-		]
-	};
-	if (bookmark) {
-		response = secondPage;
-	}
-
-	if (search) {
-		response.Items = response.Items
-			.filter(item => item.orgUnitName.toLowerCase().includes(search.toLowerCase()));
-	}
-
-	return await new Promise(resolve =>	setTimeout(() => resolve(response), 100));
-}
 
 /**
  * @property {{id: string, displayName: string, selected: boolean}[]} _filterData
@@ -81,22 +29,17 @@ class SemesterFilter extends LitElement {
 		/** @type {{id: string, displayName: string, selected: boolean}[]} */
 		this._filterData = [];
 		this._bookmark = null;
+	}
 
-		this._lms = new Lms();
+	async firstUpdated() {
+		this._lms = this.isDemo ? new FakeLms() : new Lms();
+		await this._loadData();
 	}
 
 	get selected() {
 		return this._filterData
 			.filter(semester => semester.selected)
 			.map(semester => semester.id);
-	}
-
-	async firstUpdated() {
-		await this._loadData();
-	}
-
-	async _getData(search) {
-		return this.isDemo ? await testData(this._bookmark, search) : await this._lms.fetchSemesters(this._bookmark, search);
 	}
 
 	async _loadData(clear) {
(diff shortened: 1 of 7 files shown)


Vitalii-Misechko:
This is early preview PR.
I need to 

- Update back-end part to work with search string

- Write some tests

components/semester-filter.js
```diff
@@ -0,0 +1,169 @@
+import './simple-filter';
+
+import {html, LitElement} from 'lit-element';
+
+async function fetchData(bookmark, search) {
```
bemailloux, line 5: This can probably go into `lms.js`

components/semester-filter.js
```diff
@@ -0,0 +1,169 @@
...
+		url.searchParams.set('bookmark', bookmark);
+	}
+	if (search) {
+		url.searchParams.set('search', search);
+	}
+	const response = await fetch(url);
+	return await response.json();
+}
+
+async function testData(bookmark, search) {
```
bemailloux, line 18: `Future improvement` This (and all the other testData providers) can be moved to a TestDataProvider class

components/semester-filter.js
```diff
@@ -0,0 +1,169 @@
...
+			.filter(item => item.orgUnitName.toLowerCase().includes(search.toLowerCase()));
+	}
+
+	return await new Promise(resolve =>	setTimeout(() => resolve(response), 100));
+}
+
+/**
+ * @property {{id: string, displayName: string, selected: boolean}[]} _filterData
+ * @property {string} _bookmark
+ * @property {string} _search - serach string
```
bemailloux, line 74: Hmm, looks like searching is only done on the client side, not on the server side. That means users could potentially need to click "Load more" many times to find the semester they're looking for.

For a better user experience we probably want server-side searching. (I didn't consider this when reviewing the other PR)

components/semester-filter.js
Vitalii-Misechko: Good point!

components/semester-filter.js
Vitalii-Misechko: I've updated my [back-end PR](https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/14230/overview)

Vitalii-Misechko:
Updated PR. It seems only manual tests are left

components/semester-filter.js
```diff
@@ -0,0 +1,103 @@
...
+	_setSelectedState(id, selected) {
+		this._filterData.find(semester => semester.id === id).selected = selected;
+	}
+
+	render() {
+		return html`
+			<d2l-simple-filter
+				@d2l-simple-filter-selected="${this._updateFilterSelections}"
+				name="${this.localize('components.semester-filter.name')}"
+				load-more-text="${ifDefined(this._bookmark ? this.localize('components.semester-filter.loadMore') : undefined)}"
```
Vitalii-Misechko, line 79: `TIL`: turning on and off attribute

components/semester-filter.js
```diff
@@ -0,0 +1,103 @@
...
+ * @property {string} _search - serach string
+ * @property {string[]} selected - array of selected ids
+ * @fires d2l-insights-semester-filter-change
+ */
+class SemesterFilter extends Localizer(LitElement) {
+
+	static get properties() {
+		return {
+			isDemo: { type: Boolean, attribute: 'demo' },
+			pageSize: { type: Number, attribute: 'page-size' },
```
bemailloux, line 21: Cool! I thought `attribute` was `true/false` only, I didn't know you could set a custom name like this

test/components/semester-filter.test.js
```diff
@@ -0,0 +1,113 @@
...
+			checkboxes.find(checkbox => checkbox.value === '10007').simulateClick();
+
+			const event = await listener; // if no event is fired, this will time out after 2 seconds
+			expect(event.type).to.equal('d2l-insights-semester-filter-change');
+			expect(event.target).to.equal(el);
+		});
+	});
+
+	after(() => {
+		fetchMock.reset();
```
bemailloux, line 110: 👍

Vitalii-Misechko:
Removed responsiveness from the checklist. I assume that the new component `d2l-filter-dropdown-*` is responsive by design. Moreover, we might want to implement and test responsiveness when all component of the page in place. I assume, that implementation of the responsive interface varies depending on available components on the page (some component will be hidden and some could be converted to a label when the screen shrinks). Here is screen shot below. It shows me that there is a some good amount of work on responsiveness

image

Vitalii-Misechko:
Accessibility notes:
* `Load more` is focusable and selectable
* `Load more` works if a user presses whitspace bar on keyboard
* When `Load more` disappears (when no more pages) pressing Tab navigates to `Role` filter. 
  * FF - navigates to the dashboard page (can be addressed in separate PR if needed)
  * All browsers reports that the current element it the dashboard. We might want change this in the future. (Didn't find a quick way to set focus on `<d2l-simple-filter>`

PR #23, opened 2020-08-10
Semesters show up as "OrgUnitName (Id: OrgUnitId)"

Vitalii-Misechko:
[US117427](https://rally1.rallydev.com/#/detail/userstory/396954104020?fdp=true): [Engagement][Filters] Semester

JFYI @anhill-D2L @MykolaGalian @rohitvinnakota @hyehayes

### Functional Testing

- Unit tests
- Verified manually

### UX and docs - N/A

image

### DIFF
--- a/locales/en.js
+++ b/locales/en.js
@@ -9,6 +9,7 @@ export default {
 
 	"components.semester-filter.name": "Semesters",
 	"components.semester-filter.loadMore": "Load More",
+	"components.semester-filter.semester-name": "{orgUnitName} (Id: {id})",
 
 	"components.simple-filter.search-label": "Search",
 	"components.simple-filter.search-placeholder": "Search...",
--- a/components/semester-filter.js
+++ b/components/semester-filter.js
@@ -58,7 +58,7 @@ class SemesterFilter extends Localizer(LitElement) {
 
 		this._filterData = currentData.concat(data.Items.map(item => ({
 			id: item.orgUnitId.toString(),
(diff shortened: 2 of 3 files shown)


johngwilkinson:
UX: demo to Kevin. This is not consistent with the ou filter design, which shows ids like this: "Some Org Unit (Id: 12345)". I don't have a strong opinion about which way to do it, but I would think they should match. An tiny advantage of not having the term "Id" is that the code won't have to localize it.

bemailloux:
Maybe check with both Megan and Kevin? The AC is inconsistent with the design (in screenshot below), so they should resolve that conflict.

image

Vitalii-Misechko:
> both should be  OrgUnitName (Id: OrgUnitID)
https://d2l.slack.com/archives/CMNGJ22G0/p1597077957420800

Vitalii-Misechko:
Updated screenshot

locales/en.js
```diff
@@ -9,6 +9,7 @@ export default {
 
 	"components.semester-filter.name": "Semesters",
 	"components.semester-filter.loadMore": "Load More",
+	"components.semester-filter.semester-name": "{orgUnitName} (Id: {id})",
```
johngwilkinson, line 12: No change needed, but it occurs to me that if this were `{orgUnitName} (Id: {orgUnitId})`, you could pass it `item` directly.

Vitalii-Misechko: Hm...sounds convenient

PR #24, opened 2020-08-11
US119132 Table style edits

bemailloux:
Some minor things that came out of demo.

@johngwilkinson The colour I used for the header is the same as other tables in the LMS, but those other tables have a border between the header row and other rows, so the separation is more obvious. That's one of the changes I made, along with a border on every row and rounded corners on all 4 corners.

Quality notes
* Testing
  * [x] Chrome
  * [x] FF (Note: the border appears lighter on Firefox. That's also the case for other tables in the LMS e.g. Users table in the users tool)
  * [x] Edgium
  * [x] Edge Legacy (The border-radius actually appears in Edge now)
* UX
  * [x] Demo to Kevin
* Security, Perf, Cost, Devops, Docs: N/A  

Screenshot after (left) and before (right)
image

@hyehayes @Vitalii-Misechko @rohitvinnakota @MykolaGalian


styles/table.style.js
```diff
@@ -27,6 +28,34 @@ export const tableStyle = css`
...
+	border-top-right-radius: 8px;
+}
+
+table tr:last-child td:first-child {
+	border-bottom-left-radius: 8px;
+}
+
+table tr:last-child td:last-child {
+	border-bottom-right-radius: 8px;
 }
```
bemailloux, line 59: Setting individual borders and border-radiuses like this seems hacky, but it's actually how the rest of the LMS does it.

anhill-D2L: I wonder why having a border isn't the default for d2l-table

bemailloux: Cause it's not a d2l-table. It's a regular table that uses a `d2l-table` class I defined in this file. There is another (maybe more than 1) `d2l-table` CSS class elsewhere in the LMS, but it's not being loaded on this page.

I suppose having the same name could cause conflicts, but as long as we don't inadvertently load those classes (unlikely since in Github they exist in a deprecated repo), and as long as no one loads our classes, it should be fine. Happy to rename if you think this might be an issue though

johngwilkinson:
https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-not-separate-css-into-a-separate-file-for-no-reason

Is there any reason for us to have a `styles` dir or a table styles file? If only one component is using the styles, they should go in the class file.

bemailloux:
> Is there any reason for us to have a styles dir or a table styles file?

I didn't realize this was an issue:

> With unbundled builds, each file that gets imported is a separate request made by the browser. While this can be very powerful, we want to avoid extra requests if we can.

So the only reason would be organization, which is less important than perf.

components/table.js
```diff
@@ -20,7 +19,67 @@ class Table extends Localizer(RtlMixin(LitElement)) {
...
+			:host([dir="rtl"]) table {
+				text-align: right;
+			}
+
+			.d2l-table {
+				background-color: #fff;
+				border-collapse: separate;
+				border-spacing: 0;
+				width: 100%;
+				text-align: left;
```
johngwilkinson, line 32: doesn't this override the rtl setting above? I'm surprised you need to set text-align for rtl at all (should be handled by the browser, for that particular css property), so I'm probably missing something.

components/table.js
```diff
@@ -20,7 +19,67 @@ class Table extends Localizer(RtlMixin(LitElement)) {
...
+				border-bottom: 1px solid var(--d2l-color-mica);
+			}
+
+			/* Table cell border and radius */
+			/* to get a radius on all corners *and* exactly 1px inner borders */
+			table tr:first-child th {
+				border-top: 1px solid var(--d2l-color-mica);
+			}
+
+			table tr th:first-child,td:first-child {
```
johngwilkinson, line 59: Does this work rtl? Maybe the "first-child" is on the right instead of the left.

bemailloux: :facepalm:
image

components/table.js
bemailloux: This is specifically because headers default to `text-align: center`. But no, this doesn't override the rtl setting above because `:host([dir="rtl"]) table` is more specific than `.d2l-table`. 

Maybe I should change it to `:host([dir="rtl"]) .d2l-table` so it's clear which style takes precedence.

(For anyone wondering what CSS specificity is: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity and https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#Specificity_2)

components/table.js
bemailloux: Fixing that was harder than I thought it would be. 
Test cases:
* LTR table with 1, 2, 3 columns
* RTL table with 1, 2, 3 columns
* LTR and RTL tables with 0 rows and 1,3 columns (other than header)

PR #25, opened 2020-08-12
Add results card to main page

rohitvinnakota:
Rally story [here](https://rally1.rallydev.com/#/23525809118d/custom/12155517225?detail=%2Fuserstory%2F396959491404&fdp=true)

**Quality Notes**
Created a new ResultsCard class that feeds in the proper values into the summary card. More cards can be added in the future in a similar way. 

**Functional testing**
Tested data with both Local build and LMS. Card renders as expected. I overloaded the returned data with smaller and larger(ex: 34891233 users) numbers and it does not break anything on the UI.

Tested on edge, chrome, and firefox

Security: n/a
Perf: n/a
UX: Sent some screenshots to Kevin: waiting on possible demo.

image

FYI @hyehayes @Vitalii-Misechko @anhill-D2L @MykolaGalian


engagement-dashboard.js
```diff
@@ -74,6 +74,7 @@ class EngagementDashboard extends Localizer(LitElement) {
 					margin-bottom: 25px;
 					display: flex;
 					flex-wrap: wrap;
+					width: 50%;
```
johngwilkinson, line 77: This seems odd. Would it be better to remove the styling from `.summary-card-container` until there are at least two cards and we can work on a good responsive layout?

components/summary-card.js
```diff
@@ -83,5 +91,15 @@ class SummaryCard extends MobxLitElement {
 	firstUpdated() {
 		this.addEventListener('click', () => this.data.isApplied = !this.data.isApplied);
 	}
+
+	_getDataToDisplay(dataRequestString) {
```
johngwilkinson, line 95: General note: this looks like a switch, which is a code-smell. Usually if you have a class that says "depending on some string or number, do a, or b, or c...", what you should have instead is a number of classes: one for a, one for b, etc.

One option here would be to make
- a summary-card component that just renders the values it's given: it has attributes for value, message (and other things, later)
- and an overall-summary component that is a MobxListElement that renders a summary-card based on the provided `data` property
This method would be in that second, higher-level, specialized component, and would just be `return this.data.userDataForDisplay.length`.

This approach means that this overall summary card isn't treated as a filter at all - which makes sense, given that it doesn't filter anything.

rohitvinnakota: Yeah ok, I like the no filter approach so I will change the implementation to reflect that.

engagement-dashboard.js
rohitvinnakota: I was testing what the layout would look like with 4 cards. But yes, this can be picked up separately when we are adding the final layout.

components/summary-card.js
```diff
@@ -5,7 +5,10 @@ class SummaryCard extends MobxLitElement {
 
 	static get properties() {
 		return {
-			data: { type: Object, attribute: false }
+			data: { type: Object, attribute: false },
+			cardTitle: { type: String, attribute: 'card-title' },
```
johngwilkinson, line 9: minor: I'd suggest shortening these to simply `title`, `value`, and `message`.

components/summary-card.js
```diff
@@ -55,33 +55,39 @@ class SummaryCard extends MobxLitElement {
 			}
 
 			.d2l-insights-summary-card-value {
-				color: lightsteelblue;
-				font-size: 40px;
+				color: var(--d2l-color-ferrite); /* should conditionally render this when adding more cards */
```
johngwilkinson, line 58: not sure if this comment is still useful here

components/summary-card.js
```diff
@@ -55,33 +55,39 @@ class SummaryCard extends MobxLitElement {
...
 			}
 		`;
 	}
 
 	render() {
-		console.log(`summary-card render ${this.data.id}`);
-
 		// NB: relying on mobx rather than lit-element properties to handle update detection: it will trigger a redraw for
 		// any change to a relevant observed property of the Data object
 		return html`<div class="d2l-insights-summary-card" ?applied="${this.data.isApplied}" ?loading="${this.data.isLoading}">
```
johngwilkinson, line 79: These classes - loading and applied - aren't actually used now, are they? If they were, I'd suggest moving them top as properties. But if they're not referenced by the css, they could just be removed.

Either way, this component should no longer need a reference to `data`, and can be a plain `LitElement` instead of a `MobxLitElement`.

components/results-card.js
```diff
@@ -0,0 +1,38 @@
...
+			data: { type: Object, attribute: false }
+		};
+	}
+
+	constructor() {
+		super();
+		this.data = {};
+	}
+
+	get cardMessage() {
```
johngwilkinson, line 20: these seem like private getters, in which case prepend `_` to the name

components/results-card.js
```diff
@@ -0,0 +1,38 @@
+import './summary-card.js';
+
+import { html } from 'lit-element';
+import { Localizer } from '../locales/localizer';
+import { MobxLitElement } from '@adobe/lit-mobx';
+
+class ResultsCard extends Localizer(MobxLitElement) {
```
johngwilkinson, line 7: I like how simple this class is!

locales/en.js
```diff
@@ -30,5 +30,6 @@ export default {
 
 	"components.insights-users-table.title": "User Details",
 	"components.insights-users-table.lastFirstName": "Last Name, First Name",
-	"components.insights-users-table.totalUsers": "Total Users: {num}"
+	"components.insights-users-table.totalUsers": "Total Users: {num}",
+	"components.insights-engagement-dashboard.resultsReturned": "Users returned within results."
```
johngwilkinson, line 34: This is really a question for Kevin and Megan, but I'm not sure what "within results" adds here.

components/results-card.js
```diff
@@ -0,0 +1,38 @@
...
+
+	static get properties() {
+		return {
+			data: { type: Object, attribute: false }
+		};
+	}
+
+	constructor() {
+		super();
+		this.data = {};
```
johngwilkinson, line 17: it doesn't matter much, but this default will result in an error in `get _cardValue()`, by the looks of it

test/components/results-card.test.js
```diff
@@ -0,0 +1,36 @@
...
+			const el = await fixture(html`<d2l-insights-results-card .data="${data}"></d2l-insights-results-card>`);
+			await expect(el).to.be.accessible();
+		});
+	});
+
+	describe('render', () => {
+		it('should render the proper value for users', async() => {
+			const el = await fixture(html`<d2l-insights-results-card .data="${data}"></d2l-insights-results-card>`);
+			await new Promise(resolve => setTimeout(resolve, 200)); // allow fetch to run
+			expect(el.shadowRoot.querySelector('d2l-labs-summary-card').value).to.deep.equal('4');
```
johngwilkinson, line 33: might as well assert the correct title and message, too

components/results-card.js
bemailloux: I had this same thing in the users table. Somehow it magically initializes without failing and I have no idea why

components/summary-card.js
```diff
@@ -55,33 +53,35 @@ class SummaryCard extends MobxLitElement {
...
+				<span class="d2l-insights-summary-card-message d2l-insights-summary-card-field">${html`${this.message}`}</span>
 			</div>
 		</div>`;
 	}
 
-	firstUpdated() {
-		this.addEventListener('click', () => this.data.isApplied = !this.data.isApplied);
-	}
 }
 customElements.define('d2l-labs-summary-card', SummaryCard);
```
bemailloux, line 87: This name can probably change to `d2l-insights-summary-card` since this isn't really a "labs" project anymore. (If you do change the name, don't forget to rename the `d2l-insights-summary-card` CSS class so you don't get [weird Edge problems](https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-not-name-a-css-class-the-same-as-a-component))

components/summary-card.js
```diff
@@ -55,33 +53,35 @@ class SummaryCard extends MobxLitElement {
...
-		return html`<div class="d2l-insights-summary-card" ?applied="${this.data.isApplied}" ?loading="${this.data.isLoading}">
-			<div class="d2l-insights-summary-card-title">${this.data.title}</div>
+		return html`<div class="d2l-insights-summary-card">
+			<div class="d2l-insights-summary-card-title">${this.title}</div>
 			<div class="d2l-insights-summary-card-body">
-				${(this.data.stats.delta !== undefined) ? html`<span class="d2l-insights-summary-card-delta d2l-insights-summary-card-field">${this.data.stats.delta}</span>` : ''}
-				<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field">${this.data.stats.value}</span>
-				<span class="d2l-insights-summary-card-message d2l-insights-summary-card-field">${html`${this.data.message}`}</span>
+				<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field">${this.value}</span>
+				<span class="d2l-insights-summary-card-message d2l-insights-summary-card-field">${html`${this.message}`}</span>
```
bemailloux, line 81: Why does `this.message` need to be wrapped in `html` here?

engagement-dashboard.js
```diff
@@ -123,8 +123,8 @@ class EngagementDashboard extends Localizer(LitElement) {
 				</div>
 
 				<h2 class="d2l-heading-3">${this.localize('components.insights-engagement-dashboard.summaryHeading')}</h2>
-				<div class="d2l-insights-summary-container">
-					${Object.values(this._data.filters).map(f => html`<d2l-labs-summary-card id="${f.id}" .data="${f}"></d2l-labs-summary-card>`)}
+				<div class="summary-container">
```
bemailloux, line 126: class name here should be changed back (or, maybe, just removed entirely)

rohitvinnakota: I'll change the name for now since the plan is to add more cards to the container in the future.

rohitvinnakota:
> Did you do any testing related to accessibility and responsiveness?

Yep, tested ltr ,rtl, resizing, and screen narration

PR #26, opened 2020-08-13
US119136 [Engagement][Filters] Replace simple-filter with facet-filter-sort component [WIP]

Vitalii-Misechko:
[US119136](https://rally1.rallydev.com/#/detail/userstory/410073841016?fdp=true): [Engagement][Filters] Replace simple-filter with facet-filter-sort component

Use new component for implementation of `Roles` and `Semesters` filter. I'd recommend using split view for this PR

New component PR https://github.com/Brightspace/insights-engagement-dashboard/pull/27 [now modified to remove "load more"]

### Functional Testing

-  Tests
   - [x] Selections persist across search and clear search, so you can build up a selection by searching several times
   - [x] Filter opener text is correct for 1 and many selected items
   - [x] Closing filter sends an event 
   - [x] Accessibility
   - [x] Keyboard navigation
   - [x] RTL
   - [x] Unit tests
- Browsers
   - [x]  Chrome
   - [x] Firefox
   - [x] Edge
   - [x] Edge Legacy
### UX and docs
   - [x] Demo to Kevin

image

PR #27, opened 2020-08-14
US119136  [Engagement][Filters] Replace simple-filter with facet-filter-sort component. New app level filter component

Vitalii-Misechko:
[US119136](https://rally1.rallydev.com/#/detail/userstory/410073841016?fdp=true): [Engagement][Filters] Replace simple-filter with facet-filter-sort component

This PR introduces `d2l-insights-dropdown-filter` component that I'm going to use instead of `d2l-simple-filter` to achieve all AC in the story. `d2l-insights-dropdown-filter` uses Gaudi's `d2l-filter-dropdown` component under the hood

PR Draft with changes to existing Semester and Role filters: https://github.com/Brightspace/insights-engagement-dashboard/pull/26 

### Functional Testing

-  Tests
   - [x] The search actions clear selected items
   - [x] When load more disappears input focus stays on filter element
   - [x] Filter opener text is correct for 1 and many selected items
   - [x] Load more preserves selected items
   - [x] Closing filter sends an event 
   - [x] Accessibility
   - [x] Keyboard navigation (without a screen reader filter does not open a list of items. I'll report it to Gaudi)
   - [x] Unit tests
- Browsers
   - [x]  Chrome
   - [x] Firefox
   - [x] Edge
   - [x] ~Edge Legacy (Edge HTML) - it hungs on my laptop~
### UX and docs
 - [x] Demo to Kevin

NB. This PR does not change `Semesters` and `Roles` filter
image

### DIFF
--- a/components/dropdown-filter.js
+++ b/components/dropdown-filter.js
@@ -32,6 +32,12 @@ class DropdownFilter extends Localizer(LitElement) {
 
 	static get styles() {
 		return [css`
+			:host {
+				display: inline-block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
 			d2l-button-subtle {
 				margin: 0.25rem;
 				display: grid;
--- a/components/dropdown-filter.js
+++ b/components/dropdown-filter.js
@@ -68,12 +68,14 @@ class DropdownFilter extends Localizer(LitElement) {
 	}
 
 	render() {
+		const openerSelectedText = this.localize('components.dropdown-filter.opener-text-multiple', {filterName: this.name, selectedCount: this._selectedCount});
+
 		return html`
 			<d2l-filter-dropdown
 				total-selected-option-count="${this._selectedCount}"
 				opener-text="${this.name}"
-				opener-text-single="${this.localize('components.dropdown-filter.opener-text-multiple', {filterName: this.name, selectedCount: this._selectedCount})}"
-				opener-text-multiple="${this.localize('components.dropdown-filter.opener-text-multiple', {filterName: this.name, selectedCount: this._selectedCount})}"
+				opener-text-single="${openerSelectedText}"
+				opener-text-multiple="${openerSelectedText}"
 				@d2l-filter-dropdown-cleared="${this._clearSelectionClick}"
 				@d2l-dropdown-close="${this._filterClose}"
 				>
--- a/components/dropdown-filter.js
+++ b/components/dropdown-filter.js
@@ -1,0 +1,164 @@
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import 'd2l-facet-filter-sort/components/d2l-filter-dropdown/d2l-filter-dropdown.js';
+import 'd2l-facet-filter-sort/components/d2l-filter-dropdown/d2l-filter-dropdown-category.js';
+import 'd2l-facet-filter-sort/components/d2l-filter-dropdown/d2l-filter-dropdown-option.js';
+
+import {css, html, LitElement} from 'lit-element';
+import {Localizer} from '../locales/localizer';
+import {repeat} from 'lit-html/directives/repeat.js';
+
+/**
+ * @property {string} name
+ * @property {{id: string, displayName: string, _selected: boolean}[]} data
+ * @property {boolean} more - shows Load More button if true
+ * @property {string[]} selected - returns list of ids for selected items
+ * @property {boolean} disable-search
+ * @fires d2l-insights-dropdown-filter-selected - event.detail contains {string} filterName, {string} id, and {boolean} selected
+ * @fires d2l-insights-dropdown-filter-load-more-click
+ * @fires d2l-insights-dropdown-filter-searched
+ * @fires d2l-insights-dropdown-filter-selection-cleared
+ * @fires d2l-insights-dropdown-filter-close
+ */
+class DropdownFilter extends Localizer(LitElement) {
+
+	static get properties() {
+		return {
+			name: {type: String, attribute: true},
+			data: {type: Array, attribute: false},
+			hasMore: {type: Boolean, attribute: 'more'},
+			disableSearch: {type: Boolean, attribute: 'disable-search'},
+			_selectedCount: {type: Number, attribute: true}
+		};
+	}
+
+	static get styles() {
+		return [css`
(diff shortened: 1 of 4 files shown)


Vitalii-Misechko:
FYI @hyehayes @rohitvinnakota @MykolaGalian @anhill-D2L

components/dropdown-filter.js
```diff
@@ -0,0 +1,164 @@
...
+ */
+class DropdownFilter extends Localizer(LitElement) {
+
+	static get properties() {
+		return {
+			name: {type: String, attribute: true},
+			data: {type: Array, attribute: false},
+			hasMore: {type: Boolean, attribute: 'more'},
+			disableSearch: {type: Boolean, attribute: 'disable-search'},
+			_selectedCount: {type: Number, attribute: true}
```
johngwilkinson, line 30: Seems odd to have a private attribute.

components/dropdown-filter.js
```diff
@@ -0,0 +1,164 @@
...
+				@d2l-filter-dropdown-cleared="${this._clearSelectionClick}"
+				@d2l-dropdown-close="${this._filterClose}"
+				>
+				<d2l-filter-dropdown-category
+					?disable-search="${this.disableSearch}"
+					category-text="${this.name}"
+					@d2l-filter-dropdown-option-change="${this._handleElementSelected}"
+					@d2l-filter-dropdown-category-searched="${this._handleSearchedClick}" >
+
+					${repeat(this.data, (item) => item.id + this._searchIteration, (item) => html`
```
johngwilkinson, line 81: I gather you're using `repeat` instead of `map` because "[t]he repeat directive performs efficient updates of lists based on user-supplied keys"? A comment here would be useful.

components/dropdown-filter.js
Vitalii-Misechko: Yeah...I need an update to happen whenever the `_selectedCount` changes to show a proper text on the filter. 
I'll try to get rid of it but probably in the future

components/dropdown-filter.js
Vitalii-Misechko: `repeat`+`(item) => item.id + this._searchIteration` is used to clear `selected` states from DOM. I wrote a comment at first usage of `_searchIteration`, in the constructor. I didn't find a way to write a comment right inside a Lit template.

Vitalii-Misechko: Here is the thread that describes the problem that I address with `repeat`
https://d2l.slack.com/archives/C0PHG3QB0/p1597234974419600

components/dropdown-filter.js
```diff
@@ -0,0 +1,164 @@
...
+		return [css`
+			d2l-button-subtle {
+				margin: 0.25rem;
+				display: grid;
+			}
+
+			d2l-filter-dropdown-category[disable-search] {
+				padding-top: 0px;
+			}
+		`];
```
bemailloux, line 44: Just FYI: explicitly setting the :host's display value (probably to `block`) is a good idea
https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-style-the-display-value-of-the-host

components/dropdown-filter.js
```diff
@@ -0,0 +1,164 @@
...
+		this._searchIteration = 0;
+	}
+
+	render() {
+		return html`
+			<d2l-filter-dropdown
+				total-selected-option-count="${this._selectedCount}"
+				opener-text="${this.name}"
+				opener-text-single="${this.localize('components.dropdown-filter.opener-text-multiple', {filterName: this.name, selectedCount: this._selectedCount})}"
+				opener-text-multiple="${this.localize('components.dropdown-filter.opener-text-multiple', {filterName: this.name, selectedCount: this._selectedCount})}"
```
bemailloux, line 71: The opener text could be calculated once before the `return html...` statement

components/dropdown-filter.js
```diff
@@ -0,0 +1,164 @@
...
+			</d2l-filter-dropdown>
+		`;
+	}
+
+	async updated() {
+		if (!this.hasMore && this._loadMoreClickFlag) {
+			this._loadMoreClickFlag = false;
+
+			// move focus to the first filter item/option when the last page is loaded and Load More button is removed
+			await this.updateComplete;
```
bemailloux, line 98: Seems weird to wait for `updateComplete` inside `updated`. It's like you're changing the LitElement lifecycle order, I dunno if that could have unexpected effects

components/dropdown-filter.js
johngwilkinson: I think that's fine, but it could just be a property, couldn't it (`attribute: false`)?

components/dropdown-filter.js
bemailloux: Related: I wonder if making updated `async` could cause issues. In the litelement examples it looks like it's just synchronous

components/dropdown-filter.js
Vitalii-Misechko: I'd defer applying `display` on this component to the next PR. `display: block` causing an effect which in turn requires re-styling `of view-filters-container`-div

image

components/dropdown-filter.js
Vitalii-Misechko: fixed

components/dropdown-filter.js
johngwilkinson: You could set it to `inline-block`.

components/dropdown-filter.js
Vitalii-Misechko: you're right! We can safely remove `await this.updateComplete`. 

>  I dunno if that could have unexpected effects

I think in this case `await this.updateComplete` is executed immediately because `updated` should come after `updateComplete`-promise resolved. I added this statement because I read somewhere on the internet that someone had a problem with setting focus because unfinished rendering.

components/dropdown-filter.js
Vitalii-Misechko: this works!

Vitalii-Misechko:
Manual testing is finished!

components/dropdown-filter.js
Vitalii-Misechko: `attribute: false` works properly. Thanks!

PR #28, opened 2020-08-14
Us117426/org unit filter style right arrows

johngwilkinson:
Quality Notes
* functional
  * perms: user enrolled in only some courses, not cascading (enrolled as student, end-user, instructor - perm added, administrator - only the latter two enrollment types appear)
  * browsers: chrome, ff, edgacy, edgium
  * expand/collapse, check & uncheck - all as expected
  * search field is not hooked up in this story
* perf: n/a
* security: n/a
* devops: n/a
* ux/docs
  * demoed to Kevin
  * screen reader (Narrator) and keyboard navigation: included extra info so the screen reader can indicate tree depth and parent
  * rtl: confirmed indentation and padding is as expected, and layout is reversed (tested on demo page and in LMS with arabic user locale)

image


components/ou-filter.js
```diff
@@ -10,13 +11,11 @@ const PARENTS = 3;
 const CHILDREN = 4;
 const STATE = 5;
 
-const SEMESTER_TYPE = 5;
```
bemailloux, line 13: 👍

components/tree-selector-node.js
```diff
@@ -22,15 +26,18 @@ import {css, html, LitElement} from 'lit-element/lit-element.js';
...
 	static get properties() {
 		return {
 			name: { type: String },
 			tree: { type: Object, attribute: false },
 			getTree: { type: Object, attribute: false },
 			isOpen: { type: Boolean, reflect: true, attribute: 'open' },
 			selectedState: { type: String, reflect: true, attribute: 'selected-state' },
-			isRoot: { type: Boolean, reflect: true, attribute: 'root' }
+			isRoot: { type: Boolean, reflect: true, attribute: 'root' },
+			// for screen readers
```
bemailloux, line 38: Neat!

johngwilkinson:
> Does it resize horizontally as well as vertically? Is there a cap on how large the element can get in either direction?

Just vertically, and mostly that means a scrollbar appears. The total size doesn't change a whole lot.

PR #32, opened 2020-08-19
Us118254/user table paging

bemailloux:
Brightspace UI Labs have synthesized a new component: pagination. Made from a piece of LitElement exposed to waaaaay too many edge-legacy-frustration radiation.

I guess you could call those problems I had... "edge cases" *ba-dum-tss*

Quality notes
* Testing (Chrome, FF, Edgium, Edge Legacy)
  * It changes which page of users are displayed when the page number changes
  * It changes how many users are displayed, and goes back to the first page, when the page size is changed
  * "Total users" still shows the total length, not the page size
  * NOT tested: what happens if the underlying data changes (e.g. if we do a server-side filter). I'm hoping we can defer this until the "Apply" story.
* Perf:
  * The list of users to display is stored as a calculated mobx property. It doesn't seem to be recalculated if I just access the property, which is good because the recalculation (which re-sorts the list) will probably be slow for a large list. 
  * I tested this by console.logging every time `get userDataForDisplay() ` was called.
* UX:
  * The pagination element is almost localized, except for the a11y label on the input box
  * It RTL's properly - except for a small glitch with the division symbol in Edge.
image
  * It is responsive (tested Chrome, FF, Edgium, Edge Legacy)
image
  * Screen reader and keyboard nav is ok (tested Chrome, Edge Legacy)
* Devops: 
  * https://github.com/Brightspace/brightspace-integration/pull/2398 needs to be merged as well
* Cost: N/A
* Security: N/A
* Docs: N/A

TODO list
- [x] Safari tests failing
- [x] Responsive notes
- [x] RTL notes
- [x] a11y "integration" testing

@Vitalii-Misechko @rohitvinnakota @MykolaGalian @hyehayes

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -34,8 +34,17 @@ async function demoData() {
 					[6606, 'Dev', 1, [0]]
 				],
 				users: [
-					[100, 'First', 'Last'],
-					[200, 'Test', 'Student']
+					[100, '1First', 'Last'],
+					[200, 'ATest', 'Student'],
+					[300, 'BTest', 'Student'],
+					[400, 'CTest', 'Student'],
+					[500, 'DTest', 'Student'],
+					[600, 'ETest', 'Student'],
+					[700, 'FTest', 'Student'],
+					[800, 'GTest', 'Student'],
+					[900, 'HTest', 'Student'],
+					[1000, 'ITest', 'Student'],
+					[1100, 'JTest', 'Student']
 				],
 				semesterTypeId: 25,
 				selectedOrgUnitIds: [1, 2]
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -33,18 +33,18 @@ async function demoData() {
 					[9, 'Faculty 2', 7, [6606]],
 					[6606, 'Dev', 1, [0]]
 				],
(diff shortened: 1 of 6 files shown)


locales/localizer.js
```diff
@@ -17,48 +17,9 @@ export const Localizer = superclass => class extends LocalizeMixin(superclass) {
...
 				switch (lang) {
-					case 'ar':
-						translations = await import('./ar.js');
-						break;
-					case 'da':
-						translations = await import('./da.js');
-						break;
-					case 'de':
-						translations = await import('./de.js');
-						break;
```
bemailloux, line 28: Deleted all of these because BSI build crashes if you try to import a file that doesn't exist. We can add them back in once we have all the langterm files.

engagement-dashboard.js
```diff
@@ -34,8 +34,17 @@ async function demoData() {
 					[6606, 'Dev', 1, [0]]
 				],
 				users: [
-					[100, 'First', 'Last'],
-					[200, 'Test', 'Student']
+					[100, '1First', 'Last'],
```
johngwilkinson, line 37: Maybe have some of these out of sort order to quickly see if sorting is working

PR #33, opened 2020-08-20
US119136: case-insensitive search for dropdown filters

johngwilkinson:


johngwilkinson:
> Should there be a corresponding unit test?

Yes - added.

PR #34, opened 2020-08-23
US118248 Initial local filter PR

bemailloux:
Filters records locally by selected OUs, semesters and roles. Not in scope for this PR: reloading data from server when truncated

Quality notes
* Testing
  * Tried various combinations of OU selections, semester selections and role selections. Verified record and user results by checking which ids were in the results and which weren't.
  * Browsers: Chrome, FF, Edgium, Edge Legacy
* Security: N/A
* Docs: N/A
* Cost: N/A
* Devops: N/A
* Perf
  * Loops through each record and checks if that record's ancestors are in the selected OU list. The ancestors list for a given record should be pretty small (hopefully 4-5 in most cases) and the ancestors list is pre-generated, so looking that up is O(1). So in total the runtime should be close-ish to O(n)
  * Generating the ancestors list should also be approximately O(n). It might take up a lot of space (number of records * 4 or 5) but since number of records is capped at 10k, max memory for that list might be ~50k * 8 bytes = 400 KB (but in reality would be less than that), which is kind of a similar size to the records array right now.
* UX
  * Right now it applies the filter every time a new selection is made. For future perf reasons we may want to change that to only applying when the dropdown is closed.
  
TODO
- [x] Edge fix / testing
- [x] Test on mobile to see if a phone can handle it

FYI @rohitvinnakota @anhill-D2L @MykolaGalian @hyehayes


bemailloux:
> Is cross-filtering in-scope for this story?

No, not at all. I was just wondering if we could take advantage of the eventual cross-filtering that will exist to skip having to apply semester filters. The answer is no.

> General thought: I believe it will go better if you work from a record-centric point of view

Sounds good, I'll try this

bemailloux:
It also seems like using ES6 Maps is a more optimal "dictionary" representation than objects, so I'll switch to those as well.

model/data.js
```diff
@@ -72,27 +96,90 @@ export class Data {
...
+	// the reason for separating this from getRecordsInView is to try not to reapply the top level filters if
+	// we don't need to.
+	get records() {
+		let records = this.serverData.records;
+
+		const roleFilterApplied = this.selectorFilters.roleIds.length;
+		const orgUnitFilterApplied = this.selectorFilters.orgUnitIds.length;
+		const semesterFilterApplied = this.selectorFilters.semesterIds.length;
+
+		if (roleFilterApplied || orgUnitFilterApplied || semesterFilterApplied) {
```
johngwilkinson, line 125: I don't see the advantage of this check. It's a minor perf improvement in only one case (no filters).

model/data.js
```diff
@@ -72,27 +96,90 @@ export class Data {
...
+	}
+
+	get users() {
+		let users = this.serverData.users;
+
+		const roleFilterApplied = this.selectorFilters.roleIds.length;
+		const orgUnitFilterApplied = this.selectorFilters.orgUnitIds.length;
+		const semesterFilterApplied = this.selectorFilters.semesterIds.length;
+
+		if (roleFilterApplied || orgUnitFilterApplied || semesterFilterApplied) {
```
johngwilkinson, line 155: This is still O(n^2) - think of the case where most or all users match the filters. Also, card filters will apply here, so the `if` check is going to keep changing.

I'm envisioning it like this, with a dictionary `this.users` created when the server data is first received:
```
const unique = arr => [...new Set(arr)];
const userIdsInView = unique(this.getRecordsInView().map(x => x[USERID]));
return userIdsInView.map(x => this.users[x]);
```

I believe that's O(n). Basically, I'm saying you want to use a lookup (dictionary) to do the join between records and user data, not a nested loop (which is effectively what `users.filter(user => userIdsInView.includes(user[USER.ID]))` is).

model/orgUnitAncestors.js
```diff
@@ -0,0 +1,59 @@
...
+			return this.ancestorsMap.get(orgUnitId);
+		}
+
+		const parents = parentsMap.get(orgUnitId);
+		const ancestorsSet = new Set([orgUnitId]);
+
+		if (parents.length && parents[0] !== 0) {
+			// this isn't the top level - recursively add ancestors
+			parents.forEach(parent => {
+				ancestorsSet.add(parent);
```
johngwilkinson, line 40: you can omit this line, right? because every node is it's own ancestor, `ancestorsOfParent` will include `parent`.

model/orgUnitAncestors.js
```diff
@@ -0,0 +1,59 @@
...
+
+	_addOrgUnitToAncestorsMap(orgUnitId, parentsMap) {
+		if (this.ancestorsMap.has(orgUnitId)) {
+			return this.ancestorsMap.get(orgUnitId);
+		}
+
+		const parents = parentsMap.get(orgUnitId);
+		const ancestorsSet = new Set([orgUnitId]);
+
+		if (parents.length && parents[0] !== 0) {
```
johngwilkinson, line 37: option: instead of this `if`, throw a guard clause up top: `if (orgUnitId === 0) then return []` - then inline the `parents` variable.

model/data.js
bemailloux: Good point - if enrollments are capped at 10k then looping through the whole list once shouldn't take too long

model/orgUnitAncestors.js
bemailloux: I was thinking that in some cases the list of OUs returned by the server may not include the root, so the top level would not have 0 as its parent. But I'm wrong: the root will always be returned

model/data.js
```diff
@@ -1,4 +1,21 @@
...
+};
+
+const USER = {
+	ID: 0,
+	FIRST_NAME: 1,
+	LAST_NAME: 2
+};
+
+function unique(array) {
+	return [...new Set(array)];
```
johngwilkinson, line 17: BTW, I know I suggested this, but I don't actually know if it's fastest. I'm pretty sure it's linear time, at least, because set insertion should be.

model/data.js
```diff
@@ -72,27 +101,72 @@ export class Data {
 		recordProvider().then(data => {
 			this.serverData = data;
 			this.isLoading = false;
+			this._orgUnitAncestors = new OrgUnitAncestors(data.orgUnits);
+			this._userDictionary = new Map(data.users.map(user => [user[0], user]));
```
johngwilkinson, line 105: minor: use constant `user[ID]`

model/data.js
```diff
@@ -72,27 +101,72 @@ export class Data {
...
+	}
+
+	applySemesterFilters(semesterIds) {
+		this.selectorFilters.semesterIds = semesterIds;
+	}
+
+	// the reason for separating this from getRecordsInView is to try not to reapply the top level filters if
+	// we don't need to.
+	get records() {
+		if (!this._orgUnitAncestors) {
```
johngwilkinson, line 125: Why is this here?

model/data.js
bemailloux: It's certainly the simplest way 🤷 plus I'm not sure how much we can optimize getting unique values from a list of 10k.

model/data.js
bemailloux: In this code:
```recordProvider().then(data => {
	this.serverData = data;
	this.isLoading = false;
	this._orgUnitAncestors = new OrgUnitAncestors(data.orgUnits);
	this._userDictionary = new Map(data.users.map(user => [user[0], user]));
});
```

Mobx seems to want to want to run this function immediately after `this.serverData` is set, at which point `this._orgUnitAncestors` is still null. That's probably a waste - I'll look into how to do all the setup atomically.

bemailloux: (The easiest thing might just be to throw serverData, orgUnitAncestors and userDictionary in a subclass)

johngwilkinson: You're sure this isn't due to rendering that starts before the server data is loaded? Components on the page generally render twice: once with empty/default data, and then again when the recordProvider resolves. Maybe you just need to add default values for `this._orgUnitAncestors` and `this._userDictionary`.

bemailloux: When debugging, I put a breakpoint on the first line of `get records` and saw that it ran 3 times: 
- once with no records and `orgUnitAncestors = null`
- once with all the records and `orgUnitAncestors = null`
- once with all the records and `orgUnitAncestors != null`

Since `serverData` and `orgUnitAncestors` are both observable, I think my hypothesis above is what is happening.

johngwilkinson: Ah, interesting. The change to `serverData` likely results in a new render, which calls `records` - so it's a bit of a race condition, isn't it. Setting a default value would handle that case too, right? Although it would be nicest if that data handler were atomic, and I'm not sure why it's not (javascript doesn't interrupt threads, as I understand it).

bemailloux: The better alternative does seem to be:
- make orgUnitAncestors not observable
- initialize it with `new OrgUnitAncestors([])`
- remove that null check

bemailloux: >javascript doesn't interrupt threads, as I understand it

This was my understanding as well, but mobx is still very much a black box to me

bemailloux: Wait a second: in this case we can get around the whole issue by making only serverData observable and updating serverData last, so that updates are only triggered _after_ everything else is ready.

engagement-dashboard.js
```diff
@@ -144,6 +145,7 @@ class EngagementDashboard extends Localizer(LitElement) {
 				<h2 class="d2l-heading-3">${this.localize('components.insights-engagement-dashboard.summaryHeading')}</h2>
 				<div class="d2l-insights-summary-container">
 					<d2l-insights-results-card .data="${this._data}"></d2l-insights-results-card>
+					<d2l-insights-debug-card .data="${this._data}" metric-to-display="recordsLength" title="Records" message="number of records within filter parameters"></d2l-insights-debug-card>
```
Vitalii-Misechko, line 148: This line is temporal just to show how to use `d2l-insights-debug-card`, right?

model/data.js
```diff
@@ -1,98 +1,123 @@
...
+	ORG_UNIT_ID: 0,
+	USER_ID: 1,
+	ROLE_ID: 2
+};
+
+const USER = {
+	ID: 0,
+	FIRST_NAME: 1,
+	LAST_NAME: 2
+};
```
Vitalii-Misechko, line 15: I like how these consts make the code is more readable 👍

engagement-dashboard.js
bemailloux: Yes. I'm thinking I might put a "demo" flag on it so that it only shows up in the `npm start` page

test/model/data.test.js
```diff
@@ -1,21 +1,205 @@
...
+				const userId = user[0];
+				return [100, 200, 300].includes(userId);
+			});
+
+			sut.applyOrgUnitFilters(orgUnitFilters);
+
+			expect(sut.users).to.deep.equal(expectedUsers);
+		});
+
+		it('should return filtered users when semester filter is applied', async() => {
```
Vitalii-Misechko, line 177: >> Is cross-filtering in-scope for this story? I didn't quite follow your notes about this above, but the rules are: the semester and role filters are not filtered at all (this is an optimization for the server-side case); the ou filter shows only course offerings that fall under selected semesters, and their ancestors.

> No, not at all. I was just wondering if we could take advantage of the eventual cross-filtering that will exist to skip having to apply semester filters. The answer is no.

Could you clarify why the answer is no? It seems we don' need this and the next tests if I follow the discussion correctly because 

> ... semester and role filters are not filtered at all ...

P.S. Sorry if you've already answered the question

bemailloux: Np, it was kind of a complicated proposal

The OrgUnit tree filter options will get filtered by selections in the semesters filter. Example: `Course 1 for Semester 1` will only appear in the OU tree if no semesters are selected or if Semester 1 is selected.

So, in theory, a user can only select OrgUnits that correspond to the selected semesters. If that's true, then we don't need to apply the Semesters filter, because the Semesters filter _has already been applied_ to the list of OUs the user could possibly select from.

But in practice this doesn't work for 2 reasons:
1. It assumes previous OU selections that aren't in the selected semesters will not be included. Users probably won't expect OUs they selected to be excluded. Example: if I select `Course 1 for Semester 2`, then select `Semester 1`, as a User I would still expect `Course 1 for Semester 2` to appear in my results (even though it no longer appears in the OU filter).
2. If you didn't filter records by the semester selections, with our current implementation top level OUs would also be included, even though they shouldn't.

johngwilkinson: Are you saying we will include data from checked-off org units regardless of what semesters are selected? I don't think that's what we want - and I also don't think it's what your code here does: this will show only records that match both the ou filter and the semester filter, which is good. Org Units that don't match the semester will stay selected but hidden from the filter and the results, so if the semester filter gets cleared, they will appear selected again.

When we are dealing with truncation and the a semester filter gets set, we go server side and now the list of org units in the filter should be only those matching the semester filter. In this case the semester filter won't do anything client-side *unless the data is now not truncated and the user narrows the selection* - so the "filter records by both ou and semester" approach here is also good in this case.

bemailloux: >Are you saying we will include data from checked-off org units regardless of what semesters are selected? I don't think that's what we want

Oops, you are right. What I was thinking was "as a user, I don't want my selections to be arbitrarily erased when I select a semester", which isn't the same as "always include selected OUs".

bemailloux:
I don't think I can perf test on mobile, unless there's some method of accessing a Sauce VM or something.

johngwilkinson:
> I don't think I can perf test on mobile, unless there's some method of accessing a Sauce VM or something.

There's a pretty wide range of mobile devices, but I guess you could try it out on your own via ngrok to your dev instance.

bemailloux:
> I guess you could try it out on your own via ngrok to your dev instance.

Tried this. I can't get BSI (prod or dev) to load files on mobile, even though the desktop version works fine. 
https://d2l.slack.com/archives/C0PHG3QB0/p1598381206028900

johngwilkinson:
> > I guess you could try it out on your own via ngrok to your dev instance.
> 
> Tried this. I can't get BSI (prod or dev) to load files on mobile, even though the desktop version works fine.
> https://d2l.slack.com/archives/C0PHG3QB0/p1598381206028900

Oh, yeah, because it's trying to load resources from es-dev-server, which is on a different port, and it may also be trying to hit localhost. If its' not the latter, it might be possible to get ngrok to open both ports on the same url, I'm not sure.

bemailloux:
> Oh, yeah, because it's trying to load resources from es-dev-server, which is on a different port, and it may also be trying to hit localhost. If its' not the latter, it might be possible to get ngrok to open both ports on the same url, I'm not sure.

Bingo, port-forwarding 8080 and 44444 to my phone did it (in addition to connecting my phone via USB and some Chrome remote management stuff). Thanks!

bemailloux:
My phone (Galaxy A8) doesn't appear to have any perf issues with applying the filters. But it's not a bad phone (2.2/1.8 GHz 8-core CPU, 4 GB RAM) so testing with a Galaxy S5 or iPhone 5 or worse would be more useful.

Vitalii-Misechko (approved):
LGTM!

PR #36, opened 2020-08-25
Overdue Assignments Card

MykolaGalian:
https://rally1.rallydev.com/#/detail/userstory/402312564692

Design: https://xd.adobe.com/view/1b11cf2e-1d8e-4daf-61a5-77de56ac239d-6d20/
LMS PR: https://github.com/Brightspace/lms/pull/652

Functional test:
1. In LMS created two overdue assignments for course #1 and one overdue assignment for course #2, verified that the expected payload meets expectations.
2. Checked - a user who has submitted the assignment on time or late does not appear in the overdue count.

Performance Testing - In LMS method FetchDropboxIncompleteUserIds (dropbox.OrgUnitId, dropbox.DropBoxId) takes 1-2 ms (with 300 users who are registered to this dropbox and course). Method FetchDropboxEntities (orgUnitId) - getting all dropboxes for a given course (three dropboxes) - less than 1ms. 

UX - Kevin Demo - in progress
Devops - N/A
Docs - N/A

@rohitvinnakota @hyehayes @Vitalii-Misechko @anhill-D2L @johngwilkinson


model/data.js
```diff
@@ -89,6 +89,26 @@ export class Data {
...
+		// eslint-disable-next-line prefer-const
+		let numOverdueAssignmentsByUserIds = {};
+		for (const userId in groupOverdueAssignmentsByUserIds) {
+			if (!numOverdueAssignmentsByUserIds[userId]) numOverdueAssignmentsByUserIds[userId] = 0;
+			groupOverdueAssignmentsByUserIds[userId].forEach(oa => { numOverdueAssignmentsByUserIds[userId] += oa; });
+		}
+
+		return  Object.values(numOverdueAssignmentsByUserIds).filter((number) => {
+			return number > 0;
+		}).length;
```
bemailloux, line 109: To store the user information you can use an [ES6 map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map). It should have better performance than storing things in an object.

Also, you have 3 loops here:
1. Loop over all records. Create the following map: userId => list of counts of overdue assignments for that user
2. Loop over all users. Create the following map: userId => total number of overdue assignments for that user
3. Loop over all users. Count how many users have overdue assignments > 0

This can be simplified into 1 loop:
1. Loop over all records. Create the following map: userId => total number of overdue assignments for that user. If overdue assignments > 0, then add an entry for that user in the map, and increment by the number of overdue assignments.
2. To get the total number of users with overdue assignments, just do Map.size

MykolaGalian: I will try using underscorejs library

bemailloux: If you're going to use any library it should be lodash to be consistent with the rest of D2L. But this should be doable without an external library

johngwilkinson: If you do want to use lodash, be sure to use the specific packages you need, rather than bringing in all of lodash.

model/data.js
```diff
@@ -108,6 +108,19 @@ export class Data {
 			});
 	}
 
+	get usersNumWithOverdueAssignments() {
+		const groupOverdueAssignmentsByUserIds = this.serverData.records
```
johngwilkinson, line 112: It doesn't affect the results yet, but this should be `this.serverData.getRecordsInView()`

model/data.js
```diff
@@ -108,6 +108,19 @@ export class Data {
...
+			.reduce((array, record) => {
+				if (!array[record[1]]) array[record[1]] = [];
```
johngwilkinson, line 114: A few minor things here:
* Bryan added some constants to use for array indices instead of `1` and `3`; please add one for overdue count and use them
* the thing being reduced here is not an array, but an object, so I'd suggest calling it `counts` or `courseOverdueCountByUserId`
* You're only using the length of the inner arrays - so why not just use an integer and increment it?

model/data.js
```diff
@@ -108,6 +108,19 @@ export class Data {
 			});
 	}
 
+	get usersNumWithOverdueAssignments() {
```
johngwilkinson, line 111: this getter can be decorated with "computed" to help mobx (add to the `decorate` call at the bottom of this file)

johngwilkinson:
Other cases to test before closing off the story:
* a user who has submitted the assignment on time does not appear in the overdue count
* a user who has submitted the assignment late also does not appear in the overdue count

test/components/overdueAssignments-card.test.js
```diff
@@ -0,0 +1,34 @@
...
+		it('should pass all axe tests', async() => {
+			const el = await fixture(html`<d2l-insights-overdue-assignments-card .data="${data}"></d2l-insights-overdue-assignments-card>`);
+			await expect(el).to.be.accessible();
+		});
+	});
+
+	describe('render', () => {
+		it('should render the proper number of users with overdue assignments', async() => {
+			const el = await fixture(html`<d2l-insights-overdue-assignments-card .data="${data}"></d2l-insights-overdue-assignments-card>`);
+			await new Promise(resolve => setTimeout(resolve, 200)); // allow fetch to run
```
bemailloux, line 27: You should be able to replace this `setTimeout` with `await el.updateComplete`. (Although, that might not work because Sauce tests are flaky).

In either case, the comment on this line is wrong.

components/summary-card.js
```diff
@@ -60,6 +60,15 @@ class SummaryCard extends LitElement {
 				margin-inline-start: 30px;
 			}
 
+			.d2l-insights-summary-card-value[is-value-clickable] {
+				color: var(--d2l-color-celestine);
+				cursor: pointer;
```
Vitalii-Misechko, line 65: TIL: CSS manages cursor appearance

components/summary-card.js
```diff
@@ -1,12 +1,12 @@
...
 
 class SummaryCard extends LitElement {
-
 	static get properties() {
 		return {
 			title: { type: String, attribute: 'card-title' },
 			value: { type: String, attribute: 'card-value' },
-			message: { type: String, attribute: 'card-message' }
+			message: { type: String, attribute: 'card-message' },
+			isvalueclickable: { type: Boolean, attribute: 'is-value-clickable' }
```
Vitalii-Misechko, line 9: JS property should be camelCase `isValueClickable`. [Documentation](https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-use-correct-casing-for-attributes-and-properties)

components/summary-card.js
```diff
@@ -71,17 +80,22 @@ class SummaryCard extends LitElement {
...
 	render() {
 		// NB: relying on mobx rather than lit-element properties to handle update detection: it will trigger a redraw for
 		// any change to a relevant observed property of the Data object
 		return html`<div class="d2l-insights-summary-card">
 			<div class="d2l-insights-summary-card-title">${this.title}</div>
 			<div class="d2l-insights-summary-card-body">
-				<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field">${this.value}</span>
-				<span class="d2l-insights-summary-card-message d2l-insights-summary-card-field">${this.message}</span>
+			${this.isvalueclickable ?
+			html`<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field" ?is-value-clickable=${this.isvalueclickable} @click=${this._getUsers}>${this.value}</span>` :
```
Vitalii-Misechko, line 94: AFAIK `span`-tag does not have `is-value-clickable`-attribute

Vitalii-Misechko: Oh...I see. It's styling thing

Vitalii-Misechko (approved):
LGTM!

PR #37, opened 2020-08-25
Us118249/org unit filter v2

johngwilkinson:
Quality Notes
* functional
  * test selection in tree and search results
    * confirmed filters applied as expected
    * confirmed tree is updated as expected (children and parents marked selected, indeterminate, or unselected)
  * chrome, ff, edgacy, edgium
  * test on demo app and dev lms via BSI
* security: n/a
* performance: renders only the part of the tree that's expanded
* devops: n/a
* ux/docs: demoed to Kevin


components/tree-filter.js
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	get selected() {
+		return this.tree.selected;
+	}
+
+	/**
+	 * @returns {Promise} - resolves when all tree-selector-nodes, recursively, have finished updating
+	 */
+	get treeUpdateComplete() {
```
johngwilkinson, line 165: This was needed to get the dropdown to resize properly. Without waiting for the whole tree to render, the dropdown would sometimes resize too soon and cut off some of the content.

johngwilkinson:
The main change here was to separate the data model (`Tree`) from the rendering classes (tree-selector and tree-selector-node) so that whether the user interacts with the search results list or the tree, the updates are handled in exactly the same way.

This made it possible to add search as a separate div of nodes that match the search string.

components/ou-filter.js
```diff
@@ -91,19 +54,20 @@ class OuFilter extends Localizer(MobxLitElement) {
...
+		const tree = {};
+		this.data.serverData.orgUnits
+			.filter(x => x[TYPE] !== this.data.serverData.semesterTypeId)
+			.forEach(x => {
+				tree[x[ID]] = [...x, [], 'none', false];
+			});
 
 		// fill in arrays of children, so we can look up both parents and children directly
-		this.data.serverData.orgUnits.forEach(x => x[PARENTS].forEach(y => y !== 0 && this._tree[y][CHILDREN].push(x[ID])));
+		Object.values(tree).forEach(x => x[PARENTS].forEach(y => tree[y] && tree[y][CHILDREN].push(x[ID])));
```
bemailloux, line 68: This is pretty hard to read, so I think it's worth moving to a named function. Want to make sure my understanding is correct: it looks like maps each orgUnitId to its immediate children?

components/tree-filter.js
```diff
@@ -0,0 +1,261 @@
...
+export const PARENTS = 3;
+export const CHILDREN = 4;
+export const STATE = 5;
+export const OPEN = 6;
+
+export class Tree {
+	/**
+	 * @param tree - a map from ids to arrays, each of which has fields defined by the above constants; may be modified at any time
+	 * @param selectedIds - array of ids which are selected initially (won't be modified)
+	 * @param leafTypes - array of values in the TYPE column; nodes of these types cannot be expanded
```
bemailloux, line 22: What's the reason for making this an array instead of a single value?

components/tree-filter.js
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	get selected() {
+		return this._getSelected(this.rootId);
+	}
+
+	getChildren(id) {
+		// coming soon: handle truncation case (getChildren has a callback which calls LMS)
+		if (!id) id = this.rootId;
+		if (!id) return [];
```
bemailloux, line 42: What is this condition needed for? I'm not sure what the situation is when `rootId` doesn't exist.

components/tree-filter.js
```diff
@@ -0,0 +1,261 @@
...
+			.sort((a, b) => this._nameForSort(a).localeCompare(this._nameForSort(b)));
+	}
+
+	setOpen(id, isOpen) {
+		this.tree[id][OPEN] = isOpen;
+	}
+
+	getMatching(searchString) {
+		return Object.values(this.tree).filter(x =>
+			!this._isRoot(x) && this._nameForSort(x) && this._nameForSort(x).toLowerCase().includes(searchString.toLowerCase())
```
bemailloux, line 55: `nit` the `_nameForSort` calculation could be stored in a variable. Also, it would be very weird if an orgUnit didn't have either an ID or a name. Is that check to guard against the "id=0" pseudo-orgUnit?

components/tree-filter.js
```diff
@@ -0,0 +1,261 @@
...
+		return this.tree
+			.getChildren(id)
+			.map(x => this._renderNode(x, parentName, indentLevel + 1));
+	}
+
+	_renderNode(x, parentName, indentLevel) {
+		const isOpen = x[OPEN];
+		return html`<d2l-insights-tree-selector-node slot="tree"
+					name="${this.localize('components.tree-filter.node-name', { orgUnitName: x[NAME], id: x[ID] })}"
+					data-id="${x[ID]}"
```
bemailloux, line 209: Yay :)

components/tree-selector-node.js
```diff
@@ -195,103 +176,59 @@ class TreeSelectorNode extends Localizer(RtlMixin(LitElement)) {
...
-			{ bubbles: true, composed: false }
-		));
+	async _waitForTreeUpdateComplete() {
+		await this.updateComplete;
+		const slot = this.shadowRoot.querySelector('slot');
+		// to be sure all child nodes have been added, instead of using flatten,
+		// we recursively walk down the tree, waiting for each node's update to complete
+		if (slot) {
+			const childNodes = slot.assignedNodes({ flatten: false });
+			return Promise.all(childNodes.map(node => node.treeUpdateComplete));
```
bemailloux, line 230: I wonder if updates sometimes fail for whatever reason. If so I guess the element wouldn't render properly.

components/ou-filter.js
johngwilkinson: Yes, it fills in all the CHILDREN arrays with the ids of immediate children. I'll refactor to make it more readable.

components/tree-filter.js
johngwilkinson: Trying, where it's easy, to keep it general. Some other type of tree might have more than one leaf node type. It's just about as easy to use an array, so I did.

components/tree-filter.js
johngwilkinson: It's good to be defensive anyway, but this branch actually gets taken during the initial render before the server data is available.

components/tree-filter.js
johngwilkinson: Storing everything as a local variable is encouraged by some and frowned on by others. The performance advantage in most cases is negligible, and it makes refactoring harder by adding locals scoped over multiple lines.

Again, defensive is good, but I think you're right that ou 0 caused problems without this check.

components/tree-selector-node.js
johngwilkinson: Well, the promise would reject, which in the one use case so far would result in the dropdown content not being resized. Not sure if there's a better way to handle that sort of error, since it implies something pretty unexpected happened.

johngwilkinson:
@Vitalii-Misechko bump

Vitalii-Misechko (approved):

PR #38, opened 2020-08-26
fix demo data

johngwilkinson:


engagement-dashboard.js
```diff
@@ -21,7 +21,16 @@ async function demoData() {
...
-				records: [{ UserId: 1, OrgUnitId: 1 }, { UserId: 2, OrgUnitId: 1 }, { UserId: 2, OrgUnitId: 2 }],
+				records: [
+					[1, 100, 500],
+					[1, 200, 600],
+					[2, 200, 700],
+					[2, 300, 700],
+					[2, 400, 700],
+					[2, 500, 700],
+					[8, 200, 700],
+					[6, 600, 700]
```
bemailloux, line 32: The format is `orgUnitId, userId, roleId` but some of the role ids here are userIds in the users list. Role ids and user ids don't need to be mutually exclusive, but maybe they should be here so it's a little easier to understand the data model?

johngwilkinson:
@Vitalii-Misechko bump

Vitalii-Misechko (approved):

PR #39, opened 2020-08-26
US118248 Refactoring, groundwork for calling server with filter params

bemailloux:
Moving some things around and renaming others. I also "de-classed" LMS and FakeLMS because I wanted to be able to use the functions statically.

Testing: 
* Still works as before on Chrome, Edge Legacy
* Hardcoding OrgUnit ids into `lms.fetchData` fetched the correctly filtered data from the server. Tested with a Department, a Course Template and a Course Offering. OU filter is updated based on `serverData.selectedOrgUnitIds`

@MykolaGalian @rohitvinnakota @hyehayes @anhill-D2L

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -2,67 +2,17 @@ import './components/histogram-card.js';
 import './components/ou-filter.js';
 import './components/results-card.js';
 import './components/debug-card.js';
-import './components/insights-role-filter.js';
+import './components/role-filter.js';
 import './components/semester-filter.js';
 import './components/users-table.js';
 import './components/table.js';
 
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { Data } from './model/data.js';
+import { fetchData } from './model/lms.js';
+import { fetchData as fetchDemoData } from './model/fake-lms.js';
 import { heading3Styles } from '@brightspace-ui/core/components/typography/styles';
 import { Localizer } from './locales/localizer';
-
-async function fetchData() {
-	const response = await fetch('/d2l/api/ap/unstable/insights/data/engagement');
-	return await response.json();
-}
-
-async function demoData() {
-	return new Promise(resolve =>
-		setTimeout(
-			() => resolve({
-				records: [
-					[1, 100, 500],
-					[1, 200, 600],
-					[2, 200, 700],
-					[2, 300, 700],
-					[2, 400, 700],
-					[2, 500, 700],
-					[8, 200, 700],
-					[6, 600, 700]
-				],
-				orgUnits: [
-					[1, 'Course 1', 3, [3, 4]],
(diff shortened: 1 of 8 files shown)


model/fake-lms.js
```diff
@@ -1,76 +1,118 @@
...
-	/**
-	 * @returns {{Identifier: string, DisplayName: string, Code: string|null}[]}
-	 */
-	async fetchRoles() {
-		const demoData = [
-			{ Identifier: '500', DisplayName: 'Administrator' },
-			{ Identifier: '600', DisplayName: 'Instructor' },
-			{ Identifier: '700', DisplayName: 'Student' }
-		];
+// adding variables here to match signature of real LMS. The filters don't actually work though.
```
bemailloux, line 1: I'd like to give these filters real behaviour. I don't think that's useful before I implement the "get data from server if necessary" though, so making the filters work here will come later.

engagement-dashboard.js
```diff
@@ -2,59 +2,18 @@ import './components/histogram-card.js';
...
-					[1100, 'JTest', 'JStudent']
-				],
-				semesterTypeId: 25,
-				selectedOrgUnitIds: [1, 2]
-			}),
-			100
-		)
-	);
-}
-
```
bemailloux, line 57: Moved to lms.js and fake-lms.js

model/data.js
```diff
@@ -53,7 +53,7 @@ export class Data {
 		// mobx will run _persist() whenever relevant state changes
 		autorun(() => this._persist());
 
-		recordProvider().then(data => {
+		recordProvider(this.selectorFilters).then(data => {
```
Vitalii-Misechko, line 56: why do you pass `this.selectorFilters` into `recordProvider`?

Vitalii-Misechko (approved):
LGTM!

model/data.js
bemailloux: It's because the function signature of `recordProvider` is actually:
```
export async function fetchData({ roleIds = [], semesterIds = [], orgUnitIds = [] });
```
So recordProvider can use the filters to make the correct query call.

If you're asking "Why do it initially in the constructor? Filters are empty" it's because in the future we will want filters to be saved across page reloads.

bemailloux:
> I don't see any real advantage of having an "lms" class over having the individual filters know how to make LMS calls, since it means they can't be re-used in other projects without including the whole LMS class.

That's fair. I don't know to what extent things from this repo will be shared in the future, so I went with the organization that I thought made the most sense considering this repo only. 

(Plus, that LMS "class" is pretty small for now - I'm hoping it stays that way)

model/data.js
Vitalii-Misechko: Thanks! I just have not recognized proper function behind `recordProvider`

PR #42, opened 2020-08-28
US118248 Reload data from server if necessary

bemailloux:
Reloads data from server based on logic in [the org unit spike](https://rally1.rallydev.com/#/23525809118d/workviews?detail=%2Fuserstory%2F402788988184&view=349baf7a-c63a-4fa4-a43b-2bb09274aad7) with one modification: if we know that all orgUnits in the current query have ancestors in the previous query, then we know that their related data has already been loaded. So it does not reload from the server in that case.

Quality notes
* Testing
  * Chrome, FF, Edgium, Edge Legacy
  * Calling server-side filtering has the same results as local filtering
    * NOTE: there appears to be a bug on the server-side where applying a selectedOrgUnitId and a selectedSemesterId always returns no records
* Security: N/A
* Devops: N/A
* Docs: N/A
* Perf:
  * One possible optimization which is not applied: if all data is loaded in the initial query and no filters are applied initially, then there's never any need to reload data from the server.
  * With the logic that is currently implemented there are actually very few cases where we _don't_ reload data from the server. 
* UX
  * It doesn't apply the filters on filter close anymore. Filters are applied every time a checkbox is clicked, so applying them on close is redundant.

@rohitvinnakota @MykolaGalian @hyehayes @anhill-D2L

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -88,17 +88,17 @@ class EngagementDashboard extends Localizer(LitElement) {
 				<h1 class="d2l-heading-1">${this.localize('components.insights-engagement-dashboard.title')}</h1>
 
 				<div class="view-filters-container">
-					<d2l-insights-ou-filter .data="${this._data}" @d2l-insights-ou-filter-change="${this._onOuFilterChange}"></d2l-insights-ou-filter>
+					<d2l-insights-ou-filter
+						.data="${this._data}"
+						@d2l-insights-ou-filter-change="${this._orgUnitFilterChange}"
+					></d2l-insights-ou-filter>
 					<d2l-insights-semester-filter
 						page-size="10000"
 						?demo="${this.isDemo}"
 						@d2l-insights-semester-filter-change="${this._semesterFilterChange}"
-						@d2l-insights-semester-filter-close="${this._semesterFilterChange}"
-						>
-					</d2l-insights-semester-filter>
+					></d2l-insights-semester-filter>
 					<d2l-insights-role-filter
-						@d2l-insights-role-filter-change="${this._handleRoleSelectionsUpdated}"
-						@d2l-insights-role-filter-close="${this._handleRoleSelectionsUpdated}"
+						@d2l-insights-role-filter-change="${this._roleFilterChange}"
 						?demo="${this.isDemo}"
 					></d2l-insights-role-filter>
 				</div>
@@ -116,22 +116,18 @@ class EngagementDashboard extends Localizer(LitElement) {
 		`;
 	}
 
-	_handleRoleSelectionsUpdated(event) {
+	_roleFilterChange(event) {
 		event.stopPropagation();
-
-		console.log(`List of selected role ids: ${event.target.selected}`);
 		this._data.applyRoleFilters(event.target.selected.map(roleId => Number(roleId)));
 	}
 
-	_onOuFilterChange(e) {
-		console.log(`got ou filter change with selected ids ${JSON.stringify(e.target.selected)}`);
-		this._data.applyOrgUnitFilters(e.target.selected);
+	_orgUnitFilterChange(event) {
+		event.stopPropagation();
+		this._data.applyOrgUnitFilters(event.target.selected);
 	}
 
 	_semesterFilterChange(event) {
 		event.stopPropagation();
-
-		console.log(`List of selected semesters: ${event.target.selected}`);
 		this._data.applySemesterFilters(event.target.selected.map(semesterId => Number(semesterId)));
 	}
 }
--- a/model/data.js
+++ b/model/data.js
@@ -1,8 +1,10 @@
 import { action, autorun, computed, decorate, observable } from 'mobx';
-import { CardFilter } from './cardFilter.js';
-import OrgUnitAncestors from './orgUnitAncestors.js';
+import { OrgUnitSelectorFilter, RoleSelectorFilter, SemesterSelectorFilter } from './selectorFilters.js';
 
-const RECORD = {
+import { CardFilter } from './cardFilter.js';
+import OrgUnitAncestors from './orgUnitAncestors';
+
+export const RECORD = {
 	ORG_UNIT_ID: 0,
 	USER_ID: 1,
 	ROLE_ID: 2,
@@ -10,10 +12,17 @@ const RECORD = {
 	CURRENT_FINAL_GRADE: 4
 };
 
-const USER = {
+export const USER = {
 	ID: 0,
 	FIRST_NAME: 1,
 	LAST_NAME: 2
+};
+
+export const ORG_UNIT = {
+	ID: 0,
+	NAME: 1,
+	TYPE: 2,
+	ANCESTORS: 3
 };
 
 function unique(array) {
@@ -26,26 +35,31 @@ function countUnique(records, field) {
 
 export class Data {
 	constructor({ recordProvider, cardFilters }) {
+		this.recordProvider = recordProvider;
+		this._orgUnitAncestors = null;
+		this._userDictionary = null;
+
+		// @observables
 		this.isLoading = true;
-
-		// for the purposes of these top level filters, "empty filters" is equivalent to "no filter applied"
-		this.selectorFilters = {
-			roleIds: [],
-			orgUnitIds: [],
-			semesterIds: []
-		};
-
 		this.serverData = {
 			records: [],
 			orgUnits: [],
 			users: [],
-			selectedOrgUnitIds: []
+			isRecordsTruncated: false,
+			isOrgUnitsTruncated: false,
+			semesterTypeId: null,
+			selectedOrgUnitIds: [],
+			selectedRolesIds: [],
+			selectedSemestersIds: []
 		};
 
-		this._orgUnitAncestors = null;
-		this._userDictionary = null;
+		this.roleFilter = new RoleSelectorFilter(this.serverData);
+		this.semesterFilter = new SemesterSelectorFilter(this.serverData, null);
+		this.orgUnitFilter = new OrgUnitSelectorFilter(this.serverData, null);
 
 		this.cardFilters = {};
+
+		// additional setup
 		cardFilters
 			.map(params => new CardFilter(params, this))
 			.forEach(f => this.cardFilters[f.id] = f);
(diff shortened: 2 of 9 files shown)


model/data.js
```diff
@@ -1,19 +1,28 @@
 import { action, autorun, computed, decorate, observable } from 'mobx';
+import { OrgUnitSelectorFilter, RoleSelectorFilter, SemesterSelectorFilter } from './selectorFilters.js';
```
bemailloux, line 2: Probably better viewed side-by-side

model/data.js
```diff
@@ -53,51 +67,70 @@ export class Data {
...
+		if (this.orgUnitFilter.shouldReloadFromServer(newOrgUnitIds)) {
+			this.loadData({ newOrgUnitIds });
+		} else {
+			this.orgUnitFilter.orgUnitIds = newOrgUnitIds;
+		}
+	}
 
-			return roleCriterion && orgUnitCriterion && semesterCriterion;
+	// @computed
+	get records() {
```
bemailloux, line 119: Moved records logic to the new SelectorFilters classes

model/orgUnitAncestors.js
```diff
@@ -57,5 +52,23 @@ class OrgUnitAncestors {
...
+	 */
+	hasAncestorsInList(orgUnitId, listToCheck) {
+		const ancestorsSet = this.getAncestorsFor(orgUnitId);
+		if (!ancestorsSet) {
+			return false;
+		}
+
+		return [...ancestorsSet].some((ancestor) => {
+			return listToCheck.includes(ancestor);
+		});
```
Vitalii-Misechko, line 71: `thought`: why convert `set` to `array` if we can do something like this
```JS
listToCheck.some(ou => ancestorsSet.has(ou));
```
P.S. Didn't check if proposed code actually works

model/selectorFilters.js
```diff
@@ -0,0 +1,87 @@
...
+			return true;
+		}
+
+		return this._orgUnitAncestors.hasAncestorsInList(record[RECORD.ORG_UNIT_ID], this.semesterIds);
+	}
+
+	shouldReloadFromServer(newSemesterIds) {
+		return this._isRecordsTruncated
+			|| isFilterCleared(this.semesterIds, newSemesterIds)
+			|| newSemesterIds.some(newSemesterId => !this.semesterIds.includes(newSemesterId));
```
Vitalii-Misechko, line 47: `thought`: semester array, I think is quite small, so O(n^2) is fine here, I guess

test/model/selectorFilters.test.js
```diff
@@ -0,0 +1,234 @@
...
+	describe('RoleSelectorFilter', () => {
+		describe('shouldInclude', () => {
+			it('should return true if the filter has no selected ids', () => {
+				const record = [1, 1, 1]; // doesn't matter what the ids are here
+				const sut = new RoleSelectorFilter({ selectedRolesIds: [], isRecordsTruncated: false });
+				expect(sut.shouldInclude(record)).to.be.true;
+			});
+
+			it('should return true if the record role id is in the selected ids', () => {
+				const record = [1, 1, 1];
```
Vitalii-Misechko, line 14: `proposal`: `record`=>`updateRecord` or `userModifiedRecord`

Vitalii-Misechko (approved):
LGTM!

Vitalii-Misechko:
> NOTE: there appears to be a bug on the server-side where applying a selectedOrgUnitId and a selectedSemesterId always returns no records

Looking into it

test/model/selectorFilters.test.js
bemailloux: This is just checking whether `record` should be included in the list of records to display. So maybe the function name should change to `shouldDisplay`?

model/orgUnitAncestors.js
bemailloux: This works, thanks for the suggestion!

test/model/selectorFilters.test.js
Vitalii-Misechko: I'm okay the way it is.  It seems I misunderstood `record` usage. `shouldDisplay` - sounds good to me too.

engagement-dashboard.js
```diff
@@ -114,14 +114,14 @@ class EngagementDashboard extends Localizer(LitElement) {
 		`;
 	}
 
-	_handleRoleSelectionsUpdated(event) {
+	_roleFilterChange(event) {
 		event.stopPropagation();
 
 		console.log(`List of selected role ids: ${event.target.selected}`);
```
johngwilkinson, line 120: We can remove these `console` statements now.

model/data.js
```diff
@@ -54,51 +68,70 @@ export class Data {
...
+		return this.serverData.records.filter(record => {
```
johngwilkinson, line 121: Braces and `return` can be removed.

No change needed, but personally I'd be tempted to generalize: 
```
this._viewFilters = [roleFilter, semesterFilter, orgUnitFilter]; // in constructor
...
return this.serverData.records.filter(record =>
    this._viewFilters.every(f=> f.shouldInclude(record))
);
```

model/data.js
```diff
@@ -54,51 +68,70 @@ export class Data {
...
+			return this.roleFilter.shouldInclude(record)
+				&& this.semesterFilter.shouldInclude(record)
+				&& this.orgUnitFilter.shouldInclude(record);
 		});
 	}
 
+	// @computed
```
johngwilkinson, line 128: I like these decorator comments

model/orgUnitAncestors.js
```diff
@@ -57,5 +52,23 @@ class OrgUnitAncestors {
...
+	 * @param {[Number]} listToCheck - an array of orgUnitIds which potentially has ancestors in it
+	 * @returns {boolean}
+	 */
+	hasAncestorsInList(orgUnitId, listToCheck) {
+		const ancestorsSet = this.getAncestorsFor(orgUnitId);
+		if (!ancestorsSet) {
+			return false;
+		}
+
+		return listToCheck.some(potentialAncestor => {
```
johngwilkinson, line 69: no need for braces

model/selectorFilters.js
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	shouldInclude(record) {
+		return !isFilterApplied(this.roleIds) || this.roleIds.includes(record[RECORD.ROLE_ID]);
+	}
+
+	shouldReloadFromServer(newRoleIds) {
+		return this._isRecordsTruncated
+			|| isFilterCleared(this.roleIds, newRoleIds) // filter was cleared
+			|| newRoleIds.some(newRoleId => !this.roleIds.includes(newRoleId));
```
johngwilkinson, line 25: Here and below, we don't want to reload if we went from no filters to some filters _unless_ the relevant data was truncated (because empty filters implicitly include all ids).

model/selectorFilters.js
```diff
@@ -0,0 +1,87 @@
...
+	shouldInclude(record) {
+		if (!isFilterApplied(this.semesterIds) || !this._orgUnitAncestors) {
+			return true;
+		}
+
+		return this._orgUnitAncestors.hasAncestorsInList(record[RECORD.ORG_UNIT_ID], this.semesterIds);
+	}
+
+	shouldReloadFromServer(newSemesterIds) {
+		return this._isRecordsTruncated
```
johngwilkinson, line 45: ou tree truncation is also relevant to the semester filter (reducing the semester selection from all to some or from some to fewer can reduce the number of org units in view, and so eliminate truncation).

model/selectorFilters.js
```diff
@@ -0,0 +1,87 @@
...
+		if (!isFilterApplied(this.orgUnitIds) || !this._orgUnitAncestors) {
+			return true;
+		}
+
+		return this._orgUnitAncestors.hasAncestorsInList(record[RECORD.ORG_UNIT_ID], this.orgUnitIds);
+	}
+
+	shouldReloadFromServer(newOrgUnitIds) {
+		return this._isRecordsTruncated
+			|| this._isOrgUnitsTruncated
```
johngwilkinson, line 69: I don't think ou truncation is relevant to the ou filter, because this filter is not applied to the ou tree (it's that general rule that filters don't filter themselves).

model/data.js
bemailloux: I was trying to stay away from this because originally I wasn't sure if mobx would pick up on changes correctly if I put everything in an array. But now that the `observable` is directly on each filter I think I can try this now

bemailloux: I also wasn't sure how much of a benefit it would be to separate selector filters from card filters. I think it makes sense because card filters will never cause a page refresh, but if selector and card filters share an API it may be possible to put them all into one array

model/selectorFilters.js
bemailloux: So it sounds like we should reload from server if the OU tree is truncated regardless of any other conditions?

johngwilkinson: For the semester filter, the spike says to reload on any change if either truncation flag is true. This seems right, because expanding the list of semesters could bring new data into view, so a reload is needed; and narrowing the list could remove truncation of either kind.

johngwilkinson: There's an important nuance I just realized is missing in the code: if the data _is not_ truncated, you have to compare against the last _query_, not the current state of the filter. It works like this:
* initially, let's say my data is truncated
* I select a couple of recent semesters, which narrows the view; the data is now not truncated (in this example)
* now I can play with the filters at will, and everything happens client-side, even if I deselect one of the semesters (because we've already got the non-truncated data for the other one locally); and if I select that same semester again, we've got the data, so no need to reload
* but if I clear the semester filter or add another semester (that wasn't in the last server call), then I need to reload because we don't have that data

So in the no-truncation case, you determine whether to reload by comparing with the last server call (i.e. the data that is available in `data.serverResponse`, not the current state of the filter, which may have changed.

bemailloux: Good catch! Fixing this will help reduce the number of server calls.

bemailloux:
@johngwilkinson All comments have been addressed in the latest commits. One interesting consequence of the changes is that on a dev LMS (or any LMS where all records can be viewed at once) it never needs to reload from the server anymore.

I did check that it _can_ reload from the server if it needs to by manually changing the `_latestServerQuery` list in each filter.

johngwilkinson:
> @johngwilkinson All comments have been addressed in the latest commits. One interesting consequence of the changes is that on a dev LMS (or any LMS where all records can be viewed at once) it never needs to reload from the server anymore.
> 
> I did check that it _can_ reload from the server if it needs to by manually changing the `_latestServerQuery` list in each filter.

I'll try to take a look this afternoon.

Quick note: I was thinking that for e2e testing, it would be useful to make LMS system-only configs for the two truncation limits. This might also save us in production if there's a site having performance problems.

bemailloux:
> Quick note: I was thinking that for e2e testing, it would be useful to make LMS system-only configs for the two truncation limits. This might also save us in production if there's a site having performance problems.

Good idea!

johngwilkinson:
I'd like to see e2e testing and test cases listed out so reviewers can think about whether there are any missed edge cases:
* no truncation on page load
* truncation on page load of ou tree only; fixed by selecting a semester
* truncation on page load of records only; fixed by selecting an ou
* truncation on page load of records only; fixed by selecting an ou
* the above three together
* several filter changes given truncation
* several filter changes given truncation on page load but none now
* reloading data from server
  * on filter clear (one case per filter)
  * on add semester to selection
etc.

model/data.js
```diff
@@ -54,51 +70,70 @@ export class Data {
...
+		if (this._selectorFilters.orgUnit.shouldReloadFromServer(newOrgUnitIds)) {
+			this.loadData({ newOrgUnitIds });
+		} else {
+			this._selectorFilters.orgUnit.selected = newOrgUnitIds;
+		}
+	}
 
-			return roleCriterion && orgUnitCriterion && semesterCriterion;
+	// @computed
+	get records() {
```
johngwilkinson, line 124: Will you be making the change to `getRecordsInView()` in this PR?

bemailloux: No, going to open a new one.

bemailloux:
## E2E Testing

### Default Setup
* set LMS to use [Vitalii's branch with server-side filter fix](https://github.com/Brightspace/lms/pull/897/files)
* default dev LMS data + 1 addition: add Grades Handler Test Offering and IPSIS Test Offering New to IPSIS Semester New
* default course limit 5k
* default enrollment limit 50k

### No truncation
* Setup: 
  * default dev LMS data + 1 addition: add Grades Handler Test Offering and IPSIS Test Offering New to IPSIS Semester New
  * default course limit 5k
  * default enrollment limit 50k
* Summary: it never reloads, because it doesn't need to
* Test cases
  * Initial load
  * Filters tested individually with 0, 1, multiple selections
  * Filters tested together with 0, 1, multiple selections each
  * Semester and role filters: clear button
  
### OU truncation only
* Setup
  * course limit: 10
* Summary: reloading logic works. Some UX issues
* Test cases
  * Initial load
  * OU filter individual: never reloads from server (filters don't filter themselves)
  * Semester filter individual
    * Reloads from server on 0 -> 1, 1 -> multiple, and 1 -> 0 selections, regardless of what was selected
	* Reloads from server on multiple -> 1 selections if and only if orgUnits were truncated
	* Doesn't reload from server if orgUnits was not truncated and it went from multiple -> fewer selections
  * Role filter individual: never reloads from server
  * Semester and OU filter
    * Selecting a semester: 
	  * will load entire OU tree if the entire OU tree fits inside the limit (e.g. with IPSIS Semester New which contains 1 OU)
	  * otherwise it remains truncated (e.g. with Fall test semester which contains > 300 OUs)
	  * open question: what happens if the OU tree is truncated for a given Semester?
	* Selecting an OU after selecting a semester
	  * Can lead to a problem: if the semester is deselected, then the selected OU might get truncated. Which means that it can no longer be deselected unless the corresponding semester is selected again.
	  * Possible countermeasures:
	    * If ONLY selected filters are not visible in the UI (whether by truncation or because they were cross-filtered by semesters) don't apply them - so that they don't mess up the record results.
		  * This might be contrary to what some users expect - if an incompatible OU and a semester are selected, because records are appearing some users might think that OU and that semester have courses in common when they actually don't.
		* Add a "Clear" button to the OU filter
		
### Record truncation only
* Setup
  * enrollment limit: 100 and 1000
* Summary: works as expected
* Test cases
  * Initial load
  * OU filter individual
    * Always reloads from server on 0 -> 1, 1 -> 0
	* Reloads from server on 1 -> multiple the first time, but not subsequent times, unless a new id that hasn't been loaded before is selected
	  * This is also true of org units with descendants: it reloads the first time a parent OU is selected, but subsequent selection and deselection of its children doesn't trigger a reload
	* Never reloads from server on multiple -> fewer unless records were truncated
  * Semester filter individual: same tests and results as OU filter (minus the ancestors point)
  * Role filter individual: same tests and results as OU filter (minus the ancestors point)
  * Semester + OU filters
    * Loading new ids that haven't been loaded yet always causes a server reload
	* Clearing a filter generally causes a reload, except for the following case: 
	  * if one filter has a selection, and the other filter has been cleared before, then subsequent clears and selections on the other filter don't cause reloads (except if records are truncated)
  * Ditto for other filter combinations (including all 3)
  
### OU and record truncation
* Setup
  * enrollment limit: 250
  * course limit: 100
* Test cases
  * Same as others
  * OU filter: OU truncation doesn't cause a server refresh
  * Semester filter: OU truncation does cause a server refresh

PR #43, opened 2020-09-01
US120044 Time in content vs grades card

MykolaGalian:
https://rally1.rallydev.com/#/detail/userstory/423599366764

Added the client-side code to display a scatter plot with the time in content vs. grade. X-axis scaled with the data returned.
There is a dot per user per course that looks at their current final grade in the course vs the total time in content for the course.
Keep in count students either without grade or without time in content.

Functional testing:
- No grades and time in content: Returns empty plot;
- Users with null final grade and 0 time in content: Filters out accordingly and does not include in the plot.

Accessibility
- Tested with NVDA, added description for each point on the plot.

Security: n/a
UX: demo with Kevin - done.

@rohitvinnakota @hyehayes @anhill-D2L @johngwilkinson


components/time-in-content-vs-grade-card.js
```diff
@@ -0,0 +1,151 @@
+import 'highcharts';
+import './chart/chart';
+import 'highcharts/modules/accessibility';
```
bemailloux, line 3: It looks like you imported this but you're not using it?

MykolaGalian: Yes, I will use it in next commits

components/time-in-content-vs-grade-card.js
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	get _preparedPlotData() {
+		return this.data.currentFinalGradesVsTimeInContent;
+	}
+
+	render() {
+		// NB: relying on mobx rather than lit-element properties to handle update detection: it will trigger a redraw for
+		// any change to a relevant observed property of the Data object
+		return html`<div class="d2l-insights-time-in-content-vs-grade-container">
```
bemailloux, line 68: Why do you need this extra `div` container? I think you might be able to get the same result by deleting this element and moving all its styles to `:host`.

test/model/data.test.js
```diff
@@ -220,11 +220,18 @@ describe('Data', () => {
 
 	describe('currentFinalGrades', () => {
 		it('should return the current final grades for users', async() => {
-			const expected = [20, 30, 40, 50, 30, 100, 100, 100, 70, 70, 0, 50, 30, 90, 70, 80, 90, 80, 90, 80, 90, 40, 60];
+			const expected = [20, 30, 40, 50, 30, 100, 100, 100, 70, 70, 40, 50, 30, 90, 70, 80, 90, 80, 90, 80, 90, 40, 60];
```
bemailloux, line 223: I think the zero here is important for testing the current final grades card, was there a reason for removing it?

MykolaGalian: Because here is a copy-paste error, already 6 parameters: https://github.com/Brightspace/insights-engagement-dashboard/blob/0c9c1e70ad87c8bc522f1606314b8c20edea5ba4/test/model/data.test.js#L66

components/time-in-content-vs-grade-card.js
```diff
@@ -0,0 +1,151 @@
...
+				}
+			},
+			series: [{
+				data: this._preparedPlotData,
+				marker: {
+					radius: 5
+				},
+				accessibility: {
+					pointDescriptionFormatter: function(point) {
+						return `Grade ${point.y}% for time in content ${point.x} mins.`;
```
Vitalii-Misechko, line 143: Should be this localized?

model/data.js
```diff
@@ -115,6 +116,12 @@ export class Data {
 			.map(record => Math.floor(record[RECORD.CURRENT_FINAL_GRADE] / 10) * 10);
 	}
 
+	get currentFinalGradesVsTimeInContent() {
+		return  this.getRecordsInView()
+			.filter(record => record[RECORD.CURRENT_FINAL_GRADE] !== null && record[RECORD.TIME_IN_CONTENT] !== 0)
```
Vitalii-Misechko, line 121: can be `record[RECORD.CURRENT_FINAL_GRADE]` undefined?

Vitalii-Misechko: Can be `record[RECORD.TIME_IN_CONTENT]` be null or undefined?

MykolaGalian: No

components/time-in-content-vs-grade-card.js
```diff
@@ -140,7 +140,7 @@ class TimeInContentVsGradeCard extends Localizer(MobxLitElement) {
 				},
 				accessibility: {
 					pointDescriptionFormatter: function(point) {
-						return `Grade ${point.y}% for time in content ${point.x} mins.`;
+						return `${point.y},% - ${point.x},mins`;
```
Vitalii-Misechko, line 143: Did you get approval to skip localization for `mins` or you still need localization?

MykolaGalian: Added localized point description

PR #44, opened 2020-09-01
GetRecordsInView uses filtered records instead of serverData

bemailloux:
TODO: 
- [x] make sure it doesn't recalculate records more than necessary

Quality notes
* Testing
  * Using the demo page (`npm start`) it now counts the number of users with overdue assignments based on the filtered selection, instead of using the whole serverData list.

@Vitalii-Misechko @anhill-D2L @rohitvinnakota @hyehayes

### DIFF
--- a/model/data.js
+++ b/model/data.js
@@ -76,8 +76,6 @@ export class Data {
 		this.selectorFilters.semesterIds = semesterIds;
 	}
 
-	// the reason for separating this from getRecordsInView is to try not to reapply the top level filters if
-	// we don't need to.
 	get records() {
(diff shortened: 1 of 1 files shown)


bemailloux:
On the demo page, it only calculates `get records` twice: once on load and once when the serverData arrives. That's good.

It also runs `getRecordsInView` once for each card for each time `get records` runs. It's tempting to memoize that. However, later on when cards are able to filter other cards, we'll need `getRecordsInView` to run once per card anyway, since filters don't filter themselves.

MykolaGalian (approved):

PR #46, opened 2020-09-02
Us118248/ou filter text change

bemailloux:
Changes the OU filter text:
- 0 selected: "Org Unit: All"
- 1 or more selected: "Org Unit: Selections Applied"

Quality notes
* Testing
  * Cases
    * No truncation
      * 0, 1, multiple top-level OUs selected / deselected
      * 0, 1, multiple descendant OUs selected / deselected
      * selection / deselection after searching
      * selection / deselection after semester selected
    * Truncation (there's some edge-casey behaviour here, detailed notes further down)
      * 0, 1, multiple top-level OUs selected / deselected
      * 0, 1, multiple descendant OUs selected / deselected
      * selection / deselection after searching
      * selection / deselection of an OU that disappears when a semester is selected
      * selection / deselection of an OU that is visible before and after a semester is selected
      * selection / deselection of an OU that only appears after a semester is selected
* Security, docs, perf: N/A
* UX
  * Checked with Kevin on wording
  * Localized
  * Screen reader: NVDA changes the text it reads out to match what is shown on-screen

Truncation special cases - let Sem be a semester, OU1 be an OU not in Sem, and OU2 be an OU in Sem
* Case 1
  * For this case, assume OU2 is visible before and after Sem is selected.
  * Select OU1, then select Sem. Displayed: `OU: Selections Applied`, even though no selections appear in the UI. This is ok since the selections are affecting which results are displayed. If it's just OU1 and Sem selected, the set of displayed records will be the empty set.
  * Then select OU2. Displayed: `OU: Selections Applied` (normal)
  * Then deselect OU2: Displayed: `OU: All`, which makes sense given the state of the filter. Why: selecting OU2 caused a reload from server, but since Sem was selected OU1 was no longer in the results. So the filter "forgot" it had OU1 selected.
* Case 2
  * For this case, assume OU2 is only visible after Sem is selected.
  * Select Sem, then select OU2. Displayed: `OU: Selections Applied` (normal)
  * Then unselect sem. Displayed: `OU: Selections Applied`. Why: OU2 was still present in the last server request, so it's affecting the returned results.
  * Then select OU1. Displayed: `OU: Selections Applied` (normal)
  * Then unselect OU1. Displayed: `OU: All`. Why: selecting OU1 caused a server reload but since OU2 didn't exist in the filter it wasn't sent in that server reload, so the filter forgot it had OU2 selected.

@rohitvinnakota @hyehayes @MykolaGalian @anhill-D2L

### DIFF
--- a/components/tree-filter.js
+++ b/components/tree-filter.js
@@ -24,6 +24,7 @@ export class Tree {
 	constructor(tree, selectedIds, leafTypes) {
 		this.tree = tree;
 		this.leafTypes = leafTypes;
+		this.initialSelectedIds = selectedIds;
 		if (selectedIds) {
 			selectedIds.forEach(x => this.setSelected(x, true));
 		}
@@ -124,7 +125,8 @@ decorate(Tree, {
  * This is an opinionated wrapper around d2l-insights-tree-selector which maintains state
  * in the above Tree class.
  * @property {Object} tree - a Tree (defined above)
- * @property {String} name - appears on the dropdown opener
+ * @property {String} openerText - appears on the dropdown opener if no items are selected
+ * @property {String} openerTextSelected - appears on the dropdown opener if one or more items are selected
  * @fires d2l-insights-tree-filter-select - selection has changed; selected property of this element is the list of selected ids
  */
 class TreeFilter extends Localizer(MobxLitElement) {
@@ -132,7 +134,8 @@ class TreeFilter extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
 			tree: { type: Object, attribute: false },
-			name: { type: String },
+			openerText: { type: String, attribute: 'opener-text' },
+			openerTextSelected: { type: String, attribute: 'opener-text-selected' },
 			searchString: { type: String, attribute: 'search-string', reflect: true }
 		};
 	}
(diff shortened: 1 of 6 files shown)


bemailloux:
Screenshots
image
image
image

components/tree-filter.js
```diff
@@ -169,8 +173,15 @@ class TreeFilter extends Localizer(MobxLitElement) {
 	}
 
 	render() {
+		// if selections are applied when loading from server but the selected ids were truncated out of the results,
+		// the visible selections in the UI (this.tree.selected) could be empty even though selections are applied.
+		// In that case, we should indicate to the user that selections are applied, even if they can't see them.
+		const openerText = (this.tree.selected.length || (this.tree.initialSelectedIds && this.tree.initialSelectedIds.length))
```
johngwilkinson, line 179: In the case where there were initial selections then the user clears them, this only works if the consuming component (ou filter, in the current case) forces a redraw. So I think this works for our current use case (where that only occurs due to truncation), but it seems like it won't work if there are initial selected ids but no truncation (as when we set initial filters based on the url).

johngwilkinson (approved):
(note: this PR includes changes from another PR)

components/tree-filter.js
bemailloux: I'm not sure why it wouldn't work - in that case I'm imagining it would initially load using server-side filtering, then when the user clears the filter it would force a reload because a filter has been cleared. So in that case it should start as "Selections applied" then switch to "All" when the filter is cleared.

PR #55, opened 2020-09-14
handle org units that aren't in the parents map

johngwilkinson:
This fixes a potential bug, possibly seen on Spark.

FYI @anhill-D2L @MykolaGalian  @Vitalii-Misechko

### DIFF
--- a/model/orgUnitAncestors.js
+++ b/model/orgUnitAncestors.js
@@ -36,7 +36,8 @@ class OrgUnitAncestors {
 		}
 
 		const ancestorsSet = new Set([orgUnitId]);
-		parentsMap.get(orgUnitId).forEach(parent => {
+		const parentIds = parentsMap.get(orgUnitId) || [];
+		parentIds.forEach(parent => {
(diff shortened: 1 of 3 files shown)


bemailloux (approved):
This could probably do with a unit test as well

johngwilkinson:
> Confirmed using the payload returned by Spark: this code throws without this change, and succeeds with it.
> 
> What remains a mystery is why and how Spark has (apparently) a bunch of org units with "null" parents. Might be a problem with our sql.

Update: it's not null parents, it's parents that are missing from either ORG_UNITPARENTS or ORG_STRUCTURE - which I'm guessing is ones that have been deleted, but are still listed as parents of other org units. Since they aren't in the list of returned org units, they won't appear in the the tree-selector, and I think we'll be fine after this fix.

I'll double-check that the tree-selector itself won't have trouble for the same reason.

johngwilkinson:
Looks like the ou filter/tree-filter won't have this issue, as it always checks that a node exists after looking it up in the map.

rohitvinnakota (approved):

PR #57, opened 2020-09-17
us119016 [Engagement] Interactions - TiC Quad

MykolaGalian:
https://rally1.rallydev.com/#/23525809118d/iterationstatus?sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&detail=/userstory/408910758192

The card takes into account all filters except its own. The card filter is displayed only by painting active and inactive points according to the selected quadrant. Keeping in count students with '0' grade, but removing with null. Avg values for "TiC-Grade" plot changes if records length changes. Handling page reloads is out-of-scope.

ed4

Functional: Local App and **LMS** works as expected with test data.
Accessibility: n/a
Security: n/a
Devops: n/a
UX: demo for Megan - done

@rohitvinnakota @hyehayes @anhill-D2L @johngwilkinson


components/time-in-content-vs-grade-card.js
```diff
@@ -4,6 +4,19 @@ import { css, html } from 'lit-element/lit-element.js';
...
+	LEFT_TOP: 1,
+	RIGHT_TOP: 2,
+	RIGHT_BOTTOM: 3
+};
+
+export const TimeInContentVsGradeCardFilter  = {
+	id: 'd2l-insights-time-in-content-vs-grade-card',
+	title: 'components.insights-time-in-content-vs-grade-card.timeInContentVsGrade',
+	filter: (record) => !record
+};
```
Vitalii-Misechko, line 18: I would define 4 filters. One filter per quadrant. And on click activate/deactivate one filter that corresponds to user selected quadrant. For instance, user clicked on quadrant _3_ then the code could call `data.setApply('d2l-insights-time-in-content-vs-grade-card3', true)`; However, there is drawback `getRecordsInView(id)` can work only with one id, so in this approach `getRecordsInView` should be improved. There is another issue, current implementation of `_preparedPlotData` doesn't pass any `id` to `getRecordsInView`, therefore, this card filters itself what should not happen I believe.

Vitalii-Misechko: Another approach would be to store selected quadrants numbers in `data` and make this data available inside `filter`. In this approach we need to define only one card filter `filter(record): data.selectedQuadrants.includes(getRecordQuadrant(record))`

Vitalii-Misechko: It seems the latter approach requires less efforts

MykolaGalian: I already store quadrants numbers in data this._TiCVsGradeQuadNum. And I passing the filter id  in getRecordsInView inside method setTiCVsGradeCardFilter in Data class

components/applied-filters.js
```diff
@@ -66,10 +66,20 @@ class AppliedFilters extends Localizer(MobxLitElement) {
 		if (event.detail.menuItemKey === clearAllOptionId) {
 			Object.keys(this.data.cardFilters)
 				.forEach(f => this.data.setApplied(f, false));
+
+			this.data.setTiCVsGradesCardSelection(false);
+			this.data.setOverdueAssignmentCardSelected(false);
```
Vitalii-Misechko, line 71: this code just ignores `this.data.setApplied` idea. 
Could you align your solution with existing design idea of `this.data.setApplied`?

components/overdue-assignments-card.js
```diff
@@ -51,6 +51,7 @@ class OverdueAssignmentsCard extends Localizer(MobxLitElement) {
 
 	_valueClickHandler() {
 		this.data.setApplied('d2l-insights-overdue-assignments-card', true);
+		this.data.setOverdueAssignmentCardSelected(true);
```
Vitalii-Misechko, line 54: same

components/applied-filters.js
```diff
@@ -66,10 +66,20 @@ class AppliedFilters extends Localizer(MobxLitElement) {
...
 		}
 
 		this.data.setApplied(event.detail.menuItemKey, event.detail.selected);
+
+		if (event.detail.menuItemKey === 'd2l-insights-time-in-content-vs-grade-card') {
+			this.data.setTiCVsGradesCardSelection(false);
+		}
+		if (event.detail.menuItemKey === 'd2l-insights-overdue-assignments-card') {
+			this.data.setOverdueAssignmentCardSelected(false);
+		}
```
Vitalii-Misechko, line 82: same

components/applied-filters.js
Vitalii-Misechko: `setOverdueAssignmentCardSelected` and `setTiCVsGradesCardSelection` should not exists. `setOverdueAssignmentCardSelected` is already implemented as entry of `this.data.cardFilters`

model/data.js
```diff
@@ -168,30 +174,63 @@ export class Data {
...
+			filter.filter = (record) => record[RECORD.TIME_IN_CONTENT] >= this._avgTimeInContent * 60 && record[RECORD.CURRENT_FINAL_GRADE] <= this._avgGrades;
+		}
+		else (filter.filter = (record) => record);
 	}
 
-	get avgGrade() {
-		const arrayOfGrades = this.currentFinalGradesVsTimeInContent.map(item => item[1]);
-		return arrayOfGrades.length ? Math.floor(arrayOfGrades.reduce((a, b) => a + b, 0) / arrayOfGrades.length) : 0;
+	get _avgTimeInContent() {
+		//average values calculated only from the initial set of data
```
johngwilkinson, line 212: This doesn't seem correct. If, for example, I select a few courses, shouldn't the average be just for those courses?

model/data.js
```diff
@@ -168,30 +174,63 @@ export class Data {
...
 
-	get avgTimeInContent() {
-		const arrayOfTimeInContent =  this.currentFinalGradesVsTimeInContent.map(item => item[0]);
-		return arrayOfTimeInContent.length ? Math.floor(arrayOfTimeInContent.reduce((a, b) => a + b, 0) / arrayOfTimeInContent.length) : 0;
+	setTiCVsGradesCardFilter(quadNum) {
+		this.tiCVsGradesQuadNum = quadNum;
+		const filter = this.cardFilters[TiCVsGradesFilterId];
+
+		if (this.tiCVsGradesQuadNum === QUADRANT.LEFT_BOTTOM) {
+			filter.filter = (record) => record[RECORD.TIME_IN_CONTENT] < this._avgTimeInContent * 60 && record[RECORD.CURRENT_FINAL_GRADE] < this._avgGrades;
```
johngwilkinson, line 197: if final grade is null, these expressions may be incorrect

model/data.js
```diff
@@ -168,30 +174,63 @@ export class Data {
...
 			.map(item => [item[0] !== 0 ? Math.floor(item[0] / 60) : 0, item[1]]) //keep in count students either without grade or without time in content
 			.filter(item => item[0] || item[1]);
 	}
 
-	get avgTimeInContent() {
-		const arrayOfTimeInContent =  this.currentFinalGradesVsTimeInContent.map(item => item[0]);
-		return arrayOfTimeInContent.length ? Math.floor(arrayOfTimeInContent.reduce((a, b) => a + b, 0) / arrayOfTimeInContent.length) : 0;
+	setTiCVsGradesCardFilter(quadNum) {
+		this.tiCVsGradesQuadNum = quadNum;
+		const filter = this.cardFilters[TiCVsGradesFilterId];
```
johngwilkinson, line 194: Small refactoring suggestion: move this code over to the card class: it can get the filter from `data`, and set the filter function itself.

model/data.js
```diff
@@ -168,30 +174,63 @@ export class Data {
 	}
 
 	get currentFinalGrades() {
+		//keep in count students without grade (grade = null), but with time in content
```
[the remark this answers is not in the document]
MykolaGalian: Kevin asked me to keep in count students either without grade or without time in the content. And for consistency during filtering, I added this to the grades chart as well.

components/current-final-grade-card.js
```diff
@@ -160,7 +161,7 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
 				lineWidth: 1,
 				baseSeries: 1,
 				shadow: false,
-				binWidth: 10
+				binWidth: 9.999
```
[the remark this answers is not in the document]
MykolaGalian: No, because I map records 
	`<get currentFinalGrades() {
		//keep in count students without grade (grade = null), but with time in content
		return this.getRecordsInView()
			.map(record => [!record[RECORD.TIME_IN_CONTENT] ? 0 : record[RECORD.TIME_IN_CONTENT], !record[RECORD.CURRENT_FINAL_GRADE] ? 0 : record[RECORD.CURRENT_FINAL_GRADE]])
			.map(item => (item[1] ? Math.floor(item[1] / 10) * 10 : 0))
			.map(item => (item === 100 ? 90 : item)); // put grade 100 in bin 90-100>`

model/data.js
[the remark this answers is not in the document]
MykolaGalian: I will filter all with null grades

model/data.js
MykolaGalian: I found I couldn't do it. Since the filter function must contain average values like this.tiCVsGradesAvgValues ​​so that they can be used inside the Data class. And if I move the logic to the CardFilter, then there I have to refer to the average values ​​as this.`data`.tiCVsGradesAvgValues, which no longer works in the Data class

johngwilkinson: You could get the `data` object in your closure by assigning it to a local, if I understand correctly:
```
const data = this.data;
....filter = x => { ... data.ticVsGradesAvgValues ... }
```

model/data.js
```diff
@@ -168,30 +175,58 @@ export class Data {
...
+		return !(this._recordsLength === this.records.length);
+	}
+
+	get tiCVsGradesAvgValues() {
+		return [this.avgTimeInContent, this.avgGrades];
+	}
+
+	setTiCVsGradesAvgValues() {
+		//avg values will change if records length changes
+		this._recordsLength = this.records.length;
```
johngwilkinson, line 219: `this.records.length` will generally change if the view filters (org unit, semester, role) change, but not if another filter (e.g. overdue assignments) does - but `this.tiCVsGrades` depends on `this.getRecordsInView(TiCVsGradesFilterId)`, which _will_ change if other filters do. So this doesn't seem like the right condition.

To check: is this optimization even necessary? Have you checked performance without it (i.e. just grab the value in a local variable where it is needed)?

components/time-in-content-vs-grade-card.js
```diff
@@ -57,18 +63,78 @@ class TimeInContentVsGradeCard extends Localizer(MobxLitElement) {
...
+	_calculateQuadrant(x, y) {
+		let quadrant;
+		if (x < this._avgTimeInContent && y < this._avgGrades) quadrant = 'leftBottom';
+		else if (x <= this._avgTimeInContent && y >= this._avgGrades) quadrant = 'leftTop';
+		else if (x > this._avgTimeInContent && y > this._avgGrades) quadrant = 'rightTop';
+		else quadrant = 'rightBottom';
+		return quadrant;
+	}
+
+	_coloringAllPointInAmethyst(series) {
```
johngwilkinson, line 106: nit: `colorAllPointsInAmethyst` (and similarly for the others)

johngwilkinson (approved):
Have you done e2e testing with an LMS? The notes just mention test data.

components/time-in-content-vs-grade-card.js
```diff
@@ -57,18 +63,78 @@ class TimeInContentVsGradeCard extends Localizer(MobxLitElement) {
 	}
 
 	get _preparedPlotData() {
-		return this.data.currentFinalGradesVsTimeInContent;
+		return this.data.tiCVsGrades;
+	}
+
+	get _plotDataForLeftBottomQuadrant() {
+		return  this._preparedPlotData.filter(i => i[0] < this._avgTimeInContent && i[1] < this._avgGrades);
```
Vitalii-Misechko, line 70: What do indexes mean? Could you use some constants with meaningful names instead of 0 and 1?

model/cardFilter.js
```diff
@@ -29,6 +32,21 @@ export class CardFilter {
...
+				this.filter = (record) => record[RECORD.TIME_IN_CONTENT] < this.data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] < this.data.tiCVsGradesAvgValues[1];
+			} else if (this.data.tiCVsGradesQuadrant === 'leftTop') {
+				this.filter = (record) => record[RECORD.TIME_IN_CONTENT] <= this.data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] >= this.data.tiCVsGradesAvgValues[1];
+			} else if (this.data.tiCVsGradesQuadrant === 'rightTop') {
+				this.filter = (record) => record[RECORD.TIME_IN_CONTENT] > this.data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] > this.data.tiCVsGradesAvgValues[1];
+			} else if (this.data.tiCVsGradesQuadrant === 'rightBottom') {
+				this.filter = (record) => record[RECORD.TIME_IN_CONTENT] >= this.data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] <= this.data.tiCVsGradesAvgValues[1];
+			} else (this.filter = (record) => record);
+		}
+	}
```
Vitalii-Misechko, line 49: In order to keep `CardFilter` class general you can update line 33 to `return this.filter(record, data);`

And move body of  `_setTiCVsGradesCardFilter` to `time-in-content-vs-grade-card.js`:

```JS
export const TimeInContentVsGradeCardFilter  = {
	id: 'd2l-insights-time-in-content-vs-grade-card',
	title: 'components.insights-time-in-content-vs-grade-card.timeInContentVsGrade',
	filter: (record, data) => {
      let result;
      if (data.tiCVsGradesQuadrant === 'leftBottom') {
				result = record[RECORD.TIME_IN_CONTENT] < data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] < data.tiCVsGradesAvgValues[1];
			} else if (data.tiCVsGradesQuadrant === 'leftTop') {
								result = record[RECORD.TIME_IN_CONTENT] <= data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] >= data.tiCVsGradesAvgValues[1];
			} else if (data.tiCVsGradesQuadrant === 'rightTop') {
								result = record[RECORD.TIME_IN_CONTENT] > data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] > data.tiCVsGradesAvgValues[1];
			} else if (data.tiCVsGradesQuadrant === 'rightBottom') {
								result = record[RECORD.TIME_IN_CONTENT] >= data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] <= data.tiCVsGradesAvgValues[1];
			} else (	result = false );
      return result;
}
```

This way you don't need to introduce `filter: observable` and call `_setTiCVsGradesCardFilter`

Vitalii-Misechko (approved):

PR #58, opened 2020-09-17
US117438: default view; filter ou tree by semester

johngwilkinson:
FYI @anhill-D2L @MykolaGalian @rohitvinnakota 

Quality Notes: see https://github.com/Brightspace/lms/pull/1450

### DIFF
--- a/components/dropdown-filter.js
+++ b/components/dropdown-filter.js
@@ -7,7 +7,7 @@ import { Localizer } from '../locales/localizer';
 
 /**
  * @property {string} name
- * @property {{id: string, displayName: string, _selected: boolean}[]} data
+ * @property {{id: string, displayName: string, selected: boolean}[]} data
  * @property {string[]} selected - returns list of ids for selected items
  * @property {string} filter - current search filter value; only matching items will be visible
  * @property {boolean} disable-search
@@ -47,18 +47,18 @@ class DropdownFilter extends Localizer(LitElement) {
 
 	get selected() {
 		return this.data
-			.filter(item => item._selected)
+			.filter(item => item.selected)
 			.map(item => item.id);
 	}
 
 	get _selectedData() {
-		return this.data.filter(item => item._selected);
+		return this.data.filter(item => item.selected);
 	}
 
 	constructor() {
 		super();
 
-		/** {{id: string, displayName: string, _selected: boolean}[]} */
+		/** {{id: string, displayName: string, selected: boolean}[]} */
 		this.data = [];
 		this.name = '';
 		this.filter = '';
@@ -101,7 +101,7 @@ class DropdownFilter extends Localizer(LitElement) {
 					${this.data.map(item => html`<d2l-filter-dropdown-option
  						text="${item.displayName}"
  						value="${item.id}"
- 						?selected="${item._selected}"
+ 						?selected="${item.selected}"
  						?hidden="${!!this.filter && !item.displayName.toLowerCase().includes(this.filter.toLowerCase())}"
  					></d2l-filter-dropdown-option>`)}
 
@@ -125,7 +125,7 @@ class DropdownFilter extends Localizer(LitElement) {
 	}
 
 	_setSelectedState(id, selected) {
-		this.data = this.data.map(x => (x.id === id ? { ...x, _selected: selected } : x));
+		this.data = this.data.map(x => (x.id === id ? { ...x, selected: selected } : x));
 	}
 
 	_handleSearchedClick(event) {
@@ -133,7 +133,7 @@ class DropdownFilter extends Localizer(LitElement) {
 	}
 
 	_clearSelectionClick() {
-		this.data = this.data.map(x => ({ ...x, _selected: false }));
+		this.data = this.data.map(x => ({ ...x, selected: false }));
 
 		/**
 		 * @event d2l-insights-dropdown-filter-selection-cleared
--- a/components/ou-filter.js
+++ b/components/ou-filter.js
@@ -55,27 +55,44 @@ class OuFilter extends Localizer(MobxLitElement) {
 	}
 
 	get _tree() {
-		// index by id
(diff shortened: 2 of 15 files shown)


model/data.js
```diff
@@ -98,30 +100,42 @@ export class Data {
...
+
+	set selectedOrgUnitIds(newOrgUnitIds) {
 		if (this._selectorFilters.orgUnit.shouldReloadFromServer(newOrgUnitIds)) {
 			this.loadData({ newOrgUnitIds });
 		} else {
 			this._selectorFilters.orgUnit.selected = newOrgUnitIds;
 		}
 	}
 
+	get selectedOrgUnitIds() {
```
johngwilkinson, line 135: This is a small change, but making these properties with getters and setters helped me fix bugs where selections were being calculated in two different ways: the ou filter now looks here.

engagement-dashboard.js
```diff
@@ -101,6 +93,7 @@ class EngagementDashboard extends Localizer(LitElement) {
 					<d2l-insights-semester-filter
 						page-size="10000"
 						?demo="${this.isDemo}"
+						.preSelected="${this._data.selectedSemesterIds}"
```
Vitalii-Misechko, line 96: This change preserves selected semesters between page load. There are no other reason for it, right?

components/dropdown-filter.js
```diff
@@ -47,18 +47,18 @@ class DropdownFilter extends Localizer(LitElement) {
 
 	get selected() {
 		return this.data
-			.filter(item => item._selected)
+			.filter(item => item.selected)
```
Vitalii-Misechko, line 50: Why does `_selected` become "public"?

Vitalii-Misechko: I see `_filterDataWithSelections` uses it to implement code related to new `preSelected` attribute

test/components/semester-filter.test.js
```diff
@@ -61,6 +60,20 @@ describe('d2l-insights-semester-filter', () => {
...
+
+		it('should render with selected items', async() => {
+			const expectedFilterData = [
+				{ id: '10007', name: 'IPSIS Semester New', displayName: 'IPSIS Semester New (Id: 10007)', selected: false },
+				{ id: '121194', name: 'Fall Test Semester', displayName: 'Fall Test Semester (Id: 121194)', selected: true },
+				{ id: '120127', name: 'IPSIS Test Semester 1', displayName: 'IPSIS Test Semester 1 (Id: 120127)', selected: true }
+			];
+
+			const el = await fixture(html`<d2l-insights-semester-filter .preSelected="${[121194, 120127, -1]}"></d2l-insights-semester-filter>`);
+			await new Promise(resolve => setTimeout(resolve, 200)); // allow fetch to run. Add 200 ms for Firefox on OSX
```
Vitalii-Misechko, line 72: I'm wondering about current approach of fetching data inside filters. It seems we could delegate fetching to mobX making UI components only responsible for drawing (like it works with charts).

Vitalii-Misechko (approved):
LGTM!

components/ou-filter.js
```diff
@@ -55,27 +55,44 @@ class OuFilter extends Localizer(MobxLitElement) {
 	}
 
 	get _tree() {
-		// index by id
-		// every object gets an additional three fields: an array of children, a selectedState, and an isOpen boolean
-		// the children array is so we can look up both parents (provided by the server) and children directly
-		const tree = {};
+		// on redraw (e.g. new server data), ensure open nodes stay open so the user doesn't have
+		// to go open a parent again when selecting several children
+		const openNodes = new Set((this.__tree && this.__tree.open) || []);
```
bemailloux, line 60: Not a fan of having `this._tree` and `this.__tree` - too easy to misread or swap one for the other

components/ou-filter.js
```diff
@@ -55,27 +55,44 @@ class OuFilter extends Localizer(MobxLitElement) {
...
-	}
+	_addNodeIfNew(node, filteredTree, fullTree, openNodes) {
+		if (!node || filteredTree[node[ID]]) return;
 
-	_addIdToParents(child, tree) {
-		child[PARENTS].forEach(parentId => tree[parentId] && tree[parentId][CHILDREN].push(child[ID]));
+		filteredTree[node[ID]] = [...node, [], 'none', openNodes.has(node[ID])];
+		node[PARENTS].forEach(parentId => {
+			this._addNodeIfNew(fullTree[parentId], filteredTree, fullTree, openNodes);
+			filteredTree[parentId] && filteredTree[parentId][CHILDREN].push(node[ID]);
```
bemailloux, line 94: `Optional` in these cases, I find it easier to read as 
```
if (filteredTree[parentId]) {
  filteredTree[parentId][CHILDREN].push(node[ID]);
}
```

components/role-filter.js
```diff
@@ -34,7 +33,10 @@ class RoleFilter extends Localizer(LitElement) {
 	}
 
 	get selected() {
-		return this.shadowRoot.querySelector('d2l-insights-dropdown-filter').selected;
+		return this.shadowRoot
+			.querySelector('d2l-insights-dropdown-filter')
+			.selected
+			.map(roleId => Number(roleId));
```
bemailloux, line 39: 👍

engagement-dashboard.js
johngwilkinson: It's for the default view: the server returns a list of selected semesters, and this applies them.

test/components/semester-filter.test.js
johngwilkinson: I think we already have that distinction: dropdown-filter is the renderer, and this class handles the data. I hear what you're saying, though: if everything worked straight from `data`, it would be simpler sometimes. I still like the idea, though, that the semester filter is a drop-in component with no dependencies.

components/ou-filter.js
johngwilkinson: Fixed. This arose from having a private getter, and trying to follow the convention for backing variables. Turned it into a method instead.

PR #62, opened 2020-09-24
Us119394/refactor tree classes

johngwilkinson:
FYI @anhill-D2L @MykolaGalian @Vitalii-Misechko 

This is
1. A rewrite of the Tree class to replace `orgUnitAncestors` and much of the code from the org unit filter
2. A change to how the tree gets built: it now happens on receipt of data from the server, instead on org-unit-filter render.

These changes are important for being able to dynamically add nodes to the tree in a consistent way, which is the next step in dealing with a truncated org-unit tree response.

Quality Notes
* functional
  * default view applies as expected
  * select and deselect leaf and parent nodes with and without also adjusting the semester filter
    * ou filter, table, and cards update as expected
    * server loads happen when expected (on expanding the included org units in any way)
  * select only semester with no courses: ou filter appears empty; add semester with courses and ou filter is populated
    * selected and open states are maintained even when nodes are hidden by the semester filter
  * open nodes stay open across server load (whether it's due to an ou filter change or a semester filter change)
* security: n/a
* perf
  * no significant change
  * checked that `.selected` is not being called for every row
  * checked that the Tree is being built just once per server call
* devops: n/a
* ux/docs: n/a


components/tree-filter.js
```diff
@@ -1,126 +1,257 @@
...
 			selectedIds.forEach(x => this.setSelected(x, true));
 		}
+
+		if (ancestorIds) {
+			this.setAncestorFilter(ancestorIds);
+		}
+	}
+
+	get ids() {
+		return [...this._nodes.values()].map(x => x[ID]);
```
bemailloux, line 59: Wouldn't `this._nodes.keys()` work here?

components/tree-filter.js
```diff
@@ -1,126 +1,257 @@
...
+
+	get open() {
+		return [...this._open];
 	}
 
 	get rootId() {
-		const rootOrgUnit = Object.values(this.tree).find(x => this._isRoot(x));
-		return rootOrgUnit && rootOrgUnit[ID];
+		if (!this._rootId) {
+			this._rootId = this.ids.find(x => this._isRoot(x));
```
bemailloux, line 68: The root is always meant to be the same in spite of mutations? (I assume yes, just checking)

components/tree-filter.js
```diff
@@ -1,126 +1,257 @@
...
 
-	get open() {
-		return Object.values(this.tree).filter(x => x[OPEN]).map(x => x[ID]);
+	getAncestorIds(id) {
+		if (id === 0) return new Set();
+
+		if (!this._ancestors.has(id)) {
+			const ancestors = new Set([
+				id,
+				...this.getParentIds(id).flatMap(x => [...this.getAncestorIds(x)])
```
bemailloux, line 83: Nice, the parents map doesn't have to be pre-built anymore here because it already exists :)

model/selectorFilters.js
```diff
@@ -65,19 +65,23 @@ export class SemesterSelectorFilter {
...
+	constructor({ selectedOrgUnitIds, isRecordsTruncated }, orgUnitTree) {
 		this._latestServerQuery = selectedOrgUnitIds || [];
-		this.selected = defaultViewOrgUnitIds || selectedOrgUnitIds || [];
 		this._isRecordsTruncated = isRecordsTruncated;
-		this._orgUnitAncestors = orgUnitAncestors;
+		this._orgUnitTree = orgUnitTree;
+	}
+
+	get selected() {
+		return (this._orgUnitTree && this._orgUnitTree.selected) || [];
```
bemailloux, line 75: Would be easier to read as 
```
if (this._orgUnitTree) {
  return this_orgUnitTree.selected;
}
return [];
```

test/components/tree-filter.test.js
```diff
@@ -2,24 +2,352 @@ import { expect, fixture, html, oneEvent } from '@open-wc/testing';
...
+
+		it.skip('should return expected selection given multiple parents of some nodes', () => {
+			// NOTE
+			// This test would fail, because the algorithm walks down the tree through indeterminate nodes until
+			// it finds all explicitly selected descendants - but this means if there is more than one
+			// path to a node, it can appear twice or it can appear along with an ancestor.
+			// This has only very minor performance implications, so for simplicity this behaviour
+			// is allowed.
+			const tree = new Tree({ nodes, invisibleTypes, selectedIds: [111, 1002] });
+			expect(tree.selected.sort()).to.deep.equal([1002, 111]);
```
bemailloux, line 121: So what is actually selected in this case?

test/components/tree-filter.test.js
```diff
@@ -2,24 +2,352 @@ import { expect, fixture, html, oneEvent } from '@open-wc/testing';
...
+		it('should return true for a node that is not a leaf type', () => {
+			expect(sut.isOpenable(1)).to.be.true;
+		});
+
+		it('should return false for a node that is a leaf type', () => {
+			expect(sut.isOpenable(111)).to.be.false;
+		});
+
+		it('should return some boolean if the node is unknown (no error)', () => {
+			expect(sut.isOpenable(2352235)).to.be.a('Boolean');
```
bemailloux, line 322: Huh - we don't know if it would be true or false in this case?

test/components/tree-filter.test.js
johngwilkinson: IIRC it's `[1002, 111, 3]`, because 3 is in two departments, one of which is fully selected (1002), and the other of which is partly selected (1003).

test/components/tree-filter.test.js
johngwilkinson: I know, but I'd be fine with either. It's just to make corner case crashes less likely.

bemailloux: I'd prefer for there to be a defined default behaviour... but I guess this is unlikely to happen anyway.

rohitvinnakota (approved):

bemailloux (approved):

PR #65, opened 2020-09-25
us119015 [Engagement] Interactions - Grades

MykolaGalian:
https://rally1.rallydev.com/#/23525809118d/iterationstatus?sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&detail=/userstory/408910755472

AC:
Clicking on a bar in the grade bar chart filters the returned data for just the users and courses where the user has a grade in that segment (0-10, 10-20 and so on)
The bar interacted with stays coloured and the others are greyed out.
Summarize the interaction in button "Current Grade: [x] - [x]" and filter applied with the ability to "X" out to remove and revert back.

gr3

Functional: LMS  and Local App works as expected with test data.
Accessibility: n/a
Security: n/a
Devops: n/a
UX: demo for Megan - done.

@rohitvinnakota @hyehayes @anhill-D2L @johngwilkinson @bemailloux


components/current-final-grade-card.js
```diff
@@ -2,6 +2,21 @@ import { css, html } from 'lit-element/lit-element.js';
...
 import { MobxLitElement } from '@adobe/lit-mobx';
+import { RECORD } from '../model/data';
+
+export const CurrentFinalGradeCardFilter  = {
+	id: 'd2l-insights-current-final-grade-card',
+	title: 'components.insights-current-final-grade-card.currentGrade',
+	filter: (record, data) => {
+		let result;
+		if (data.gradesCategory === 90) {
+			result = record[RECORD.CURRENT_FINAL_GRADE] !== null && record[RECORD.CURRENT_FINAL_GRADE] >= data.gradesCategory && record[RECORD.CURRENT_FINAL_GRADE] <= (data.gradesCategory + 10);
```
Vitalii-Misechko, line 13: It seems you don't need this part `record[RECORD.CURRENT_FINAL_GRADE] !== null`

components/current-final-grade-card.js
```diff
@@ -63,6 +78,44 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
+
+	get isApplied() {
+		return this.data.cardFilters['d2l-insights-current-final-grade-card'].isApplied;
+	}
+
+	_valueClickHandler() {
+		this.data.setApplied('d2l-insights-current-final-grade-card', true);
+	}
+
+	_colorNonSelectedPointsInMica(data) {
```
Vitalii-Misechko, line 97: Could you rename variable `data` it can be easily confused with `this.data`?
`data` => `seriesData`

components/current-final-grade-card.js
```diff
@@ -63,6 +78,44 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
+		data.forEach(point => {
+			if (this.category !== Math.ceil(point.x)) {
+				point.update({ color: 'var(--d2l-color-mica)' });
+			}
+		});
+	}
+
+	_colorNonSelectedPointsInMicaAfterRender(data) {
```
Vitalii-Misechko, line 105: data => seriesData

components/current-final-grade-card.js
```diff
@@ -63,6 +78,44 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
+		data.forEach(data => {
+			if (Math.ceil(data.category) !== this.category) {
+				data.update({ color: 'var(--d2l-color-mica)' }, false);
+			}
+		});
+	}
+
+	_colorAllPointsInAmethystAfterRender(data) {
```
Vitalii-Misechko, line 113: data => seriesData

components/current-final-grade-card.js
```diff
@@ -85,9 +138,22 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
+
 		return {
 			chart: {
-				height: 230
+				height: 230,
+				events: {
+					render: function() {
+						if (that.isApplied) {
+							that._colorNonSelectedPointsInMicaAfterRender(this.series[0].data);
+							this.render();
```
Vitalii-Misechko, line 150: It seems this should cause infinite loop. Have you checked brawser's developer console for errors?

components/current-final-grade-card.js
```diff
@@ -156,12 +238,11 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
 			},
 			series: [{
 				type: 'histogram',
-				color: 'var(--d2l-color-amethyst)',
 				animation: false,
 				lineWidth: 1,
 				baseSeries: 1,
 				shadow: false,
-				binWidth: 9.999
+				binWidth: 9.9999
```
Vitalii-Misechko, line 245: Could you explain why we need fractional digit here? Is it possible to draw 9.9999 px on the screen?
Have you tried to replace `binWidth` with `binsNumber`? https://api.highcharts.com/highcharts/plotOptions.histogram.binWidth

MykolaGalian: This is done in order to correctly indicate the width of the bin. So in this case, the width of the values ​​(not pixels) for each bin is set like this 0-9.999, 9.99-19.99, ..., 89.99-99.99. Since we have whole sets of values, for example, 10, 20, 20, 70, then all values ​​definitely fall into the correct bins

components/current-final-grade-card.js
MykolaGalian: Yes, I checked, everything works without errors, the video above confirms this

components/current-final-grade-card.js
Vitalii-Misechko: Oh...I see. TIL: `binWidth` - is statistical thing

components/current-final-grade-card.js
MykolaGalian: Corrected not to be misleading

Vitalii-Misechko: I'm glad to hear that. I'm concerned because `render` function calls `render`. I've checked out this branch and when I click clear all and then a bar my browser doesn't allow me click unselected bar for a while (UI hangs). Could you confirm/negate this behavior?

MykolaGalian: Now all going well
dd1

Vitalii-Misechko: Hung

Vitalii-Misechko: It seems my local stuff

johngwilkinson:
> Users then choose to also click on another be to be added to the filter which will re-colour the bar and update the results.

This a.c. seems to mean the user can select multiple bars. But the demo at https://xd.adobe.com/view/cf5c5823-d1c3-47eb-45b6-8297bf92f944-b96d/screen/e6aa869d-2268-4f7f-a405-1d3c9f9a879b/ shows different behaviour (a single click selects all to one side of the click). Have you clarified this with Megan and/or Kevin?

components/current-final-grade-card.js
```diff
@@ -85,9 +138,22 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
 	get chartOptions() {
+		const that = this;
+
 		return {
 			chart: {
-				height: 230
+				height: 230,
+				events: {
+					render: function() {
+						if (that.isApplied) {
```
johngwilkinson, line 148: If I understand right, the reason you set the colours here - instead of in the series data as normal - is that the series data is generated by highcharts, so it doesn't exist before this point. A comment explaining that would be useful here.

MykolaGalian: Ok, after redrawing the chart as a result of updating (for example, when the user disable any of the filters), we need to keep the color of the selected/nonselected  bars

MykolaGalian:
> > Users then choose to also click on another be to be added to the filter which will re-colour the bar and update the results.
> 
> This a.c. seems to mean the user can select multiple bars. But the demo at https://xd.adobe.com/view/cf5c5823-d1c3-47eb-45b6-8297bf92f944-b96d/screen/e6aa869d-2268-4f7f-a405-1d3c9f9a879b/ shows different behaviour (a single click selects all to one side of the click). Have you clarified this with Megan and/or Kevin?

Yes, I missed it. Megan said we need multi-select option for bars only for Grades. Will add this functional

components/current-final-grade-card.js
```diff
@@ -1,7 +1,32 @@
...
+	filter: (record, data) => {
+		let result;
+		const category = Array.from(data.gradesCategory).slice().sort();
+		// eslint-disable-next-line no-unused-vars
+		const rec = record[RECORD.CURRENT_FINAL_GRADE];
+		let str = [];
+		if (!data.gradesCategory.has(90)) {
+			category.forEach(cat => str.push(`(rec >= ${cat} && rec < ${cat} + 10) ||`));
+			const strFilter = str.join('').slice(0, -3);  // remove last "OR" statement
+			result = eval(strFilter);
```
Vitalii-Misechko, line 20: `eval` is [evil](https://stackoverflow.com/questions/197769/when-is-javascripts-eval-not-evil).  In this part of code, my concern is performance. I believe that code can be rewritten without it

components/current-final-grade-card.js
```diff
@@ -1,7 +1,32 @@
...
+		} else {
+			const categoryArray = category.filter(i => i !== 90);
+			categoryArray.forEach(cat => str.push(`(rec >= ${cat} && rec < ${cat} + 10) ||`));
+			const strFilter = str.join('').concat(' (rec >= 90 && rec <= 100)');
+			result = eval(strFilter);
```
Vitalii-Misechko, line 25: the same

components/current-final-grade-card.js
```diff
@@ -170,6 +249,24 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
 								val = point.y;
 							return `${ix - 10} to ${ix}, ${val}.`;
 						}
+					},
+					allowPointSelect: true,
+					color: 'var(--d2l-color-amethyst)',
+					states: {
+						select: {
+							color: 'var(--d2l-color-amethyst)'
```
Vitalii-Misechko, line 257: I'm wondering why we need custom color renders when highchart has these option?  `color: 'var(--d2l-color-mica)' should make the applying color code unnecessary

Vitalii-Misechko:
There is not indication of what is selected now in `Applied filter`-panel. Should we change label of `Current Grade` button or Kevin approved the current way?

image

components/current-final-grade-card.js
```diff
@@ -170,6 +249,24 @@ class CurrentFinalGradeCard extends Localizer(MobxLitElement) {
...
+						}
+					},
+					point: {
+						events: {
+							select: function() {
+								const selectedPoints = this.series.chart.getSelectedPoints();
+								selectedPoints.forEach(point => that.addToCategory(Math.ceil(Number(point.category))));
```
Vitalii-Misechko, line 264: Does this line cause multiple application state recalculation? Should we figure out all categories before calling `addToCategory`? Just wondering, how much it takes to re-calulate app state once?

MykolaGalian:
> 
> 
> There is not indication of what is selected now in `Applied filter`-panel. Should we change label of `Current Grade` button or Kevin approved the current way?
> 
> image

Megan said that this is functionality for a separate story.

components/current-final-grade-card.js
MykolaGalian: By default, for the point select event, the color is #cccccc (https://api.highcharts.com/highcharts/plotOptions.column.states.select.color). We need to re-assign it to blue, since recoloring the dots, in this case, does not help, the selection still leaves the color #cccccc

Vitalii-Misechko: Could you examine this exapmle http://jsfiddle.net/8truG/ ? It does not require any aditional JS code

components/current-final-grade-card.js
```diff
@@ -2,6 +2,30 @@ import { css, html } from 'lit-element/lit-element.js';
...
+import { RECORD } from '../model/data';
+
+function or(pred1, pred2) {
+	return el => pred1(el) || pred2(el);
+}
+
+export const CurrentFinalGradeCardFilter  = {
+	id: 'd2l-insights-current-final-grade-card',
+	title: 'components.insights-current-final-grade-card.currentGrade',
+	filter: (record, data) => {
```
johngwilkinson, line 14: The goal here is for the filter to pass the record iff it is in one of the enabled categories, right? This can be expressed pretty clearly using `Array.some`:
```
filter: (record, data) => {
  const grade = record[RECORD.CURRENT_FINAL_GRADE];
  return [...data.gradesCategory].some(category =>
    category === 90 ?
      (grade >= 90) :
      (grade >= category && grade < (category + 10))
  );
}
```

But I think there's an even simpler way that would have the category-calculation code exist in only one place (which is good!): refactor `data.currentFinalGrades` to use a helper function `gradeCategory(grade)` that returns the bin number (with 100 mapped to 90, nulls mapped to null). Then this code is just
```
filter: (record, data) => data.gradesCategory.has(gradeCategory(record[RECORD.CURRENT_FINAL_GRADE]));
```

model/data.js
```diff
@@ -38,6 +39,9 @@ export class Data {
 		this._userDictionary = null;
 
 		// @observables
+		this.gradesCategory = new Set();
```
johngwilkinson, line 42: naming: `selectedGradesCategories` would be better

model/data.js
```diff
@@ -38,6 +39,9 @@ export class Data {
...
+		this.gradesCategory.add(0);
```
johngwilkinson, line 43: This is confusing. It would be better to write your code so that an empty set matches all records (or so the filter is not applied by default).

model/data.js
```diff
@@ -169,14 +173,22 @@ export class Data {
...
-		return this.getRecordsInView()
+		return this.getRecordsInView(CurrentFinalGradeFilterId)
 			.filter(record => record[RECORD.CURRENT_FINAL_GRADE] !== null && record[RECORD.CURRENT_FINAL_GRADE] !== undefined)
 			.map(record => [record[RECORD.TIME_IN_CONTENT], record[RECORD.CURRENT_FINAL_GRADE]])
 			.filter(item => item[0] || item[1])
 			.map(item => (item[1] ? Math.floor(item[1] / 10) * 10 : 0))
 			.map(item => (item === 100 ? 90 : item)); // put grade 100 in bin 90-100
 	}
 
+	setGradesCategoryEmpty() {
```
johngwilkinson, line 184: I believe these two methods should be marked with mobx `action`, although it might not matter.

johngwilkinson (commented):
Looks like this could use some unit tests.

components/current-final-grade-card.js
MykolaGalian: After rendering the chart as a result of updating (for example, when the user disables any of the filters) or after select other bars, we need to keep the color of the bars that were selected/nonselected before that. In your example, everything works as there are no chart rendering (updating) events

Vitalii-Misechko: That's correct! My proposal it to keep events(manage selected bars) and get rid of the code that applies colors because highcharts already applies colors correctly

MykolaGalian: I tried programmatically control not the colors, but the selected properties for the points. But we have other events (update, render) when the filter is applied, and it's more difficult programmatically set this (the property selected is not set the first time). Also, we will still need to work with the color of the points, because unselected points will need to be painted in gray.

Vitalii-Misechko: Discussed. The main problem that leads to this solution with custom points rendering is the inability to provide into highcharts preselected points between chart rendering

Vitalii-Misechko (approved):

johngwilkinson (approved):

PR #67, opened 2020-09-25
Us119394/load ou children

johngwilkinson:
Quality Notes
* functional (all impersonating Power User)
  * NB: search does not fully work, as it only includes data that is already available client-side
  * with CourseLimit config in LMS set to 5 (instead of default 5000)
    * default view as expected
    * dept containing default selection appears partially selected
    * navigate into unselected dept
    * navigate into selected dept
    * select and deselect dept
      * note that with truncation at only 5 courses, the default selection can become hidden if the user adds another selection without first expanding the department(s) containing default selections: the client doesn't have enough information to mark the default department(s) partially selected because the selected courses are left out of the response . So long as the truncation limit is significantly larger than the number of default courses, this won't happen
    * select and deselect courses within a department
    * confirm that server response includes five of the selected ou ids, whether they are in one department, two, or three
    * use case: deselect semester filter and default courses, then select the department with test final grades data; data appears in graphs
  * with CourseLimit set to default (no truncation regression testing)
* performance
  * many more server hits when truncated, but each is very quick
  * confirmed ou tree data structure is only rebuilt on load of main data from server (not every time children are added)
  * without lms client cache of children, there is noticeable lag and jumpiness when selecting courses; with cache, this is not noticeable
* devops: n/a
* security: n/a
* ux/docs
  * NB: there's no loading spinner/gray-out; it might be desirable to add this in another PR
  * alerting the user about truncation is out-of-scope
  * tree navigation and selection/deselection works the same whether or not there's ou truncation (so no specifically UX demo - functionality to be demoed to team)

FYI: @anhill-D2L @MykolaGalian @rohitvinnakota

### DIFF
--- a/components/tree-filter.js
+++ b/components/tree-filter.js
@@ -25,6 +25,8 @@ export class Tree {
 	 * @param {Tree} [oldTree] - tree to copy previous state from (e.g. which nodes are open)
 	 * @param {Boolean} isDynamic - if true, the tree is assumed to be incomplete, and tree-filter will fire events as needed
 	 * to request children
+	 * @param {Map}[extraChildren] - Map from parent node ids to arrays of nodes; these will be added to the tree before
+	 * any selections are applied and the parents marked as populated. Useful for adding cached lookups to a dynamic tree.
 	 */
 	constructor({
 		nodes = [],
@@ -33,7 +35,8 @@ export class Tree {
 		selectedIds,
 		ancestorIds,
 		oldTree,
-		isDynamic = false
+		isDynamic = false,
+		extraChildren
 	}) {
 		this.leafTypes = leafTypes;
 		this.invisibleTypes = invisibleTypes;
@@ -58,6 +61,12 @@ export class Tree {
 			});
 		});
 
+		if (extraChildren) {
+			extraChildren.forEach((children, orgUnitId) => {
+				this.addNodes(orgUnitId, children);
+			});
+		}
+
 		if (selectedIds) {
 			selectedIds.forEach(x => this.setSelected(x, true));
 		}
@@ -69,6 +78,10 @@ export class Tree {
 
 	get ids() {
 		return [...this._nodes.values()].map(x => x[ID]);
+	}
+
+	get isDynamic() {
+		return !!this._populated;
 	}
 
 	get open() {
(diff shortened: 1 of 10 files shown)


components/tree-filter.js
```diff
@@ -59,6 +80,10 @@ export class Tree {
 		return [...this._nodes.values()].map(x => x[ID]);
 	}
 
+	get isDynamic() {
+		return !!this._populated;
```
bemailloux, line 84: Can a dynamic tree ever switch to a static tree if all the nodes were loaded in one-by-one? (I don't think this would be necessary, just wondering)

components/tree-filter.js
```diff
@@ -74,6 +99,50 @@ export class Tree {
...
+		if (this.getState(parentId) === 'explicit') {
+			newChildren.forEach(x => this._state.set(x[ID], 'explicit'));
+		}
+
+		// Ancestors may need updating: if one or more of newChildren was already present (due to
+		// being added under another parent), then they may also have been opened and have descendants,
+		// which now need a new ancestor.
+		// For simplicity and correctness, we simply reset the ancestors map, which will be
+		// regenerated as needed by getAncestorIds.
+		this._ancestors = new Map();
```
bemailloux, line 139: Reloading the ancestors map every time is ok for perf, right? Since it only has to get ancestor ids for maybe 4-5 ancestors max in the chain? (maybe double that if a course offering has multiple parents?)

components/tree-filter.js
johngwilkinson: It could be coded to know if every non-leaf node has been expanded then switch to static, but the likelihood of this actually happening is very small, so I don't think it's worth adding any code for.

components/tree-filter.js
johngwilkinson: It's not the best for perf, but calculating ancestors happens on every full load anyway, and we find that performance acceptable. I haven't noticed any lagginess. So, as it say, this is a choice in favour of simplicity and correctness, as opposed to writing code to identify and update just the relevant parts of the tree.

bemailloux (approved):

model/lms.js
```diff
@@ -55,3 +56,26 @@ export async function fetchSemesters(pageSize, bookmark, search) {
 	const response = await fetch(url.toString());
 	return await response.json();
 }
+
+const relevantChildrenCache = new Map();
```
Vitalii-Misechko, line 60: Just curious. Do we have a plan to store cache longer than page exists (session long, predefined amount of time)?

Vitalii-Misechko (approved):
LGTM!

model/lms.js
johngwilkinson: No plan, no. When the page is reloaded, even if the url contains filter settings, the ou tree will have all nodes closed, so there's no real need. The cache is mostly there to prevent blinking/jumping when a node is selected: the performance benefit is minor.

PR #68, opened 2020-09-28
[Engagement] Interactions - Course Access

MykolaGalian:
https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F408910758796&sharedViewId=328666453148

Clicking on a bar in the course access bar chart filters the returned data for just the users and courses where the user has a course access in that segment (never, >1 day, 1 - 3 days ago, and so on).
The bar interacted with stays coloured and the others are greyed out.
Summarize the interaction in button "Course Access: [x - x days ago, Never]" and filter applied with the ability to "X" out to remove and revert back.
A multi-select option for bars does not need here.

ez

Accessibility: n/a
Security: n/a
Devops: n/a
UX: demo for Megan - done


johngwilkinson:
This one also has this A.C.:
> Users then choose to also click on another be to be added to the filter which will re-colour the bar and update the results.

components/course-last-access-card.js
```diff
@@ -2,6 +2,34 @@ import { css, html } from 'lit-element/lit-element.js';
...
+	id: 'd2l-insights-course-last-access-card',
+	title: 'components.insights-course-last-access-card.courseAccess',
+	filter: (record, data) => {
+		const fourteenDayMillis = 1209600000;
+		const sevenDayMillis = 604800000;
+		const fiveDayMillis = 432000000;
+		const oneDayMillis = 86400000;
+		let result;
+		const date = record[RECORD.COURSE_LAST_ACCESS] === null ? -1 : (Date.now() - record[RECORD.COURSE_LAST_ACCESS]);
+		if (data.lastAccessCategory === 'Never') {
```
johngwilkinson, line 17: The logic here duplicates that in `data._bucketCourseLastAccessDates`: can you refactor so that it is only in one place?

MykolaGalian:
> 
> 
> This one also has this A.C.:
> 
> > Users then choose to also click on another be to be added to the filter which will re-colour the bar and update the results.

Megan said that a multi-select option for bars does not need for this story.

PR #72, opened 2020-10-01
[Engagement] Interactions - Course Access

MykolaGalian:
https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F408910758796&sharedViewId=328666453148

Clicking on a bar in the course access bar chart filters the returned data for just the users and courses where the user has a course access in that segment (never, >1 day, 1 - 3 days ago, and so on).
The bar interacted with stays coloured and the others are greyed out.
Summarize the interaction in button "Course Access: [x - x days ago, Never]" and filter applied with the ability to "X" out to remove and revert back.
a1
Accessibility: n/a
Security: n/a
Devops: n/a
UX: demo for Megan - done


johngwilkinson:
There are some conflicts.

model/data.js
```diff
@@ -277,14 +287,14 @@ export class Data {
 	_persist() {
 		localStorage.setItem('d2l-insights-engagement-dashboard.state', JSON.stringify(
 			Object.keys(this.cardFilters)
-				.map(f => ({ id: f, applied: this.cardFilters[f].isApplied }))
```
johngwilkinson, line 280: Can you just leave a comment that this is intentionally disabled? Someone might be confused about how this was expected to work.

components/course-last-access-card.js
```diff
@@ -123,6 +174,22 @@ class CourseLastAccessCard extends Localizer(MobxLitElement) {
...
 				height: '250px',
+				events: {
+					render: function() {
+						//after redrawing the chart as a result of updating (for example, when the user disable any of the filters),
+						// we need to keep the color of the selected/nonselected bars
+						if (that.isApplied) {
+							that._colorNonSelectedPointsInMica(this.series[0].data);
+							this.render(false);
+							that._colorSelectedPointsInCelestine(this.series[0].data);
+							this.render(false);
```
bemailloux, line 185: Would rendering once at the end work here? Like 
```
if (that.isApplied) {
  // set colors
} else {
  // set other colors
}
this.render(false);
```

components/course-last-access-card.js
```diff
@@ -108,6 +117,48 @@ class CourseLastAccessCard extends Localizer(MobxLitElement) {
...
+
+	_colorSelectedPointsInCelestine(seriesData) {
+		seriesData.forEach(point => {
+			if (this.category.has(point.index)) {
+				point.update({ color: 'var(--d2l-color-celestine)' }, false);
+			}
+		});
+	}
+
+	_colorAllPointsInCelestine(seriesData) {
```
bemailloux, line 156: I think these functions are too specific. It might be better for code reuse if you split them instead into 
```
_applyColor(points, color) {
  points.forEach(point => point.update(color))
}

_colorPointsBySelection(allPoints, someSelected) {
  if (someSelected) {
    //get selected points
    // apply selected color
    // get unselected points
    // apply unselected color
  } else {
    // apply selected color to all points
  } 
  this.render(false);
}
```

components/course-last-access-card.js
```diff
@@ -202,8 +269,25 @@ class CourseLastAccessCard extends Localizer(MobxLitElement) {
...
+							color: 'var(--d2l-color-celestine)'
+						}
+					},
+					point: {
+						events: {
+							select: function() {
+								that.addToCategory(this.index);
+								that._valueClickHandler();
+								that._colorNonSelectedPointsInMica(this.series.data);
+								that._colorSelectedPointsInCelestine(this.series.data);
```
bemailloux, line 286: Why is `render` not required here?

MykolaGalian: Look like the render is needed only for the global event (for the entire chart)

components/course-last-access-card.js
MykolaGalian: I tested everything once again. When the bar is selected for the first time, in addition to the "select" event, the "render" event is also triggered (with the next selections, it will no longer be triggered), and all unselected points become gray. Then in the next selection we do not need to paint the unselected points in gray once again (additional performance loss). So inside the "select" event, we paint only selected points.

johngwilkinson (approved):

bemailloux (approved):

PR #73, opened 2020-10-02
TiC v Grades Hover Text

rohitvinnakota:
[US118263](https://rally1.rallydev.com/#/detail/userstory/402807206872?fdp=true): [Engagement][Cards] TiC v Grades Hover Text

image

After a discussion with Kevin, it was agreed that we want the arrow of the tooltip to point to the midpoint of the quadrant. The tradeoff here is that the tooltip will not pop up if we hover at the edge of the chart, otherwise it works as expected. 

**Quality Notes**
Functional testing
Tested the following cases
-> No data in quadrant: Shows "0 users have x grade.."
-> Tooltip updates with filters

Accessibility
->No effect on accessibility 

- [x] Perf(): make query more optimal 
UX: Went over design with Kevin
Security: n/a

fyi: @bemailloux @hyehayes @Vitalii-Misechko @anhill-D2L

### DIFF
--- a/components/time-in-content-vs-grade-card.js
+++ b/components/time-in-content-vs-grade-card.js
@@ -118,15 +118,6 @@ class TimeInContentVsGradeCard extends Localizer(MobxLitElement) {
 			[(maxTimeInContent + this._avgTimeInContent) / 2, 75]];
 	}
 
-	get _plotDataLengths() {
-		return [
-			this._plotDataForLeftBottomQuadrant.length,
-			this._plotDataForLeftTopQuadrant.length,
-			this._plotDataForRightTopQuadrant.length,
-			this._plotDataForRightBottomQuadrant.length
-		];
-	}
-
 	_setQuadrant(quadrant) {
 		this.data.setTiCVsGradesQuadrant(quadrant);
 	}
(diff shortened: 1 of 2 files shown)


johngwilkinson:
I don't see how this is as useful as having the tooltip appear for each point, giving the data for the point (and maybe also the user and course)... but maybe that's just me.

When you say it won't appear "if we hover at the edge of the chart", do you mean just hovering over an axis doesn't produce a tooltip? Or does the user have to put the mouse right in the middle of a quadrant to get a tooltip?

rohitvinnakota:
> When you say it won't appear "if we hover at the edge of the chart", do you mean just hovering over an axis doesn't produce a tooltip? Or does the user have to put the mouse right in the middle of a quadrant to get a tooltip?

I think for charts with a lot of points, having a tooltip for each point is not as useful and a little confusing but an overall summary might be. Hovering over the axis, and a little inside the outer edges of the chart(around 20%) won't produce a tooltip, but it will in and around the middle.

image

Hovering between the blue lines on the chart produces the tooltip as of now

johngwilkinson (approved):

components/time-in-content-vs-grade-card.js
```diff
@@ -189,7 +204,35 @@ class TimeInContentVsGradeCard extends Localizer(MobxLitElement) {
...
 			animation: false,
-			tooltip: { enabled: false },
+			tooltip: {
+				formatter: function() {
+					if (this.series.name === 'midPoint') {
+						const midPoints = that._dataMidPoints;
+						const currentMidPoint = [this.x, this.y];
+						if (currentMidPoint.toString() === midPoints[0].toString()) {
+							return that._toolTipTextByQuadrant(this.series.chart.series[0].name, this.series.chart.series[0].data.length);
+						}
```
MykolaGalian, line 214: I think you can try using the shorter path: `this.series.chart.series [3] .name -> this.series [3].name`

MykolaGalian (approved):

PR #77, opened 2020-10-05
Us119394/truncated search

johngwilkinson:
(depends on an LMS PR coming soon)

FYI @anhill-D2L @MykolaGalian @rohitvinnakota 

Quality Notes

* functional (with added delay for loading-state testing)
  * spinner appears on first load before full set of root's children arrive
  * drill down the tree into nodes that already have all or some children: spinner appears during load, nodes are not duplicated
  * drill down the tree into nodes that don't yet have any children: spinner appears, and children appear when loaded
  * close and reopen: no extra network call, no spinner
  * select a new child: selection is applied and server data loaded; ou filter does not jump or blink in and out
  * search ('chem' on dev lms): shows known children and spinner; shows first batch of children when available, along with load more button
  * click search load more: spinner appears, results load; no further load more button if no more results
  * with search open, adjust semester filter so there are no hits and go back to ou filter: search results are empty
  * clear semester filter and do same search again: new results loaded
* performance
  * client-side perf should not be an issue with search loading 10 items per call
* security: n/a
* devops: n/a
* ux/docs:
  * Kevin is looking at desired loading-indicator and load-more behaviour; this design may end up as a separate story, and he is okay with this for now
  * Kevin approved the trade-off of relying on page scroll instead of component scroll to avoid jumps on every load

image

image

image


johngwilkinson:
heads-up: one small change will be coming, to not ask the server for results when the search field is cleared

components/ou-filter.js
```diff
@@ -55,9 +56,18 @@ class OuFilter extends Localizer(MobxLitElement) {
...
 	async _onRequestChildren(event) {
+		const el = event.target;
 		const id = event.detail.id;
 		const children = await fetchRelevantChildren(id, this.data.selectedSemesterIds);
-		this.data.orgUnitTree.addNodes(id, children);
+		el.addChildren(id, children);
+	}
+
+	async _onSearch(event) {
+		const el = event.target;
```
Vitalii-Misechko, line 66: TIL: nice way to use child's component API

components/tree-filter.js
```diff
@@ -464,6 +538,24 @@ class TreeFilter extends Localizer(MobxLitElement) {
...
+		if (!this._isSearch) return html``;
+
+		if (this._isLoadingSearch) {
+			return html`<d2l-loading-spinner slot="search-results"></d2l-loading-spinner>`;
+		}
+
+		if (this.isLoadMoreSearch)  {
+			return html`<d2l-button slot="search-results"
+				@click="${this._onSearchLoadMore}"
+			>Load More</d2l-button>`;
```
Vitalii-Misechko, line 555: minor: `Load More`  should be localized

components/tree-filter.js
```diff
@@ -492,6 +584,33 @@ class TreeFilter extends Localizer(MobxLitElement) {
...
+		this._isLoadingSearch = true;
+
+		/**
+		 * @event d2l-insights-tree-filter-search
+		 */
+		this.dispatchEvent(new CustomEvent(
+			'd2l-insights-tree-filter-search',
+			{
+				bubbles: true,
+				composed: false,
```
Vitalii-Misechko, line 610: `nit`: `composed: false` is default. It seems we can just omit it

model/lms.js
```diff
@@ -79,3 +80,40 @@ export async function fetchRelevantChildren(orgUnitId, selectedSemesterIds) {
...
+		url.searchParams.set('selectedSemestersCsv', selectedSemesterIds.join(','));
+	}
+	if (bookmark) {
+		url.searchParams.set('bookmark', bookmark);
+	}
+	const response = await fetch(url.toString());
+	const results = await response.json();
+
+	const key = cacheKey(selectedSemesterIds);
+	if (orgUnitSearchCache.searchString === searchString && orgUnitSearchCache.selectedSemesterIds === key) {
```
Vitalii-Misechko, line 102: it seems that `searchString` is the cache key as well but the code stores it in a separate field. Why is that?

Vitalii-Misechko (approved):
LGTM!

bemailloux:
I wanted to play around with this locally and was wondering why it wasn't working... then remembered `(depends on an LMS PR coming soon)`

model/lms.js
johngwilkinson: Because the search string is used when writing the cache but not when reading it:

When a reload happens during search (e.g. when the user selects one of the results), Data uses this cache to add the search results. It needs to do this only if the semester filter matches (changing that filter also causes a reload, but we need new search results then). But Data doesn't know the search string, and it is safe to assume it hasn't changed. On the other hand, we need to invalidate the cache if the user starts a new search.

components/tree-selector.js
```diff
@@ -68,10 +68,13 @@ class TreeSelector extends Localizer(LitElement) {
 	}
 
 	render() {
+		// Using no-auto-fit on d2l-dropdown-content to avoid having the component jump to the top on every
+		// node open and load. The tradeoff is that the user has to scroll the whole page now.
+		// We have a defect logged to improve this in future.
```
bemailloux, line 73: 👍

bemailloux (approved):

components/tree-filter.js
```diff
@@ -350,7 +387,9 @@ decorate(Tree, {
  * @property {String} openerText - appears on the dropdown opener if no items are selected
  * @property {String} openerTextSelected - appears on the dropdown opener if one or more items are selected
  * @fires d2l-insights-tree-filter-select - selection has changed; selected property of this element is the list of selected ids
- * @fires d2l-insights-tree-filter-request-children - owner should call addNodes with children of event.detail.id
+ * @fires d2l-insights-tree-filter-request-children - (dynamic tree only) owner should call tree.addNodes with children of event.detail.id
+ * @fires d2l-insights-tree-filter-search - (dynamic tree only) owner may call this.addSearchResults with nodes and ancestors matching
```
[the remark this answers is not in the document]
bemailloux: That's fine, no need for a unit test. Just wondering if you tested that manually.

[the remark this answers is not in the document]
johngwilkinson: Oh, yeah, I did.

PR #78, opened 2020-10-05
US120279 Users table name column

bemailloux:
Adds additional info to the name column in the users table.

Quality notes
* Testing
  * Responsiveness - the table as a whole is wonky (and should be fixed along with the wonky cards) but this column in particular is good: the text resizes
* Security, perf, docs: N/A
* UX
  * Checked with Kevin on text sizes (main text is d2l-body-standard, subtext is d2l-body-small)
  * Checked with Megan that if a username is blank, it can appear as blank in the table (consistent with Class Engagement page)
  * works in RTL
  * a11y: still needs work. I need to ask Kevin what a screen reader user should hear for the column and the table during loading

Screenshots - Loaded
image

Loading: medium size screen
image

Loading: full size screen
image

@anhill-D2L @johngwilkinson @MykolaGalian

### DIFF
--- a/components/table.js
+++ b/components/table.js
@@ -127,16 +127,6 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 			:host([dir="rtl"]) .d2l-insights-table-table .d2l-insights-table-row-last > .d2l-insights-table-cell-last {
 				border-bottom-left-radius: 8px;
 			}
-
-			:host([skeleton]) .d2l-insights-table-cell > div {
-				width: 45%;
-			}
-
-			@media (max-width: 1024px) {
-				:host([skeleton]) .d2l-insights-table-cell > div {
-					width: 100%;
-				}
-			}
 		`];
 	}
 
@@ -163,6 +153,8 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 	_renderThead() {
 		const styles = {
 			'd2l-insights-table-row-first': true,
+			// if skeleton view is displayed, then there will definitely be skeleton rows in the table,
+			// even if data.length is 0. Therefore if skeleton is true, don't apply the "last row" style to header
 			'd2l-insights-table-row-last': !this.skeleton && this.data.length === 0
 		};
 
@@ -217,18 +209,22 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 			</td>
 		`;
 
-		if (!this.skeleton) {
-			if (columnType === COLUMN_TYPES.TEXT_SUB_TEXT) {
-				return html`
-					<td class="${classMap(styles)}">
-						<div class="d2l-body-standard">${value[0]}</div>
-						<div class="d2l-body-small">${value[1]}</div>
-					</td>
-				`;
-			} // future work: else if COLUMN_TYPES.SUBCOLUMNS...
+		if (this.skeleton) {
+			return defaultHtml; // regardless of the column type, because the data hasn't been loaded yet
 		}
 
-		return defaultHtml;
+		if (columnType === COLUMN_TYPES.TEXT_SUB_TEXT) {
+			return html`
+				<td class="${classMap(styles)}">
+					<div class="d2l-body-standard">${value[0]}</div>
+					<div class="d2l-body-small">${value[1]}</div>
+				</td>
+			`;
+		} else if (columnType === COLUMN_TYPES.NORMAL_TEXT) {
+			return defaultHtml;
+		} // future work: else if COLUMN_TYPES.SUBCOLUMNS...
+
+		throw new Error('Users table: unknown column type');
 	}
 }
 customElements.define('d2l-insights-table', Table);
--- a/components/table.js
+++ b/components/table.js
@@ -1,13 +1,18 @@
+import { bodySmallStyles, bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html, LitElement } from 'lit-element';
-import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
 import { classMap } from 'lit-html/directives/class-map.js';
 import { Localizer } from '../locales/localizer';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
+export const COLUMN_TYPES = {
+	NORMAL_TEXT: 0,
+	TEXT_SUB_TEXT: 1
+};
+
 /**
  * @property {String} title - for use by screen reader users
- * @property {Array} columns - list of column header text
+ * @property {Array} columnInfo - list of column info. Contains headerText and columnType
  * @property {Array} data - a row-indexed 2D array of rows and columns.
  * E.g. data[0] gets the entire first row; data[0][0] gets the first row / first column
  */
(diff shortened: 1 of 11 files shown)


components/table.js
```diff
@@ -194,18 +203,32 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
 			'd2l-insights-table-cell-first': idx === 0,
-			'd2l-insights-table-cell-last': idx === row.length - 1
+			'd2l-insights-table-cell-last': idx === this._numColumns - 1
 		};
 
-		return html`
+		const defaultHtml = html`
 			<td class="${classMap(styles)}">
-				<div class="d2l-skeletize d2l-body-standard">${value}</div>
+				<div class="d2l-skeletize d2l-skeletize-95 d2l-body-standard">${value}</div>
```
Vitalii-Misechko, line 216: The style contains `d2l-skeletize-95` so it seems that styles `:host([skeleton]) .d2l-insights-table-cell > div` are obsolete.

Vitalii-Misechko: I consider `skeleton` as a exception/special case. therefore, I would suggest to structure code respectively. I would invert `if (!this.skeleton)` and exit from the function as soon as `skeleton` code is completed

```JS
if( this.skeleton ) {
 const defaultHtml = ...
 return defaultHtml
}

if (columnType === COLUMN_TYPES.TEXT_SUB_TEXT) {
...
return html`...
} if (columnType === COLUMN_TYPES....){
...
return html`...
}

throw new Error('Unknown column type')
```

components/users-table.js
```diff
@@ -61,66 +62,64 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 		return itemsCount ? Math.ceil(itemsCount / this._pageSize) : 0;
 	}
 
 	get _displayData() {
 		if (this.skeleton) {
-			return this._loadingData;
+			// a DEFAULT_PAGE_SIZE x columnInfoLength 2D array filled with a generic string
+			// the text does not matter here, as long as it's not an empty string.
+			// (an empty string leads to an empty div which in turn does not render skeleton rectangle)
+			return Array(DEFAULT_PAGE_SIZE).fill(Array(this.columnInfo.length).fill('Loading'));
```
Vitalii-Misechko, line 82: neat!

components/table.js
```diff
@@ -151,30 +156,34 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
 
+	get _numColumns() {
+		return this.columnInfo.length;
+	}
+
 	_renderThead() {
 		const styles = {
 			'd2l-insights-table-row-first': true,
-			'd2l-insights-table-row-last': this.data.length === 0
+			'd2l-insights-table-row-last': !this.skeleton && this.data.length === 0
```
Vitalii-Misechko, line 166: `skeleton == true` means that `this.data.length` is 0 but the code renders some default data. I guess comment here would be good

components/users-table.js
```diff
@@ -61,66 +62,64 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		return [...Array(numberOfColumns).keys()].map(() => 'Loading');
+		this._pageSize = DEFAULT_PAGE_SIZE;
 	}
 
 	get _itemsCount() {
-		if (this.skeleton) {
-			return this._loadingData.length;
-		}
-
 		return this.data.userDataForDisplay.length;
```
Vitalii-Misechko, line 69: Should we use `_displayData` instead of `this.data.userDataForDisplay` everywhere in this component to provide consistent data depending on `this.skeleton` ?

test/components/results-card.test.js
```diff
@@ -1,38 +1,38 @@
...
-		it('should render the proper value for users', async() => {
-			const el = await fixture(html`<d2l-insights-results-card .data="${data}"></d2l-insights-results-card>`);
-			await new Promise(resolve => setTimeout(resolve, 200)); // allow fetch to run
-			expect(el.shadowRoot.querySelector('d2l-labs-summary-card').value).to.deep.equal('4');
-			expect(el.shadowRoot.querySelector('d2l-labs-summary-card').title).to.deep.equal('Results');
-			expect(el.shadowRoot.querySelector('d2l-labs-summary-card').message).to.deep.equal('Users returned within results.');
-		});
-	});
-});
+import '../../components/users-table.js';
```
Vitalii-Misechko, line 1: Is this `LF` to `CRLF`? If true could you configure your IDE to use `.editorconfig` in the root of the project

test/components/results-card.test.js
```diff
@@ -1,38 +1,38 @@
...
+
+import { expect, fixture, html } from '@open-wc/testing';
+import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
+
+describe('d2l-insights-results-card', () => {
```
Vitalii-Misechko, line 6: Good catch! I was wondering why is there two set of tests for `d2l-insights-users-table`.

test/components/users-table.test.js
```diff
@@ -55,136 +70,107 @@ describe('d2l-insights-users-table', () => {
...
-				await pageSelector.updateComplete;
+				await finishUpdate();
 
 				expect(pageSelector.selectedCountOption).to.equal(50);
 				expect(pageSelector.maxPageNumber).to.equal(0);
 				expect(pageSelector.pageNumber).to.equal(0);
 
-				const displayedUsers = Array.from(innerTable.shadowRoot.querySelectorAll('tbody > tr > td:first-child > div'));
-				expect(displayedUsers.length).to.equal(0);
+				verifyUserInfoColumn(innerTable, 0, 0);
```
Vitalii-Misechko, line 148: neat!

Vitalii-Misechko (approved):

test/components/results-card.test.js
bemailloux: Looks like it was `CRLF` before and my IDE converted it to `LF`

components/users-table.js
bemailloux: I'll ask Kevin about this - basically you're asking if we should show "total users: 20" or "total users: 0" when we're displaying a skeleton view

components/users-table.js
```diff
@@ -61,66 +62,64 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		}
-
 		const itemsCount = this._itemsCount;
 		return itemsCount ? Math.ceil(itemsCount / this._pageSize) : 0;
 	}
 
 	get _displayData() {
 		if (this.skeleton) {
-			return this._loadingData;
+			const loadingPlaceholderText = this.localize('components.insights-users-table.loadingPlaceholder');
```
bemailloux, line 79: @Vitalii-Misechko FYI a screen reader can read this so I'm localizing it

components/users-table.js
bemailloux: It looks like from a UX perspective we want to should "total users: 0" during load. So more generally I think we should continue to keep skeleton view and the data state separate (that is, don't use `_displayData` everywhere)

rohitvinnakota (approved):

components/users-table.js
Vitalii-Misechko: Sounds reasonable! Thanks for digging into it. Could you add some comment inside `_displayData`? Because using `_displayData`  everywhere is too tempting )

components/users-table.js
Vitalii-Misechko: Thanks! The other option would be add `aria-hidden="true"`

Vitalii-Misechko (approved):

bemailloux:
Regarding a11y changes:
- You can't add an aria-label to a `<tr>` element and expect it to work with all screen readers
- Therefore, "Course(s)" was changed to just "Courses" because NVDA was reading "Course(s)" as "course s" in NVDA and "course left paren s right paren" in JAWS

PR #79, opened 2020-10-06
Access interaction fix

MykolaGalian:
Issue:
 For the "Course Access" card if I add in a `select` event for a point the `_valueClickHandler()` method, which sets `isApplied = true` for the card filter, we have a problem with that event. The `select` event for the last bar ("< 1 day ago") becomes callable  only in some cases:
1)  If I select the last bar first, before selecting other bars;
2) If the mouse goes outside the plot and then again clicking on this bar. 

If I add one more last bar, then I have the same problems with it. This problem still exists even if I disable all other events and logic (for example, color painting points process). Looks like some kind of conflict with mobx. Because If I remove the observable decorator from `isApplied` in `CardFilter` class,  the `select`event for the last bar start to works.

The proposed solution is to add another last bar and hidе it. But at the same time, the Y-axis is slightly lengthened.

Kevin approve that for now (temporary)

@anhill-D2L @bemailloux @rohitvinnakota @hyehayes

### DIFF
--- a/components/course-last-access-card.js
+++ b/components/course-last-access-card.js
@@ -198,8 +198,8 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 					if (this.point.y === 1) {
 						return `${that._cardTooltipTextSingleUser[this.point.x]}`;
 					}
-					if (this.point.x !== 6) { 
-					return `${that._cardTooltipText(this.point.y)[this.point.x]}`;
+					if (this.point.x !== 6) {
+						return `${that._cardTooltipText(this.point.y)[this.point.x]}`;
 					} else return false;
 				},
 				backgroundColor: 'var(--d2l-color-ferrite)',
--- a/components/course-last-access-card.js
+++ b/components/course-last-access-card.js
@@ -186,6 +186,9 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 							that._colorAllPoints(this.series[0].data, 'var(--d2l-color-celestine)');
 						}
 						this.render(false);
+					},
+					load: function() {
+						this.series[0].data[6].graphic.hide();
 					}
 				}
 			},
@@ -195,7 +198,9 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 					if (this.point.y === 1) {
 						return `${that._cardTooltipTextSingleUser[this.point.x]}`;
 					}
+					if (this.point.x !== 6) { 
 					return `${that._cardTooltipText(this.point.y)[this.point.x]}`;
+					} else return false;
 				},
 				backgroundColor: 'var(--d2l-color-ferrite)',
 				borderColor: 'var(--d2l-color-ferrite)',
@@ -230,7 +235,8 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 					}
 				},
 				width: '108%',
-				categories: this._cardCategoriesText
+				categories: this._cardCategoriesText,
+				showLastLabel: false
 			},
 			yAxis: {
 				tickAmount: 5,
--- a/model/data.js
+++ b/model/data.js
@@ -220,7 +220,7 @@ export class Data {
 
 	get courseLastAccessDates() {
 		// return an array of size 6, each element mapping to a category on the course last access bar chart
-		const dateBucketCounts = [0, 0, 0, 0, 0, 0];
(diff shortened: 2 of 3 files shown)


johngwilkinson:
Let's see if we can get to the root of this. To find mobx issues, I know of two main tools:
1. console.log statements in render functions, event handlers, etc. to see if they are being called unexpectedly
2. `import { trace } from "mobx"` and `trace()` inside of render (or code called by render(), directly or otherwise) to find out why mobx is triggering it

MykolaGalian:
> Where is the select handler? I don't see it. I see a `_valueClickHandler()` method, but I don't see where it is referenced.
We call this method here: https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L122

johngwilkinson:
> > Where is the select handler? I don't see it. I see a `_valueClickHandler()` method, but I don't see where it is referenced.
> > We call this method here: https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L122

That link takes me to the method - but where does that method get called or wired up?

MykolaGalian:
> > > Where is the select handler? I don't see it. I see a `_valueClickHandler()` method, but I don't see where it is referenced.
> > > We call this method here: https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L122
> 
> That link takes me to the method - but where does that method get called or wired up?

https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L282

johngwilkinson:
> > > > Where is the select handler? I don't see it. I see a `_valueClickHandler()` method, but I don't see where it is referenced.
> > > > We call this method here: https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L122
> > 
> > 
> > That link takes me to the method - but where does that method get called or wired up?
> 
> https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L282

Okay, thanks. If the issue is indeed mobx-related, it may be due to the two separate changes this code makes: it sets the filter as enabled, and it adds the category. It would be good to make a single method on `Data`, decorated with `action`, that enables the filter and adds the category. It might solve your problem, but it would be good to do anyway.

From the mobx docs https://doc.ebichu.cc/mobx/refguide/action.html:
> It takes a function and returns a function with the same signuature. But wrapped with transaction, untracked and allowStateChanges. Especially the fact that transaction is applied automatically yields great performance benefits; actions will batch mutations and only notify computed values and reactions after the (outer most) action has finished. This makes sure intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has finished.
> It is advised to use (@)action on any function that modifies observables or has side effects.

MykolaGalian:
> > > > > Where is the select handler? I don't see it. I see a `_valueClickHandler()` method, but I don't see where it is referenced.
> > > > > We call this method here: https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L122
> > > 
> > > 
> > > That link takes me to the method - but where does that method get called or wired up?
> > 
> > 
> > https://github.com/Brightspace/insights-engagement-dashboard/blob/6d8da782ec7814d06716e380bc5af10245677bcd/components/course-last-access-card.js#L282
> 
> Okay, thanks. If the issue is indeed mobx-related, it may be due to the two separate changes this code makes: it sets the filter as enabled, and it adds the category. It would be good to make a single method on `Data`, decorated with `action`, that enables the filter and adds the category. It might solve your problem, but it would be good to do anyway.
> 
> From the mobx docs https://doc.ebichu.cc/mobx/refguide/action.html:
> 
> > It takes a function and returns a function with the same signuature. But wrapped with transaction, untracked and allowStateChanges. Especially the fact that transaction is applied automatically yields great performance benefits; actions will batch mutations and only notify computed values and reactions after the (outer most) action has finished. This makes sure intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has finished.
> > It is advised to use (@)action on any function that modifies observables or has side effects.

if I remove from event two other methods - selection of the category and the coloring of the points, then the problem does not disappear

PR #80, opened 2020-10-06
Add course count to user table

rohitvinnakota:
[US120513](https://rally1.rallydev.com/#/detail/userstory/435482082300?fdp=true): [Engagement][UserTable] Course(s)

### DIFF
--- a/model/data.js
+++ b/model/data.js
@@ -172,14 +172,14 @@ export class Data {
 	get userDataForDisplay() {
 		// map to a 2D userData array, with column 0 as the lastFirstName
 		// then sort by lastFirstName
-
+		const recordsByUser = this.recordsByUser;
 		return this.users
 			.map(user => [
 				// When add/remove column here DO NOT forget to update `user-table._lodaingData` to show skeleton properly
 
 				`${user[USER.LAST_NAME]}, ${user[USER.FIRST_NAME]}`, // last first name
 				// 'N/A', // last accessed system
-				'', // courses
+				recordsByUser.get(user[USER.ID]).length, // courses
 				'', // average grade
 				'', // average time in content
 				// 'N/A', // average discussion activity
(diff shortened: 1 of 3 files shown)


model/data.js
```diff
@@ -173,11 +173,11 @@ export class Data {
 	get userDataForDisplay() {
 		// map to a 2D userData array, with column 0 as a sub-array of [lastFirstName, username - id]
 		// then sort by lastFirstName
-
+		const recordsByUser = this.recordsByUser;
```
bemailloux, line 176: This could be optimized differently. `userDataForDisplay` is recalculated every time something changes on the page, and you're rebuilding the `recordsByUser` map every time that happens. This is fast if a small number of records are displayed, but slow if a large number of records are displayed.

A different way would be to build the map once for all records after new data is loaded from the server. That would add to the initial load time but make recalculating `userDataForDisplay` very fast.

I guess the right way to optimize depends on the other metrics we want to calculate (avg grade, avg time in content...) If calculating those on the fly takes too long we should probably move towards bulk calculating them on data reload.

test/model/data.test.js
```diff
@@ -383,8 +383,8 @@ describe('Data', () => {
 		it('should only display users in view', async() => {
 			const roleFilters = [mockRoleIds.instructor];
 			const expectedUsers = [
-				[['Harrison, George', 'gharrison - 300'], '', '', ''],
-				[['McCartney, Paul', 'pmccartney - 200'], '', '', '']
+				[['Harrison, George', 'gharrison - 300'], 1, '', ''],
+				[['McCartney, Paul', 'pmccartney - 200'], 3, '', '']
```
bemailloux, line 387: users-table should have a few tests as well

model/data.js
johngwilkinson: But user data for display must be recalculated every time a any filter changes, not just on server reload. I believe this is the right way.

bemailloux: Oh right: the list of records for each user will change depending on which filters are applied. I was thinking that list wouldn't change except on server reload. Never mind then

johngwilkinson (approved):

bemailloux (approved):

PR #82, opened 2020-10-07
Access interaction fix

MykolaGalian:
Issue: When some bar is selected in "Course Access" card for the first time, the filter button is drawn using "d2l-applied-filters". After that, I cannot select the last bar on this card. It can be selected only if the mouse goes outside the plot and then again clicking on this bar.   
The proposed solution is fix CSS for parent div container .d2l-insights-summary-container-applied-filters (https://stackoverflow.com/a/19221684/14336208).
However, due to the reservation of the minimum space for the container, the indent has increased.

UX: I showed it on demo after stand-up.

Previous PR: https://github.com/Brightspace/insights-engagement-dashboard/pull/72

Next story for that defect: https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fdefect%2F440857153012&sharedViewId=328666453148

@anhill-D2L @bemailloux @rohitvinnakota @hyehayes

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -58,6 +58,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 				}
 
 				.d2l-insights-summary-container-applied-filters {
+					height: 30px;
 					width: 100%;
 				}
 
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -58,8 +58,8 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 				}
 
 				.d2l-insights-summary-container-applied-filters {
+					height: auto;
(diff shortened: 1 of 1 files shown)


Vitalii-Misechko:
You should demo it to Kevin. 
Could you also add screenshots to the story?

johngwilkinson:
I don't think having a bunch of empty space like that is going to be acceptable. Also, what if the buttons wrap to a third line? Seems like you're on the right track, but there is probably a better solution. Maybe setting the size of the char's parent container would work? Possibly we need to call the `reflow()` function mentioned in your link, but I hope not.

johngwilkinson (approved):

Vitalii-Misechko (approved):

PR #83, opened 2020-10-07
US119525: use new roles api (only need insights portal perm)

johngwilkinson:
Quality: roles filter shows roles as expected

FYI @bemailloux @MykolaGalian @NicholasHallman  @Vitalii-Misechko


model/lms.js
```diff
@@ -1,4 +1,4 @@
-const rolesEndpoint = '/d2l/api/lp/1.23/roles/';
+const rolesEndpoint = '/d2l/api/ap/unstable/insights/data/roles';
```
anhill-D2L, line 1: It's an unstable route so we have the right to change the api, right?

anhill-D2L (approved):

model/lms.js
johngwilkinson: Yup

rohitvinnakota (approved):

PR #84, opened 2020-10-07
US120515, US120514: add avg tic and grade to users table

johngwilkinson:
Quality Notes
* functional
  * user w. time and grade in two courses (NB: to get time in content, need to log in as the user - impersonation doesn't seem to work)
  * users with no time in content or grades
  * filtered by role so one or the other of the user's courses were in view; numbers update accordingly
  * i18n: tried german and confirmed dots and commas get switched
* security: n/a
* performance: no significant change
* devops: n/a
* ux/docs
  * uses standard d2l number/percent formatting
  * screen-reader + keyboard works (need to use arrow keys from the last graph to reach the table)

image

### DIFF
--- a/model/data.js
+++ b/model/data.js
@@ -1,10 +1,13 @@
 import { action, autorun, computed, decorate, observable } from 'mobx';
 import { COURSE_OFFERING, RECORD, USER } from '../consts';
 import { fetchCachedChildren, fetchLastSearch } from './lms.js';
+import { formatNumber, formatPercent } from '@brightspace-ui/intl/lib/number.js';
 import { OrgUnitSelectorFilter, RoleSelectorFilter, SemesterSelectorFilter } from './selectorFilters.js';
 import { CardFilter } from './cardFilter.js';
 import { TABLE_USER } from '../components/users-table';
 import { Tree } from '../components/tree-filter';
+
+const numberFormatOptions = { maximumFractionDigits: 2 };
 
 function unique(array) {
 	return [...new Set(array)];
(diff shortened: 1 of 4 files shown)


johngwilkinson:
Kevin approved the screen-shot, and specified 2 decimal places at most.

rohitvinnakota (approved):

model/data.js
```diff
@@ -160,12 +168,17 @@ export class Data {
...
-				'', // average grade
-				'', // average time in content
-			])
+			.map(user => {
+				const records = recordsByUser.get(user[USER.ID]);
+				const recordsWithGrades = records.filter(r => r[RECORD.CURRENT_FINAL_GRADE] !== null);
+				const avgFinalGrade = avgOf(recordsWithGrades, RECORD.CURRENT_FINAL_GRADE);
+				return [
+					[`${user[USER.LAST_NAME]}, ${user[USER.FIRST_NAME]}`, `${user[USER.USERNAME]} - ${user[USER.ID]}`],
+					records.length, // courses
```
bemailloux, line 177: I missed this on the last PR: should records be filtered by OU type == `COURSE_OFFERING`? Right now we're listing the number of total enrollments (not necessarily "courses") I think.

model/data.js
```diff
@@ -18,6 +21,11 @@ const OverdueAssignmentsFilterId = 'd2l-insights-overdue-assignments-card';
 const CourseLastAccessFilterId = 'd2l-insights-course-last-access-card';
 const CurrentFinalGradeFilterId = 'd2l-insights-current-final-grade-card';
 
+function avgOf(records, field) {
+	const total = records.reduce((sum, r) => sum + r[field], 0);
+	return total / records.length;
```
bemailloux, line 26: This could lead to division by zero (=Infinity). I'm not sure if the `formatPercent` function from intl handles that correctly.

model/data.js
johngwilkinson: The records from the server are for course offerings only.

model/data.js
johngwilkinson: It comes to `NaN` (til: 0/0 is NaN, but 1/0 is Infinity), which is falsy. For average grade, where there might be no records, this is handled below. For average tic, there is always at least one record (because if the user has no records they aren't in the table).

bemailloux: Yay javascript quirkyness!

bemailloux (approved):

PR #86, opened 2020-10-08
Us118210/default view popup

bemailloux:
Opens a popup informing users about the details of the default view.

Quality notes
* Testing
  * It shows the dialog when data is received from the server and there are default view org units
  * It does not show the dialog when data is loading, or when there are no default view org units
  * It shows the name and id of all org units in the expandable course list inside the dialog
  * Clicking on the ok button closes the dialog
  * Tested with a very long list (100) of default courses and course names that go off the screen; popup produces scrollbars when necessary
  * Popup resizes when list is expanded and collapsed
  * Responsiveness
* Security, perf, docs: N/A
* UX
  * Accessibility
    * All controls and text in the dialog are keyboard-navigable
    * Expansion control informs screen reader users if the expanding element is expanded or collapsed 

Screenshots - Collapsed courses list
image

Expanded courses list
image

@MykolaGalian @rohitvinnakota @anhill-D2L

### DIFF
--- a/components/default-view-popup.js
+++ b/components/default-view-popup.js
@@ -32,10 +32,7 @@ class DefaultViewPopup extends Localizer(LitElement) {
 	}
 
 	get _displayData() {
-		if (!this.data.serverData
-			|| !this.data.serverData.defaultViewOrgUnitIds
-			|| !this.data.serverData.defaultViewOrgUnitIds.length
-		) {
+		if (!this.data.isDefaultView) {
 			return [];
 		}
 
--- a/components/default-view-popup.js
+++ b/components/default-view-popup.js
@@ -15,7 +15,11 @@ class DefaultViewPopup extends Localizer(LitElement) {
 
 	static get styles() {
 		return css`
-			.default-view-description {
+			.d2l-insights-default-view-popup-description {
+				margin: 0;
+			}
+
+			.d2l-insights-default-view-popup-course-list {
 				margin: 0;
 			}
 		`;
@@ -28,11 +32,10 @@ class DefaultViewPopup extends Localizer(LitElement) {
 	}
 
 	get _displayData() {
-		if (this.data.isLoading) {
-			return [];
-		}
-
-		if (!this.data.serverData.defaultViewOrgUnitIds || !this.data.serverData.defaultViewOrgUnitIds.length) {
+		if (!this.data.serverData
+			|| !this.data.serverData.defaultViewOrgUnitIds
+			|| !this.data.serverData.defaultViewOrgUnitIds.length
+		) {
 			return [];
 		}
 
@@ -51,10 +54,10 @@ class DefaultViewPopup extends Localizer(LitElement) {
 					width="615"
 					@d2l-dialog-close="${this._closeDialog}">
 
-				<p class="default-view-description">
+				<p class="d2l-insights-default-view-popup-description">
 					${this.localize('components.insights-default-view-popup.defaultViewDescription1', { numDefaultCourses: displayData.length })}
 				</p>
-				<p class="default-view-description">
+				<p class="d2l-insights-default-view-popup-description">
 					${this.localize('components.insights-default-view-popup.defaultViewDescription2')}
 				</p>
 
@@ -64,7 +67,7 @@ class DefaultViewPopup extends Localizer(LitElement) {
 					@d2l-insights-expander-with-control-expanded="${this._resize}"
 					@d2l-insights-expander-with-control-collapsed="${this._resize}">
 
-					<ul>
+					<ul class="d2l-insights-default-view-popup-course-list">
 						${displayData.map(data => html`
 							<li>${data}</li>
 						`)}
--- a/components/default-view-popup.js
+++ b/components/default-view-popup.js
@@ -1,0 +1,90 @@
+import '@brightspace-ui/core/components/dialog/dialog';
+import '@brightspace-ui/core/components/button/button';
+import './expander-with-control';
+import { css, html, LitElement } from 'lit-element';
+import { Localizer } from '../locales/localizer';
+
+class DefaultViewPopup extends Localizer(LitElement) {
+
+	static get properties() {
+		return {
+			data: { type: Object, attribute: false },
+			opened: { type: Boolean, attribute: true, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return css`
+			.default-view-description {
+				margin: 0;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.data = {};
+		this.opened = false;
+	}
+
+	get _displayData() {
+		if (this.data.isLoading) {
+			return [];
+		}
+
+		if (!this.data.serverData.defaultViewOrgUnitIds || !this.data.serverData.defaultViewOrgUnitIds.length) {
+			return [];
+		}
+
+		return this.data.serverData.defaultViewOrgUnitIds.map(id => {
+			const name = this.data.orgUnitTree.getName(id);
+			return this.localize('components.tree-filter.node-name', { orgUnitName: name, id });
+		});
+	}
+
+	render() {
+		const displayData = this._displayData;
+		return html`
+			<d2l-dialog
+				?opened="${this.opened}"
+				title-text="${this.localize('components.insights-default-view-popup.title')}"
+					width="615"
+					@d2l-dialog-close="${this._closeDialog}">
+
+				<p class="default-view-description">
+					${this.localize('components.insights-default-view-popup.defaultViewDescription1', { numDefaultCourses: displayData.length })}
+				</p>
+				<p class="default-view-description">
+					${this.localize('components.insights-default-view-popup.defaultViewDescription2')}
+				</p>
+
+				<d2l-insights-expander-with-control
+					control-expanded-text="${this.localize('components.insights-default-view-popup.collapseDefaultCourseList')}"
+					control-collapsed-text="${this.localize('components.insights-default-view-popup.expandDefaultCourseList')}"
+					@d2l-insights-expander-with-control-expanded="${this._resize}"
+					@d2l-insights-expander-with-control-collapsed="${this._resize}">
+
+					<ul>
+						${displayData.map(data => html`
+							<li>${data}</li>
+						`)}
+					</ul>
+
+				</d2l-insights-expander-with-control>
+
+				<d2l-button primary slot="footer" @click="${this._closeDialog}">
+					${this.localize('components.insights-default-view-popup.buttonOk')}
+				</d2l-button>
+			</d2l-dialog>
+		`;
+	}
+
+	_resize() {
+		this.shadowRoot.querySelector('d2l-dialog').resize();
+	}
+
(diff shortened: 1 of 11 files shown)


components/expander-with-control.js
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	_onCollapsed(event) {
+		event.detail.collapseComplete.then(() => {
+			this.dispatchEvent(new Event('d2l-insights-expander-with-control-collapsed'));
+		});
+	}
+
+	_handleKeydown(event) {
+		if (event.key === 'Enter') {
```
Vitalii-Misechko, line 81: Should the code handle the same way `Spacebar`?

test/components/applied-filters.test.js
```diff
@@ -19,7 +19,7 @@ describe('d2l-insights-applied-filters', () => {
 
 	describe('accessibility', () => {
 		it('should pass all axe tests', async function() {
-			this.timeout(3000);
+			this.timeout(3500);
```
Vitalii-Misechko, line 22: I believe, that at some point we should try to get to the root cause of these timeouts

test/components/expander-with-control.test.js
```diff
@@ -0,0 +1,113 @@
...
+	describe('constructor', () => {
+		it('should construct', () => {
+			runConstructor('d2l-insights-expander-with-control');
+		});
+	});
+
+	describe('accessibility', () => {
+		it('should pass all axe tests', async() => {
+			await expect(elCollapsed).to.be.accessible();
+			await expect(elExpanded).to.be.accessible();
```
Vitalii-Misechko, line 40: Nice!

test/components/expander-with-control.test.js
```diff
@@ -0,0 +1,113 @@
...
+		it('should fire collapsed event if element is expanded and user hits Enter', async function() {
+			this.timeout(3000);
+
+			const listener = oneEvent(elExpanded, 'd2l-insights-expander-with-control-collapsed');
+
+			const controlDiv = elExpanded.shadowRoot.querySelector('div');
+			controlDiv.dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter' }));
+
+			await listener; // will time out if event is not fired
+		});
```
Vitalii-Misechko, line 100: `TIL`:  Awesome! This is the first test in this repo that emulates keyboard interaction.

Vitalii-Misechko (approved):
LGTM!

components/expander-with-control.js
bemailloux: Good question, I'll ask Kevin

test/components/expander-with-control.test.js
bemailloux: It would be nice if it worked :/ For some reason Chrome on OSX fails every time unless I set the timeout to 10 seconds

test/components/applied-filters.test.js
bemailloux: It seems like many of the failures are happening on OSX, it would be nice to have an actual Mac to debug with

One thing I suspect might be happening is the Sauce test runners did not maximize the browser windows before running the tests. If browser windows are minimized then tests take a lot longer to run than normal. I played around with Sauce config options to try and maximize the windows but they didn't seem to help.

components/default-view-popup.js
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	constructor() {
+		super();
+		this.data = {};
+		this.opened = false;
+	}
+
+	get _displayData() {
+		if (!this.data.serverData
```
johngwilkinson, line 35: If you look at the server-side code, you'll see there's another case to consider. Addressing it will likely require passing an extra flag back from the server, although I haven't thought it through in detail. So probably a matter for another PR or two.

Case 1 (covered here): data is not truncated; server returns all data for the top N semesters along with a list of default view courses.
Case 2: enrollments are truncated: the server sets selectedOrgUnitIds and returns data only for those selected ous (in almost all cases, this means there's no longer any truncation). It does this so the client knows it only has that data (whereas defaultViewOrgUnitIds doesn't indicate any missing data), and will apply reload logic if other ous are added to the selection.

For case 2, the server response doesn't distinguish between default view and some other call with an ou filter. One solution would be to use client-side state: if this is the initial load, then the default view to report is defaultOus || selectedOus. But it's probably better to have the server return the defaultView flag value that was passed in so that the client doesn't need to be stateful that way. Then this code would just be something like `this.data.serverData.defaultView ? (defOus || selectedOus) : []`.

engagement-dashboard.js
```diff
@@ -123,13 +124,18 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
...
+				<d2l-insights-default-view-popup .data="${this._data}" ?opened="${!this._isLoading && this._isDefaultView}"></d2l-insights-default-view-popup>
 		`;
 	}
 
 	get _isLoading() {
 		return this._data.isLoading;
 	}
 
+	get _isDefaultView() {
+		return this._data.serverData.defaultViewOrgUnitIds && this._data.serverData.defaultViewOrgUnitIds.length;
```
johngwilkinson, line 136: roughly this check is happening in two places (here and in the popup), so maybe move to a getter on `Data`?

test/components/applied-filters.test.js
johngwilkinson: Lately I have to run every build several times. :( I get failures of various kinds across browsers. It just seems like Sauce is super flaky.

test/components/expander-with-control.test.js
johngwilkinson: Given all the wrinkles around keyboard interaction, I wonder if you could use a standard interactable component instead of div - I've seen it suggested as best practice. A button, anchor, etc. with styling to make it look like an ordinary div.

test/engagement-dashboard.test.js
```diff
@@ -6,11 +6,19 @@ describe('d2l-insights-engagement-dashboard', () => {
...
 	describe('accessibility', () => {
 		it('should pass all axe tests', async function() {
-			this.timeout(4000);
+			this.timeout(5000);
 
 			const el = await fixture(html`<d2l-insights-engagement-dashboard demo></d2l-insights-engagement-dashboard>`);
 			// need for this delay might be tied to the mock data async loading in engagement-dashboard.js
 			await new Promise(resolve => setTimeout(resolve, 1500));
+
+			// close the default view dialog that shows up. It causes browsers on OSX to assign aria-attributes and
```
johngwilkinson, line 15: So, isn't that an actual bug in the code under test?

components/expander-with-control.js
bemailloux: It turns out any d2l-buttons will be triggered with spacebar and enter by default. Somehow they appear to be transforming the keyboard event into a mouse event, which is caught by `@click`. So `_handleKeydown` is not even necessary.

test/engagement-dashboard.test.js
bemailloux: The issue is that on OSX, buttons in the background are given `role="presentation"` to stop them from being tabbable. Then the test fails because those buttons now have `role="presentation"` and that role isn't allowed for buttons. Also, those buttons have aria-attributes and there shouldn't be aria-attributes on something with `role="presentation"`.

This doesn't happen on Windows.

It seems to me like it is a bug, but not ours to fix. It's probably something for whoever is maintaining the dialog. I think it is safe to ignore for this dashboard because it does work as expected for screen reader users: if the dialog exists, you can't navigate to elements outside the dialog.

components/default-view-popup.js
bemailloux: It sounds like in case 2, the default view popup should be appearing but it wouldn't with the current code, so I agree with addressing this in another PR. (If it were the other way around - default view appears too often - I'd try to address it here)

bemailloux:
> Given all the wrinkles around keyboard interaction, I wonder if you could use a standard interactable component instead of div - I've seen it suggested as best practice. A button, anchor, etc. with styling to make it look like an ordinary div.

"Styling to make it look like an ordinary div" is harder than you'd think at first glance. The reason is because browsers will by default apply an outline to elements if they're tab-focused, and that outline looks awful sometimes (see screenshot).
image

The sneaky way I'm getting around this is:
- the spec wants the whole div to be clickable. Ok, fine
- For sighted users, I make the whole div clickable, but not tabbable so it doesn't get that ugly outline
- For screen reader users, they can tab to the button icon instead and use that. The button icon has an aria-label and aria-expanded on it so the users know what it applies to

components/expander-with-control.js
bemailloux: TIL: buttons in modern browsers also fire a MouseEvent when a button is "submitted" using Enter or Spacebar keys. Verified on Chrome, FF, Edgium, Legacy

engagement-dashboard.js
bemailloux: Done

johngwilkinson (approved):

components/expander-with-control.js
Vitalii-Misechko: TIL: 😲 👍

bemailloux:
@Vitalii-Misechko and @johngwilkinson : I am still stuck with the one test that is failing on Chrome for OSX and which occasionally seems to crash Safari. 
```
Chrome 86.0.4240 (Mac OS X 10.13.6) d2l-insights-expander-with-control interactions/eventing should fire collapsed event if element is expanded and control is clicked FAILED
	Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

I've checked with Henoch that it actually works. Is it fine to skip the test? I'd be willing to troubleshoot further if I actually had access to a Mac, but I don't know how to get access to a VM.

johngwilkinson:
> @Vitalii-Misechko and @johngwilkinson : I am still stuck with the one test that is failing on Chrome for OSX and which occasionally seems to crash Safari.
> 
> ```
> Chrome 86.0.4240 (Mac OS X 10.13.6) d2l-insights-expander-with-control interactions/eventing should fire collapsed event if element is expanded and control is clicked FAILED
> 	Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
> ...
> Safari 13.0.1 (Mac OS X 10.13.6) ERROR
>   Disconnected, because no message in 30000 ms.
> ```
> 
> I've checked with Henoch that it actually works on Safari and Chrome on OSX. Is it fine to skip the test? I'd be willing to troubleshoot further if I actually had access to a Mac, but I don't know how to get access to a VM.

I blocked my story on the need for a mac (vm or otherwise); maybe you should do the same. I think we're reaching a point where we don't have what we need to do our work, and that needs to be known.

johngwilkinson (approved):

bemailloux:
> I blocked my story on the need for a mac (vm or otherwise); maybe you should do the same. I think we're reaching a point where we don't have what we need to do our work, and that needs to be known.

Sounds good. I can get started on fixing the logic for the truncation case in the meantime.

Vitalii-Misechko (approved):

PR #88, opened 2020-10-09
Remove space under filter and fix bar after selection DE40834

NicholasHallman:
Related to #79 

I managed to trace the problem into HighCharts. 

When a bar in the chart is selected and the graphs are shifted down in the document the graphPosition is not updated in the pointer when the chart is not close to the top of the page. If the chart is visible from the top of the page or if you zoom out enough to see the top of the page then the chartPosition is updated. 

By reseting that state on every click the problem is resolved. 

Wondering your thoughts on this fix or if you think we should submit a PR directly to HighCharts.

@anhill-D2L  @bemailloux @rohitvinnakota @hyehayes

### DIFF
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -122,6 +122,8 @@ class Chart extends SkeletonMixin(LitElement) {
 				}
 			});
 			this.chart = H[constructorType](this.chartContainer, this.options, this.chartCreated.bind(this));
+			// force highcharts to recalculate the chart position incase the filter move the graph
+			this.chartContainer.addEventListener('click', (e) => (delete this.chart.pointer.chartPosition));
 		}
 	}
 
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -123,7 +123,7 @@ class Chart extends SkeletonMixin(LitElement) {
 			});
 			this.chart = H[constructorType](this.chartContainer, this.options, this.chartCreated.bind(this));
 			// force highcharts to recalculate the chart position incase the filter move the graph
-			this.chartContainer.addEventListener('click', (e) => (delete this.chart.pointer.chartPosition));
+			this.chartContainer.addEventListener('click', () => (delete this.chart.pointer.chartPosition));
 		}
 	}
 
--- a/components/course-last-access-card.js
+++ b/components/course-last-access-card.js
@@ -157,7 +157,7 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	_pointUpdateColor(point, colorForPoint) {
 		point.update({ color: colorForPoint }, false);
 	}
- 
+
 	render() {
 		// NB: relying on mobx rather than lit-element properties to handle update detection: it will trigger a redraw for
 		// any change to a relevant observed property of the Data object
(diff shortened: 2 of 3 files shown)


components/chart/chart.js
```diff
@@ -122,6 +122,8 @@ class Chart extends SkeletonMixin(LitElement) {
 				}
 			});
 			this.chart = H[constructorType](this.chartContainer, this.options, this.chartCreated.bind(this));
+			// force highcharts to recalculate the chart position incase the filter move the graph
+			this.chartContainer.addEventListener('click', () => (delete this.chart.pointer.chartPosition));
```
Vitalii-Misechko, line 126: Awesome! 
@NicholasHallman, just curious how you found this solution?

Vitalii-Misechko:
> Wondering your thoughts on this fix or if you think we should submit a PR directly to HighCharts.

If we don't have any rendering issues (slow rendering or other weirdness) then it looks good to me. 

It seems highcarts has [a bunch open bugs](https://www.google.com/search?q=%22chart.pointer.chartPosition%22&rlz=1C1GCEU_enUA914UA914&oq=%22chart.pointer.chartPosition%22&aqs=chrome..69i57.1111j0j7&sourceid=chrome&ie=UTF-8) related to `this.chart.pointer.chartPosition`.

Vitalii-Misechko (approved):

components/chart/chart.js
NicholasHallman: I added breakpoints to the tooltip for hovering, I noticed that the hover tooltip was appearing and then fading out immediately. Then I found the hit detection code they were using relied on an internal calculation of the bounds based on the chartPosition. Once I added the above code and tested it fixed the hover and the select events.

Vitalii-Misechko: Cool!

NicholasHallman:
I found the issue on Highcharts https://github.com/highcharts/highcharts/issues/13220 looks like they haven't gotten around to fixing it yet...

johngwilkinson (commented):
Good find! Although this is a tiny PR, would you mind adding Quality Notes (https://desire2learn.atlassian.net/wiki/spaces/DRACO/pages/486309905/Development+Outcomes)? We try to do a quality review before each merge, and the easiest way to do that is to include the notes in the PR, and then approval of the PR includes approval of the quality notes.

It's mostly just for practice in this review, but it should help answer a couple of small questions I had: which browsers did you test on, and did you interact with the other charts (particularly the scatter plot, since it is quite different) to make sure this doesn't change their behaviour?

NicholasHallman:
Sure thing @johngwilkinson 

Functional Testing

- Verified it works by manual testing in Edgium and Firefox, 
- Other charts, such as the scatter plot, have not shown any change in behavior when
  - Clicking, mouse over, mouse moving, or mouse out
- Change does not impact the behavior of other bar charts (high enough on the page reflow is properly triggered)

Performance

- The biggest possible performance impact is the event listener on each graph container, but it's limited to on click, and only performs one line of code.

Security

- N/A modifies the internal state of a 3rd party frontend library

Documentation

- Line is commented with reason for existence, once issue is closed in Highcharts we can remove it.

UX

- Reverted space back to design approved spec. 

Demo: UX/UI changes should be demoed to either the Product Manager or the Product Designer

- Done in standup

PR #89, opened 2020-10-13
Us118251/org unit children load more

johngwilkinson:
This is only relevant if an org has some org unit with more than 5000 children - in which case it's now possible to hit "load more" to eventually see all of them. Hopefully gets no use in practice. :)

Quality Notes
* functional
  * without ou truncation, with and without limiting page size on the children call: no change
  * with ou truncation (set CourseLimit config to 50)
    * no page size set: no change
    * temporarily have client request page size of 10
      * load more appears for root
      * load more appears for a course template with lots of children
      * load more button does not appear for a department (only one child)
      * repeatedly clicking load more pulls in all children, then button disappears
      * UX note: for simplicity (not rewriting and optimizing the sql), the children arrive in orgUnitId order, even though they are displayed alphabetically; this is not ideal, but this is very much an edge case so user impact should be minimal
      * select children of fully or partially loaded node
      * open a template so "load more" appears both at root and inside the template node, and alternate loading in each
* performance: no significant change
* security: n/a
* devops: n/a (server already has the necessary support)
* ux/docs: approved by Kevin via Slack

image

image

FYI @bemailloux @MykolaGalian @NicholasHallman @rohitvinnakota

### DIFF
--- a/test/components/tree-filter.test.js
+++ b/test/components/tree-filter.test.js
@@ -188,30 +188,55 @@ describe('Tree', () => {
 			expect(staticTree.isPopulated(4)).to.be.true;
 		});
 
-		it('should add the given nodes', () => {
-			dynamicTree.addNodes(1001, [[991, 'new1', mockOuTypes.course], [992, 'new2', mockOuTypes.courseOffering]]);
+		[true, false].forEach(hasBookmark => {
+			it(`should add the given nodes ${hasBookmark ? 'with a bookmark' : ''}`, () => {
+				dynamicTree.addNodes(
+					1001,
+					[[991, 'new1', mockOuTypes.course], [992, 'new2', mockOuTypes.courseOffering]],
+					hasBookmark,
+					'bookmark'
+				);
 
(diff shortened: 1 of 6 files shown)


test/components/tree-filter.test.js
```diff
@@ -188,30 +188,55 @@ describe('Tree', () => {
...
+			});
 		});
 
-		it('should add the given nodes (constructor)', () => {
-			const tree = new Tree({ nodes, selectedIds, leafTypes, invisibleTypes, isDynamic: true,
-				extraChildren: new Map([
-					[1001, [[991, 'new1', mockOuTypes.course], [992, 'new2', mockOuTypes.courseOffering]]]
-				])
-			});
+		it('should turn off loading for the parent node', () => {
```
anhill-D2L, line 233: Should we also test turning loading on when requesting more children?

anhill-D2L (approved):

Vitalii-Misechko (approved):
LGTM!

PR #90, opened 2020-10-14
Add discussion activity card

rohitvinnakota:
image

[US117451](https://rally1.rallydev.com/#/detail/userstory/396960462964/tasks?fdp=true): [Engagement][Cards] Discussion Activity

Quality Notes
-> Changed the size of the loading spinner so it does not render outside the summary card. The discussion activity card is the only card with a loading spinner. 

LMS PR: https://github.com/Brightspace/lms/pull/2361

Functional testing
Tested the following cases
-> No data returns no chart with a visible legend
-> Data with no data for only 1 or 2 slices does not render those slices as expected 
-> Data changes with filters

Accessibility
->Tested with NVDA. Added chart description and point description for each slice on the pie chart. LTR and RTL works as expected. Tab navigation works as expected. 

Security: n/a
- [x] Perf testing

UX: Demoed with Kevin, ok with design. 

This PR relates to @johngwilkinson's story to add interactions.Currently hovering over slices highlights them but it should be overridden/extended once we add interactions.

fyi: @bemailloux @hyehayes @anhill-D2L @MykolaGalian @NicholasHallman


components/discussion-activity-card.js
```diff
@@ -0,0 +1,168 @@
...
+					colors: ['var(--d2l-color-amethyst-plus-1)', 'var(--d2l-color-zircon-plus-1)', 'var(--d2l-color-zircon)'],
+					borderWidth: 0,
+					cursor: 'pointer',
+					dataLabels: {
+						enabled: false
+					},
+					showInLegend: true,
+					point: {
+						events: {
+							legendItemClick: function() {
```
johngwilkinson, line 114: Why are we disabling a perfectly fine free feature? I just want to make sure this is really what the designer wants.

components/discussion-activity-card.js
```diff
@@ -0,0 +1,168 @@
...
+				text: this._cardTitle, // override default title
+				style: {
+					display: 'none'
+				}
+			},
+			plotOptions: {
+				pie: {
+					size: 80,
+					colors: ['var(--d2l-color-amethyst-plus-1)', 'var(--d2l-color-zircon-plus-1)', 'var(--d2l-color-zircon)'],
+					borderWidth: 0,
```
johngwilkinson, line 106: Is Kevin okay with the hover behaviour? We are going to use a 100% opacity halo for selected slices, so the 10% opacity halo on hover seems undesirable.

model/data.js
```diff
@@ -321,6 +320,23 @@ export class Data {
...
+			return [0, 0, 0];
+		}
+		let threadSum, replySum, readSum;
+		threadSum = replySum = readSum = 0;
+		discussionActivity.forEach(item => {
+			threadSum += item[0];
+			replySum += item[1];
+			readSum += item[2];
+		});
+		return [threadSum, replySum, readSum];
```
johngwilkinson, line 337: Prefer `{ threadSum, replySum, readSum }` so you have named fields instead of an array.

model/data.js
```diff
@@ -321,6 +320,23 @@ export class Data {
 			}, 	new Set()).size;
 	}
 
+	get discussionActivityStats() {
+		const discussionActivity = this.getRecordsInView()
+			.map(record => [record[RECORD.DISCUSSION_ACTIVITY_THREADS], record[RECORD.DISCUSSION_ACTIVITY_REPLIES], record[RECORD.DISCUSSION_ACTIVITY_READS]]);
```
johngwilkinson, line 325: I don't see any advantage to mapping an array to another array. This array isn't any easier to use, and results in you using magic numbers below as indices. Because the data size is potentially large, I wouldn't recommend using an object here, so instead I would just drop this `.map` and use the index names (`RECORD.DISCUSSION...`) below.

model/data.js
```diff
@@ -321,6 +320,23 @@ export class Data {
...
+
+		if (discussionActivity === undefined || discussionActivity.length === 0) {
+			return [0, 0, 0];
+		}
+		let threadSum, replySum, readSum;
+		threadSum = replySum = readSum = 0;
+		discussionActivity.forEach(item => {
```
johngwilkinson, line 332: Is it possible for any of these fields to be `undefined`? Trying it in chrome console, I find that `15 + null` is `15`, but `15 + undefined` is `NaN`.

model/data.js
```diff
@@ -321,6 +320,23 @@ export class Data {
 			}, 	new Set()).size;
 	}
 
+	get discussionActivityStats() {
```
johngwilkinson, line 323: Please inline this method (i.e. move its contents to `DiscussionActivity._discussionActivityStats`). You can add the `computed` decoration there.

johngwilkinson (commented):
Why does this card have a loading spinner if none of the others do? Did you discuss that with Kevin?

General demo question: did you demo live to Kevin (or record some gifs), or just show a snapshot. I find these charts have enough behaviour that seeing them respond to things like mouseover is probably important.

model/data.js
rohitvinnakota: No, it defaults to [0,0,0] server side if there are no items

rohitvinnakota:
> General demo question: did you demo live to Kevin (or record some gifs), or just show a snapshot? I find these charts have enough behaviour that seeing them respond to things like mouseover is probably important.

This user story does not cover any interactions so any of the behavior we have currently is subject to change

components/discussion-activity-card.js
rohitvinnakota: I will reenable this for now and we can take a look again when demoing interactions

components/discussion-activity-card.js
rohitvinnakota: https://github.com/Brightspace/insights-engagement-dashboard/pull/90#issuecomment-710110662

components/discussion-activity-card.js
```diff
@@ -0,0 +1,175 @@
...
+import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import { Localizer } from '../locales/localizer';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { RECORD } from '../consts';
+import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+
+class DiscussionActivityCard extends SkeletonMixin(Localizer(MobxLitElement)) {
+	static get properties() {
+		return {
+			isValueClickable: { type: Boolean, attribute: 'is-value-clickable' }
```
Vitalii-Misechko, line 15: `minor`: I've noticed that `is`-suffix is not used in attribute' names. I'd suggest using `isValueClickable: { type: Boolean, attribute: 'clickable' }`. For instance, `<d2l-insights-discussion-activity-card clickable>` - means the class isValueClickable === true;  in contrast `<d2l-insights-discussion-activity-card is-value-clickable>`  can be read as "you should read value of `is-value-clickable` to know particular value". Maybe is just me )

components/chart/chart.js
```diff
@@ -83,8 +83,8 @@ class Chart extends SkeletonMixin(LitElement) {
 	render() {
 		return html`
 			<div id="chart-container" tabindex="${this.skeleton ? -1 : 0}"></div>
-			<d2l-insights-overlay spinner-size="200" ?loading="${this.skeleton}"></d2l-insights-overlay>
-      	`;
+			<d2l-insights-overlay spinner-size="150" ?loading="${this.skeleton}"></d2l-insights-overlay>
```
Vitalii-Misechko, line 86: `thougts, no changes needed`: It's good that you have shown a demo to Kevin because this value changes spinner size for all charts. As alternative, solution you might want to pass this value as a attribute

components/discussion-activity-card.js
```diff
@@ -0,0 +1,175 @@
...
+	get _chartDescriptionTextLabel() {
+		return this.localize('components.insights-discussion-activity-card.textLabel');
+	}
+
+	render() {
+		// NB: relying on mobx rather than lit-element properties to handle update detection: it will trigger a redraw for
+		// any change to a relevant observed property of the Data object
+		return html`<div class="d2l-insights-discussion-activity-card">
+			<div class="d2l-insights-discussion-activity-card-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this._cardTitle}</div>
+			<div class="d2l-insights-discussion-activity-card-body" aria-hidden="${this.skeleton}">
```
Vitalii-Misechko, line 92: Why does the code need to manage `aria-hidden` property here? I do not see any inner text element that can be read by screen reader while data is loading.

johngwilkinson (approved):

Vitalii-Misechko (approved):
LGTM!

PR #92, opened 2020-10-16
[Engagement][Cards] System Last Access

MykolaGalian:
Added card with a number of users who have no system access in the last 14 days.

Functional test:
Manually invoked the bdp-ingestion-draco-ingester lambda with SystemAccessEvents. Metron POST query takes ~200ms on the AWS side. 
-> Returns data when exists as expected
-> Returns empty array when no data as expected

Devops - N/A
UX and docs - N/A

LMS PR: https://github.com/Brightspace/lms/pull/2304

End-to-end test: done
UX demo:  done

@anhill-D2L @johngwilkinson @rohitvinnakota @hyehayes @NicholasHallman


model/data.js
```diff
@@ -209,6 +209,24 @@ export class Data {
 			});
 	}
 
+	get usersCountsWithLastAccessMoreThanFourteenDays() {
+		const fourteenDayMillis = 1209600000;
+		const userIdsSet = new Set();
```
bemailloux, line 214: Instead of creating your own userIdsSet, you can just use `this.users` directly. It already filters the users by which ones are visible.

So this function can be simplified to
```
return this.users
    .filter(user => user[USER.LAST_SYS_ACCESS] && (Date.now() - user[USER.LAST_SYS_ACCESS] > fourteenDayMillis))
    .length;
```

----------------------------------------

FYI: for a JavaScript Set, you don't have to check if the value exists in the Set before adding it. For example, the following works just fine:
```
s = new Set();
s.add(1);
s.add(2);
s.add(1);
console.log(s); // s = Set(2) {1, 2}

// no need to check `if (!s.has(1))` before doing `s.add(1)` again
```

MykolaGalian: Ok, I just made my own userIdsSet to further use it for the card filter. Like: `this.getRecordsInView('d2l-insights-last-access-card'').forEach(...`

bemailloux (approved):

model/data.js
```diff
@@ -209,6 +209,14 @@ export class Data {
 			});
 	}
 
+	get usersCountsWithLastAccessMoreThanFourteenDays() {
```
johngwilkinson, line 212: please remove this method and move the code to `LastAccessCard._cardValue()`. You can decorate it with `computed` there.

components/last-access-card.js
```diff
@@ -25,7 +27,11 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	get _cardValue() {
-		return this.data.usersCountsWithLastAccessMoreThanFourteenDays;
+		const fourteenDayMillis = 1209600000;
+
+		return this.data.users
+			.filter(user => user[USER.LAST_SYS_ACCESS] && (Date.now() - user[USER.LAST_SYS_ACCESS] > fourteenDayMillis))
```
johngwilkinson, line 33: I'm not 100% sure, but I think users with no last access at all should be counted as having no session in the last 14 days (so the null/undefined case should be included by the filter).

johngwilkinson (approved):

PR #93, opened 2020-10-16
US119012: start moving card logic out of Data class

johngwilkinson:
This
1. Deletes some leftover stuff from the POC, which includes using card filters directly instead of wrapping in the no-op CardFilter class. This change is backward-compatible with the existing card and filter logic.
2. Moves final grade stats and filtering logic out of `Data` and into `current-final-grade.js`. I'm proposing this as a better pattern for all our cards (and maybe the user table), and plan to refactor the others in follow-up PRs.

Quality Notes
* functional (regression, local e2e)
  * select a final grade bar (filter is applied); remove filter by name
  * select final grade bar; remove filter via "clear all"
  * select both a final grade bar and a last access bar; remove one or both filters
* perf: n/a
* devops: n/a
* security: n/a
* ux/docs: n/a


bemailloux (approved):

bemailloux:
I feel like each component file is a bit cluttered now. But that's way better than having a massive data.js file.

johngwilkinson:
> I feel like each component file is a bit cluttered now. But that's way better than having a massive data.js file.

Yeah, the component files could be moved into their own directories and split up a bit.

NicholasHallman:
> I feel like each component file is a bit cluttered now. But that's way better than having a massive data.js file.

I agree, maybe we need something like a "services" layer that does the filtering and data transformation for the cards so we go from

lms -> data -> data transformer -> visualization card ?

then the data transformer can work like a hook where the binning, filtering etc can be isolated from rendering

I think this pr is a good way to clean up the data module and a better pattern to follow.

NicholasHallman (approved):

PR #94, opened 2020-10-16
prevent massive amounts of recalculation

johngwilkinson:
Yikes! Console logging in this getter shows that without the `computed` decoration, it's unsurprisingly getting called many many times per render (it gets called by the filter method for the corresponding card, so once per row, calculate the average of all the rows... so n^2).

FYI @bemailloux @MykolaGalian @NicholasHallman @Vitalii-Misechko

### DIFF
--- a/model/data.js
+++ b/model/data.js
@@ -325,6 +325,7 @@ decorate(Data, {
 	usersCountsWithOverdueAssignments: computed,
 	courseLastAccessDates: computed,
 	tiCVsGrades: computed,
+	tiCVsGradesAvgValues: computed,
 	cardFilters: observable,
 	isLoading: observable,
 	tiCVsGradesQuadrant: observable,


anhill-D2L (approved):

rohitvinnakota (approved):

model/data.js
```diff
@@ -325,6 +325,7 @@ decorate(Data, {
 	usersCountsWithOverdueAssignments: computed,
 	courseLastAccessDates: computed,
 	tiCVsGrades: computed,
+	tiCVsGradesAvgValues: computed,
```
Vitalii-Misechko, line 328: `computed` adds a cache layer and lazy calculations layer to prevent `n^2` iteration, right?

johngwilkinson: Correct.

PR #96, opened 2020-10-19
System Last Access interaction

MykolaGalian:
[US119014](https://rally1.rallydev.com/#/detail/userstory/408910754660?fdp=true): [Engagement] Interactions - System Access

UX demo:  done


components/last-access-card.js
```diff
@@ -1,9 +1,45 @@
...
+
+	filter(record) {
+		return this.usersWithSysLastAccessMoreThanFourteenDays.has(record[RECORD.USER_ID]);
+	}
+}
+
+decorate(LastAccessFilter, {
+	isApplied: computed,
+	addUsersToSet: action,
+	usersWithSysLastAccessMoreThanFourteenDays: observable
```
bemailloux, line 41: I think this should have a shorter, less specific name (since we would have to change the name if it wasn't 14 days anymore). Suggestion: `usersWithoutRecentAccess`

components/last-access-card.js
```diff
@@ -26,12 +62,24 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+
 	get _cardValue() {
+		return this._usersWithSysLastAccessMoreThanFourteenDays.length;
+	}
+
+	get _usersWithSysLastAccessMoreThanFourteenDays() {
 		const fourteenDayMillis = 1209600000;
 
 		return this.data.users
-			.filter(user => user[USER.LAST_SYS_ACCESS] === null || user[USER.LAST_SYS_ACCESS] === undefined || (Date.now() - user[USER.LAST_SYS_ACCESS] > fourteenDayMillis))
```
bemailloux, line 33: Since you mentioned last time about `this.getRecordsInView(<filter id>)`: does this filter's data get cross-filtered by the other filters?

components/last-access-card.js
```diff
@@ -26,12 +62,24 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-			.length;
+			.filter(user => user[USER.LAST_SYS_ACCESS] === null || user[USER.LAST_SYS_ACCESS] === undefined || (Date.now() - user[USER.LAST_SYS_ACCESS] > fourteenDayMillis));
```
bemailloux, line 77: I wonder if you can just use `!user[USER.LAST_SYS_ACCESS]`. It's (almost) equivalent to `user[USER.LAST_SYS_ACCESS] === null || user[USER.LAST_SYS_ACCESS] === undefined || user[USER.LAST_SYS_ACCESS] === 0`

model/data.js
```diff
@@ -330,7 +329,6 @@ decorate(Data, {
 	isLoading: observable,
 	tiCVsGradesQuadrant: observable,
 	selectedLastAccessCategory: observable,
-	selectedGradesCategories: observable,
```
bemailloux, line 333: This was moved to `current-final-grade-card.js`, right?

components/last-access-card.js
MykolaGalian: yes, I need to fix that

model/data.js
MykolaGalian: yes

components/last-access-card.js
```diff
@@ -26,12 +62,24 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+	}
+
+	_addUsersToSet() {
+		const userIds = this._usersWithSysLastAccessMoreThanFourteenDays.map(user => user[USER.ID]);
```
johngwilkinson, line 81: This approach basically caches the selected users at the time the filter is selected. So you need to ask yourself: could the value change without the filter being selected again? The answer is yes, so this way doesn't work.

Scenario:
a. turn on some other filter so only a few users are in view
b. turn on this filter: of the few users in view, those with no recent access are added to the set for this filter
c. turn off the other filter: now there potentially should be more users in view, but the won't appear, because they aren't in the set

The solution will be something like:
* this filter is just a boolean: enabled or not
* the filter method looks up the user for the record and passes if it has an access in the last 14 days

To do that, you may want the `Data` class to provide something like `get allUsersById()` which returns a map from user id to user record. That map could be built when data arrives from the server.

johngwilkinson: Just noticed: there's already a `data._userDictionary` to help with this.

components/last-access-card.js
```diff
@@ -1,9 +1,46 @@
...
+
+	get title() {
+		return 'components.insights-engagement-dashboard.lastSystemAccessHeading';
+	}
+
+	addUsersToSet(userIds) {
+		userIds.forEach(item => this.usersWithoutRecentAccess.add(item));
+	}
+
+	filter(record) {
```
johngwilkinson, line 34: There's no need to build a set of user ids: `data._userDictionary` is already an optimal data structure for this method:
```
filter(record, data) {
  const user = data._userDictionary.get(record[RECORD.USER_ID]);
  return isWithoutRecentAccess(user);
}
```
`isApplied` can simply be an `observable` field on this filter (see `OverdueAssignmentsFilter `)

components/last-access-card.js
```diff
@@ -26,12 +63,24 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 	get _cardValue() {
-		const fourteenDayMillis = 1209600000;
+		return this._usersWithoutRecentAccess.length;
+	}
 
+	get _usersWithoutRecentAccess() {
 		return this.data.users
-			.filter(user => user[USER.LAST_SYS_ACCESS] === null || user[USER.LAST_SYS_ACCESS] === undefined || (Date.now() - user[USER.LAST_SYS_ACCESS] > fourteenDayMillis))
-			.length;
+			.filter(user => !user[USER.LAST_SYS_ACCESS] || ((Date.now() - user[USER.LAST_SYS_ACCESS]) > fourteenDayMillis));
```
johngwilkinson, line 76: It would be useful to pull out a `function isWithoutRecentAccess(user){...}` to pass to this `.filter()`

components/last-access-card.js
```diff
@@ -26,12 +63,24 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+	}
+
+	_addUsersToSetWithoutRecentAccess() {
```
johngwilkinson, line 79: Not needed

components/last-access-card.js
```diff
@@ -50,11 +99,12 @@ class LastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	_valueClickHandler() {
```
johngwilkinson, line 101: just need to set `this.filter.isApplied = true` here

bemailloux (approved):

johngwilkinson (approved):

PR #99, opened 2020-10-19
Us119012/overdue refactor

johngwilkinson:
Regression testing: local e2e; with dummy data; apply filter alone and in combination with others

FYI @anhill-D2L @MykolaGalian @NicholasHallman @Vitalii-Misechko


components/overdue-assignments-card.js
```diff
@@ -32,7 +46,17 @@ class OverdueAssignmentsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 	get _cardValue() {
-		return this.data.usersCountsWithOverdueAssignments;
+		return this.data.getRecordsInView(filterId)
+			.reduce((acc, record) => {
+				if (!acc.has(record[RECORD.USER_ID]) && record[RECORD.OVERDUE] !== 0) {
+					acc.add(record[RECORD.USER_ID]);
+				}
+				return acc;
+			}, 	new Set()).size;
```
bemailloux, line 55: `Out of scope / optional` Readability improvements
```
const usersWithOverdueAssignments = new Set(...this.data.getRecordsInView(filterid)
  .filter(record => record[RECORD.OVERDUE] > 0)
  .map(record => record[RECORD.USER_ID])
);
return usersWithOverdueAssignments.size;
```

test/model/data.test.js
```diff
@@ -358,13 +358,6 @@ describe('Data', () => {
 		});
 	});
 
-	describe('overdueAssignments', () => {
-		it('should return a number of users with overdue assignments', async() => {
```
bemailloux, line 362: Shouldn't there be a corresponding addition for this deletion? I don't see one

johngwilkinson: It's sneaky, but it's there: the overdue-assignments-card tests now use the records this test was using - and happened to already expect the same count.

bemailloux (approved):

rohitvinnakota (approved):

components/overdue-assignments-card.js
Vitalii-Misechko: Interesting, passing `filterId` into `getRecordsInView` we exclude `OverdueAssignmentsFilter`  effect on records. Then we manually apply `OverdueAssignmentsFilter` here. Should we just contract @bemailloux's code? to 
```JS
const usersWithOverdueAssignments = new Set(...this.data.getRecordsInView()
  .map(record => record[RECORD.USER_ID])
);
return usersWithOverdueAssignments.size;
```

PR #98, opened 2020-10-19
US119012: refactor: move last access logic out of Data

johngwilkinson:
Regression testing (local e2e):
  * load page; use role and grade filter and confirm last access chart updates
  * w. dummy data for last access, apply filter and confirm other views update

FYI @anhill-D2L @bemailloux @NicholasHallman @rohitvinnakota

PR #100, opened 2020-10-19
US120286 - Sorting User Table

NicholasHallman:
To keep with @johngwilkinson pattern of moving data filtering logic into the components this change does it's sorting logic in the user-table module. I will need to adapt this pr to #98 (Merged ✅)

Functional Testing
- Unit tests
  - Tests the sorting of all 5 column types in ascending and descending order
- Manual testing with
  - Local LMS data
  - Demo data
- Tested
  - Reacted to existing filters (semester, role, graph data, etc)
  - Clicking the header sorts the fields 
    - Name by last name (regardless of case)
    - Other columns by numerical quantity
  - Clicking the same field a second time flips the sort from ascending to descending

Performance
 - No performance tests conducted, however, js's Array.sort is not as performant as other algorithms, can be migrated in the future if required but I observed no noticeable slow down.
 - The sorting is done on every render (render changes are prompted by state changes to the user data so we need to resort it OR reset to no sorting when the state changes, this implementation assumes a resort)

UX
- The cursor changes to a pointer when hovering over the header
- a up or down arrow is appended to the end of the header td if the column is being sorted
- the arrow flips between the up and down state on additional clicks.

@rohitvinnakota @hyehayes @anhill-D2L @MykolaGalian @bemailloux


components/users-table.js
```diff
@@ -85,15 +86,67 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+	* @param selectedSort
+	*{
+	*	column: number,
+	*	order: 'asc' | 'desc'
+	*}
+	*
+	*/
+	setSelectedSort(selectedSort) {
+		this.selectedSort = selectedSort;
+		this.requestUpdate();
```
johngwilkinson, line 131: with lit, ideally you don't have to call requestUpdate directly; here, making `selectedSort` a property will take care of it

components/users-table.js
```diff
@@ -121,6 +174,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 				title="${this.localize('components.insights-users-table.title')}"
 				.columnInfo=${this.columnInfo}
 				.data="${this._displayData}"
+				.filterColumns="${this.setSelectedSort.bind(this)}"
```
johngwilkinson, line 209: Maybe better to use an event & event-handler for this? The flow is very similar, but it seems more idiomatic.

components/table.js
```diff
@@ -18,6 +18,8 @@ export const COLUMN_TYPES = {
  */
 class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 
+	selectedSort = { column: undefined, order: 'desc' };
```
johngwilkinson, line 21: It would be nice if these were a property, or a pair of properties reflected to attributes - that way you can, for example directly set the sort order based on the desired initial state (which might be saved between sessions later).

components/users-table.js
NicholasHallman: Like a custom event listener and handler instead of a callback? I'll look into it

components/users-table.js
```diff
@@ -94,15 +95,67 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const colmnNumber = this.selectedSort.column;
+		const order = this.selectedSort.order;
+		if (colmnNumber === 0) {
+			this.sortedUserDataForDisplay = [ ...this.userDataForDisplay.sort((a, b) => {
+				if (this._getLastName(a[colmnNumber]) > this._getLastName(b[colmnNumber])) return ORDER[order][0];
+				else if (this._getLastName(a[colmnNumber]) < this._getLastName(b[colmnNumber])) return ORDER[order][1];
+				else ORDER[order][2];
+			})];
+		} else {
+			this.sortedUserDataForDisplay = [ ...this.userDataForDisplay.sort((a, b) => {
```
johngwilkinson, line 151: just noticed the code is sorting twice now: the `sort()` in `userDataForDisplay()` can be removed - or, probably even better, you can do this sort in that method. MobX will make sure it gets recalculated when needed and only then due to the below decoration:
```
decorate(UsersTable, {
	userDataForDisplay: computed
});
```

components/users-table.js
johngwilkinson: Yeah, you can find that pattern in a few of our components here, and it seems generally to work well.

components/table.js
```diff
@@ -168,14 +189,23 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
+
 		const styles = {
 			'd2l-insights-table-cell': true,
+			'd2l-insights-table-cell-header': true,
 			'd2l-insights-table-cell-first': idx === 0,
-			'd2l-insights-table-cell-last': idx === this._numColumns - 1
+			'd2l-insights-table-cell-last': idx === this._numColumns - 1,
+			'd2l-insights-table-arrow-spacing': isSortedColumn
 		};
+		const spaceArrow =  isSortedColumn ? 'd2l-insights-table-cell-sort-indicator' : '';
```
Vitalii-Misechko, line 202: `nit`: you can use `classMap` for this.
```JS
const iconStyles = {
'd2l-insights-table-cell-sort-indicator':  isSortedColumn
}
<d2l-icon ... class="${classMap(iconStyles)}">
```

components/table.js
```diff
@@ -226,5 +256,24 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
+			if (this.sortOrder === 'asc') {
+				this.sortOrder = 'desc';
+			} else {
+				this.sortOrder = 'asc';
+			}
+		}
+
+		this.sortColumn = colmnNumber;
+
+		dispatchEvent(new CustomEvent('columnSort', { detail: { column: this.sortColumn, order: this.sortOrder } }));
```
Vitalii-Misechko, line 276: You should qualify events with the component name. Documentation
https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#events

Vitalii-Misechko: You should use `kebab-case` instead of `camelCase`

components/users-table.js
```diff
@@ -72,6 +75,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		};
 		this._currentPage = 1;
 		this._pageSize = DEFAULT_PAGE_SIZE;
+		addEventListener('columnSort', this.setColumnSort.bind(this));
```
Vitalii-Misechko, line 78: `curious`. Why have this option of event subscription is chosen? Shouldn't we use `<d2l-insights-table @columnSort="${this.setColumnSort}"` instead?

components/users-table.js
```diff
@@ -94,20 +98,74 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+	*{
+	*	column: number,
+	*	order: 'asc' | 'desc'
+	*}
+	*
+	*/
+	setColumnSort(e) {
+		this._sortOrder = e.detail.order;
+		this._sortColumn = e.detail.column;
+		//this.requestUpdate();
```
Vitalii-Misechko, line 135: Should this event handler also reset pagination as well?

components/table.js
```diff
@@ -22,7 +22,9 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 		return {
 			title: { type: String, attribute: true },
 			columnInfo: { type: Array, attribute: false },
-			data: { type: Array, attribute: false }
+			data: { type: Array, attribute: false },
+			sortColumn: { type: Number, attribute: true },
+			sortOrder: { type: String, attribute: true },
```
Vitalii-Misechko, line 27: `minor`: Could you improve documentation above the class declaration as well?

components/users-table.js
NicholasHallman: I had that initially but john commented above with
> Maybe better to use an event & event-handler for this? The flow is very similar, but it seems more idiomatic.

components/users-table.js
NicholasHallman: So sorting by a column should bring you back to the first page?

components/table.js
```diff
@@ -18,13 +18,13 @@ export const COLUMN_TYPES = {
...
 
-	selectedSort = { column: undefined, order: 'desc' };
-
 	static get properties() {
 		return {
 			title: { type: String, attribute: true },
 			columnInfo: { type: Array, attribute: false },
-			data: { type: Array, attribute: false }
+			data: { type: Array, attribute: false },
+			sortColumn: { type: Number, attribute: true },
```
johngwilkinson, line 28: D2L has a convention that attribute names must be lower-kebab-case, so it should be `attribute: 'sort-column'` instead of `attribute: true`.

It's also good practice to add `reflect: true` so that the dom updates if the property is changed by code.

components/table.js
```diff
@@ -151,11 +151,15 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
 		this.title = '';
-		this.filterColumns = () => {};
+		this.sortColumn = 0;
+		this.sortOrder = 'desc';
 	}
 
 	render() {
 		return html`
+			<script type="module">
+				import '@brightspace-ui/core/components/icons/icon.js';
```
johngwilkinson, line 163: I don't fully understand how BSI packaging works, but I believe this should be imported at the top of the file with the other imports, not in the render() method.

components/table.js
```diff
@@ -226,5 +258,24 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
+			if (this.sortOrder === 'asc') {
+				this.sortOrder = 'desc';
+			} else {
+				this.sortOrder = 'asc';
+			}
+		}
+
+		this.sortColumn = colmnNumber;
+
+		dispatchEvent(new CustomEvent('d2l-table-column-sort', { detail: { column: this.sortColumn, order: this.sortOrder } }));
```
johngwilkinson, line 278: `d2l-insights-table-column-sort` (or just `d2l-insights-table-sort`) please - the components in this project are project-specific (at least for now), and could potentially conflict with a more general d2l-table component in future.

components/users-table.js
```diff
@@ -72,6 +75,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		};
 		this._currentPage = 1;
 		this._pageSize = DEFAULT_PAGE_SIZE;
+		addEventListener('d2l-table-column-sort', this.setColumnSort.bind(this));
```
johngwilkinson, line 78: Check https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-add-event-listeners-properly if you haven't. I believe this belongs in the the render template.

components/users-table.js
johngwilkinson: That sounds right to me, but maybe check with Kevin.

johngwilkinson:
BTW, don't forget to add unit tests.

components/table.js
NicholasHallman: Ah that makes more sense thank you!

components/users-table.js
Vitalii-Misechko: Oh...I see. The original comment was made at this piece of code [here ](https://github.com/Brightspace/insights-engagement-dashboard/pull/100#discussion_r508027596). This code passes function into  `filterColumns` property.
```JS
.filterColumns="${this.setSelectedSort.bind(this)}"
```
@johngwilkinson have you meant[ lit-element way of binding event handler](https://lit-element.polymer-project.org/guide/events) there?

components/users-table.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable lit/no-template-bind */
```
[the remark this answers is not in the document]
NicholasHallman: I can remove this exception now since making it an event listener seems to have fixed the binding

test/components/users-table.test.js
```diff
@@ -194,6 +194,109 @@ describe('d2l-insights-users-table', () => {
...
+
+			const table = await fixture(html`<d2l-insights-users-table .data="${data}"></d2l-insights-users-table>`);
+			await new Promise(resolve => setTimeout(resolve, 200));
+			await table.updateComplete;
+
+			userTable = table.shadowRoot.querySelector('d2l-insights-table');
+			headers = userTable.shadowRoot.querySelectorAll('th');
+		});
+
+		const checkIfSorted = (column, order, convertToNumber = false, convertToLower = false, data = undefined) => {
```
johngwilkinson, line 215: This method is a bit complicated because of all the options it covers. It's often preferable to have separate methods instead where you have boolean parameters that result in different code paths - like `checkIfNumbersSorted`, `checkIfLowercaseSorted`, etc. - especially when, as here, you get into weaving back and forth between different options.

Another option overall: have a method to return an array of the sort column values in the correct order, then `expect(actual.map(x => x[column]).to.deep.equal(expected)`. Or some variant of that. Seems easier to follow that the "compare with previous line"  you do here, although it has the disadvantage of being more similar to the code under test.

test/components/users-table.test.js
```diff
@@ -194,6 +194,109 @@ describe('d2l-insights-users-table', () => {
...
+				}
+				if (convertToLower) {
+					record = record.toLowerCase();
+				}
+				if (i > 0) {
+					let previous = records[i - 1];
+					if (convertToNumber) {
+						previous = Number.parseFloat(previous.replace('%', '').trim());
+					}
+					console.log(`Before: ${previous} After: ${record}`);
```
johngwilkinson, line 235: remove

test/components/users-table.test.js
```diff
@@ -195,6 +195,109 @@ describe('d2l-insights-users-table', () => {
...
+					if (order === 'desc' && !(record <= previous)) {
+						inOrder = 'Out of Order';
```
johngwilkinson, line 237: `inOrder` should probably be a boolean - but since this is a validation method, you can replace the conditional here with an `expect`, so it throws as soon as it finds something out of order - then you don't need `inOrder` at all

test/components/users-table.test.js
```diff
@@ -195,6 +195,109 @@ describe('d2l-insights-users-table', () => {
...
+
+			userTable = table.shadowRoot.querySelector('d2l-insights-table');
+			headers = userTable.shadowRoot.querySelectorAll('th');
+		});
+
+		const checkIfSorted = (column, order, convertToNumber = false, convertToLower = false, data = undefined) => {
+
+			const records = data === undefined ? userTable.data.map(user => user[column]) : data;
+
+			console.log(records);
```
johngwilkinson, line 219: remove

test/components/users-table.test.js
NicholasHallman: I had it as a boolean initially but I liked how the tests read when they were a string haha

johngwilkinson: A message string in the assertion works better, I think. `expect(false, 'nooo why out of order??').to.be.true;`

test/components/users-table.test.js
```diff
@@ -212,36 +212,33 @@ describe('d2l-insights-users-table', () => {
...
-			records.forEach((record, i) => {
+		const checkIfSorted = (data, order) => {
 
-				if (convertToNumber) {
-					record = Number.parseFloat(record.replace('%', '').trim());
-				}
-				if (convertToLower) {
-					record = record.toLowerCase();
-				}
+			const records = data;
```
johngwilkinson, line 229: might as well rename the parameter and get rid of this extra variable

test/components/users-table.test.js
```diff
@@ -212,36 +212,33 @@ describe('d2l-insights-users-table', () => {
...
 				if (i > 0) {
-					let previous = records[i - 1];
-					if (convertToNumber) {
-						previous = Number.parseFloat(previous.replace('%', '').trim());
-					}
-					console.log(`Before: ${previous} After: ${record}`);
+					const previous = records[i - 1];
 					if (order === 'desc' && !(record <= previous)) {
-						inOrder = 'Out of Order';
+						expect(false).to.equal(true);
```
johngwilkinson, line 234: If you just want to trigger an exception, you can `throw new Error('out of order')` or the like - but I was thinking you could use the assertion library, like `if(order === 'desc') expect(record <= previous, 'out of order').to.be.true;` (there's syntax for numeric comparisons: `expect(record).to.be.at.most(previous)`, but I don't know off-hand if there's an equivalent for strings)

NicholasHallman: Oh that's cool! thanks for the guidance I will read up more on this

johngwilkinson (approved):
Looks good to me!

test/components/users-table.test.js
```diff
@@ -195,6 +195,122 @@ describe('d2l-insights-users-table', () => {
 			});
 		});
 	});
+
+	describe('sorting', () => {
```
Vitalii-Misechko, line 199: Tests look neat!

Vitalii-Misechko (approved):

PR #105, opened 2020-10-21
Us119012/filter refactor

johngwilkinson:
Note: this touches a lot of files, but in most the changes are quite small. The main changes are
* split `FilteredData` out of `Data`
  * notice filters now get passed just the record and the user dictionary: this means if they need access to all the data, they need to get it through their constructor - which forces us to be explicit about when that is happening and which filters are involved
* move tic/grades logic into the tic/grades file (mostly in the filter)
* tests for those classes

Regression testing:
  * unit tests
- [x]  local e2e with various filter combinations
- [ ] check that mobx doesn't recalculate due to `.excluding(filterId)`

FYI @MykolaGalian @NicholasHallman @rohitvinnakota @Vitalii-Misechko


components/time-in-content-vs-grade-card.js
```diff
@@ -1,34 +1,91 @@
...
-		} else if (data.tiCVsGradesQuadrant === 'rightBottom') {
-			result =  record[RECORD.TIME_IN_CONTENT] >= data.tiCVsGradesAvgValues[0] * 60 && record[RECORD.CURRENT_FINAL_GRADE] !== null && record[RECORD.CURRENT_FINAL_GRADE] < data.tiCVsGradesAvgValues[1];
-		} else (result = false);
-		return result;
+const filterId = 'd2l-insights-time-in-content-vs-grade-card';
+
+const TIC = 0;
+const GRADE = 1;
+
+function avgOf(items, field) {
```
anhill-D2L, line 16: users-table.js also has an `avgOf` function, I'm not sure if this is something that could/should also be refactored 🤷

model/filteredData.js
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * @param {Object}filter - a filter must have fields id, title, and isApplied,
+	 * and a filter(record, userDictionary) method; beyond that, it can keep state however it wishes.
+	 * Ideally, these should be classes and the id need not
+	 * be known outside the defining file (see, e.g., current-final-grade-card).
+	 * @returns {FilteredData}
+	 */
+	filter(filter) {
+		return new FilteredData(this._data, [...this.filters, filter]);
+	}
```
bemailloux, line 32: I'm a little unclear on the difference between `FilteredData.filter vs filter.isApplied == true` and `FilteredData.excluding vs filter.isApplied == false`. 

As far as I understand it's because `isApplied` lets _other_ filters know that this filter should be applied, whereas `excluding` is so that individual filters don't filter themselves when filtering. Is that kind of the idea?

model/filteredData.js
```diff
@@ -0,0 +1,87 @@
...
+
+	get recordsByUser() {
+		const recordsByUser = new Map();
+		this.records.forEach(r => {
+			if (!recordsByUser.has(r[RECORD.USER_ID])) {
+				recordsByUser.set(r[RECORD.USER_ID], []);
+			}
+			recordsByUser.get(r[RECORD.USER_ID]).push(r);
+		});
```
bemailloux, line 59: Does mobx's `computed` work correctly when we're using immutable objects? I'm wondering: if you run something like `filteredData.excluding('some-filter').recordsByUser` twice in a row, does the value get recomputed, or does mobx optimize it?

engagement-dashboard.js
```diff
@@ -144,37 +145,46 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
...
-
-			this.__data = new Data({
-				recordProvider: this.isDemo ? fetchDemoData : fetchData,
-				cardFilters: cardFilters
-			});
+			// There are row filters - which look at each record individually when deciding to include or
+			// exclude it - and there are aggregate filters, which also look at all records selected by
+			// other filters (e.g. to include a record if it has an above-average value in some field).
+			// Aggregate filters are potentially ambiguous if there are more than one and each depends
+			// on the results of the other: we avoid this by building them on specific sets of filters.
```
bemailloux, line 152: For my understanding: this says that we apply all single-row filters first, then aggregate filters after that?

components/time-in-content-vs-grade-card.js
```diff
@@ -1,34 +1,91 @@
...
+	get isApplied() {
+		return this.quadrant !== null;
+	}
+
+	set isApplied(isApplied) {
+		if (!isApplied) this.quadrant = null;
+	}
+
+	get tiCVsGrades() {
+		return this._data.records
```
bemailloux, line 47: Why doesn't this filter need to do `this._data.excluding(filterId)` like the other filters? Something to do with "row" filters vs "aggregate" filters?

test/model/filteredData.test.js
```diff
@@ -0,0 +1,102 @@
...
+		]),
+		isLoading: true
+	};
+
+	let sut;
+	beforeEach(() => {
+		sut = new FilteredData(serverData)
+			.filter(new TestFilter('f1', r => r[4] > 0))
+			.filter(new TestFilter('f2', r => r[0] < 10))
+			.filter(new TestFilter('f3', r => r[1] === 200));
```
bemailloux, line 38: These are all row filters. I think it might be useful to have aggregate filter tests as well, but that's more of an integration level test (whereas this is definitely a unit test)

components/time-in-content-vs-grade-card.js
johngwilkinson: I considered it, but they are a bit different, so it seemed premature to try and combine them.

model/filteredData.js
johngwilkinson: That's accurate, but I think of it a bit differently: `isApplied` is state that the user can control; `excluding` is for wiring up the application - it means 'give me a view of the data passed through all applied filters but this one'. If you can think of a way I can clarify that in the code, I'm interested.

model/filteredData.js
johngwilkinson: It should, but I'll test that.

engagement-dashboard.js
johngwilkinson: Roughly, yes: the tic/grades filter depends on the FilteredData including all the other filters, i.e. the answer to the question "do I include this record" depends on the record as well as the entire FilteredData with the row filters.

If we had another aggregate filter, say, "below-average discussion engagement", we would have to decide on a dependency order: does the discussion engagement filter depend on the tic/grades filter or visa versa? The code design here tries to force you to make that decision explicit by building a FilteredData on one of them, then passing it into the constructor for the other.

That said, the FilteredData that results here includes all the filters, so graphs and summary numbers don't have any notion of filter ordering: so long as the filters are defined unambiguously (which we achieve by dependency ordering on aggregate filters), each component can show the data obtained by applying all filters excluding itself.

components/time-in-content-vs-grade-card.js
johngwilkinson: Because we are in the filter class itself here: `this._data` can't include this filter, because it was used to build this filter. It's the one with row filters only.

model/filteredData.js
bemailloux: That makes more sense. I would have suggested `FIlteredData.filter -> withFilter` just to keep in line with the Java conventions from SSR, but that doesn't help distinguish between "user-controlled filter" and "permanent exclusion".

johngwilkinson: I like `withFilter`. And maybe `withoutFilter` instead of `excluding`.

model/filteredData.js
bemailloux: Maybe it doesn't matter so much - if someone changes any filters it will have to be recalculated anyway.

johngwilkinson: Ah, nvm: it's not an issue at all because methods that call `withoutFilter` are decorated with "computed". I did a better test where I alter some other rendered property, and `get records` never gets called again.

components/time-in-content-vs-grade-card.js
bemailloux: Ah ok, I didn't notice the slight difference in behaviour between this card and other cards. Other cards exclude their corresponding filter, then apply the filter function themselves. This card doesn't operate on the data itself: it gets its data from the filter class.

johngwilkinson: Yeah, the "filter" class in this case is really a full data model, because it calculates the averages that are needed for filtering and also for rendering.

model/filteredData.js
johngwilkinson: done

bemailloux (approved):

anhill-D2L (approved):

PR #108, opened 2020-10-22
Us119012/discussion pie interaction

johngwilkinson:
Pulled out a `CategoryFilter` class, as this is the third place needing essentially the same class (filter by category match).

I couldn't find a way to set the pie chart legend aria labels except by highcharts global options. This works for this dashboard because we only have one chart with a legend.

Quality Notes
* functional (local e2e)
  * select various combinations of threads, read, and replies: other cards and table update accordingly
  * select by clicking on pie slices and by clicking on legend; toggle on and off
  * select combinations of several card filters; clear discussion filter directly and via "clear all"
  * select a slice before there is any data via legend; clear ou filter so there is data: slice is selected as expected
* security: n/a
* perf: n/a
* devops: n/a
* ux/docs:
  * designer-approved
  * keyboard navigation to and within this chart is functional with and without screen reader (MS Narrator)

FYI @anhill-D2L @bemailloux @MykolaGalian @NicholasHallman 

final-pie


model/categoryFilter.js
```diff
@@ -0,0 +1,41 @@
...
+		this.title = title;
+		this.filter = filter;
+		this.selectedCategories = new Set();
+	}
+
+	get isApplied() {
+		return this.selectedCategories.size > 0;
+	}
+
+	set isApplied(isApplied) {
```
Vitalii-Misechko, line 15: So, `filter.IsApplied = true` does nothing, right? 
Should we throw an error in this case to help diagnose wrong usage?

Vitalii-Misechko:
> keyboard navigation to and within this chart is functional with and without screen reader (MS Narrator)

This is cool that hightcharts allows to select a section of pie chart with keyboard

Vitalii-Misechko (approved):
LGTM!

model/categoryFilter.js
johngwilkinson: I considered it, but I wouldn't want to break the page for that. It would be great if we could easily send a warning that goes in (for example) the LMS log.

rohitvinnakota (approved):

PR #114, opened 2020-10-26
[Testing in progress] Us117436/users table selection column

bemailloux:
Quality notes
* Testing [in progress]
  * [ ] Browsers
  * [ ] Responsive
  * [ ] Screen reader
  * [ ] Keyboard nav
    * [ ] Enter key does not trigger selection or deselection
  * [ ] Test cases
    * Selection (one row at a time)
      * Can select multiple
      * Can deselect
      * Only selected values are present in `UsersTable.selectedUserIds`
    * Select all
      * Selects all if none are selected
      * Selects all if 1 is selected
      * Selects all if multiple but not all are selected
      * Deselects all if all are selected
      * Selections reflected in `UsersTable.selectedUserIds`
      * Not possible to sort by selection column
    * State persistence
      * `selectedUserIds` is reset when data changes
      * `selectedUserIds` is reset when page number changes
      * `selectedUserIds` is reset when page size changes
      * `selectedUserIds` is reset when sort options change
    * Dialog
      * Warning appears if `selectedUserIds` is empty
* Security, devops: N/A
* UX: nothing other than testing notes above
* Perf: 
  * Any selection causes `userDataForDisplay` to be recalculated. Perf seems ok for 20 users, haven't tried with more yet
  * Select all is processed once, instead of <# users per page> times
  * [ ] Perf test with 100 users per page
* Docs:
  * [ ] mention somewhere that only 1 row selector column is allowed

FYI @rohitvinnakota @MykolaGalian @Vitalii-Misechko @anhill-D2L


components/users-table.js
```diff
@@ -174,16 +214,21 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		// then sort by lastFirstName
 		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
 		return this.data.users
-			.map(this._preProcessData.bind(this))
-			.sort(sortFunction.bind(this))
-			.map(this._formatDataForDisplay.bind(this));
+			.map(this._preProcessData, this)
+			.sort(sortFunction)
+			.map(this._formatDataForDisplay, this);
```
bemailloux, line 219: FYI @NicholasHallman lots of array functions provide a `thisArg` option specifically so you don't have to bind `this` to a function. It's just a bit neater - `bind` will literally create a new function while `thisArg` won't.

engagement-dashboard.js
```diff
@@ -221,5 +252,18 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		event.stopPropagation();
 		this._serverData.selectedSemesterIds = event.target.selected;
 	}
+
+	_handleEmailButtonPress() {
+		const usersTable = this.shadowRoot.querySelector('d2l-insights-users-table');
+		const selectedUserIds = usersTable.selectedUserIds;
```
johngwilkinson, line 258: @MykolaGalian here's an example of getting data from a component for use elsewhere - rather similar to what the export button needs to do.

johngwilkinson (approved):

NicholasHallman (approved):
Looks great! thanks for the feedback on the user table sort feature 😃

PR #115, opened 2020-10-27
us117433: csv export

MykolaGalian:
[US117433](https://rally1.rallydev.com/#/detail/userstory/396954111816?fdp=true): [Engagement][Actions] Export

### DIFF
--- a/components/users-table.js
+++ b/components/users-table.js
@@ -11,7 +11,7 @@ import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
-const TABLE_USER = {
+export const TABLE_USER = {
 	NAME_INFO: 0,
 	COURSES: 1,
 	AVG_GRADE: 2,
@@ -40,6 +40,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 	static get properties() {
 		return {
 			data: { type: Object, attribute: false },
+			exportData: { type: Object, attribute: false },
 			_currentPage: { type: Number, attribute: false },
 			_pageSize: { type: Number, attribute: false },
 			_sortColumn: { type: Number, attribute: false },
@@ -74,6 +75,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		this.data = {
 			users: []
 		};
+		this.exportData = {};
 		this._currentPage = 1;
 		this._pageSize = DEFAULT_PAGE_SIZE;
 		this._sortOrder = 'desc';
@@ -182,7 +184,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	setDataForExport(userData) {
-		this.data._data.dataForExport = userData;
+		this.exportData.setData(userData);
 	}
 
 	get columnInfo() {
--- a/components/users-table.js
+++ b/components/users-table.js
@@ -182,7 +182,7 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	setDataForExport(userData) {
-		this.data.dataForExport = userData;
+		this.data._data.dataForExport = userData;
 	}
 
 	get columnInfo() {
--- a/components/users-table.js
+++ b/components/users-table.js
@@ -1,7 +1,7 @@
 import '@brightspace-ui/core/components/inputs/input-text';
 import '@brightspace-ui-labs/pagination/pagination';
 import './table.js';
-import { computed, decorate, observable } from 'mobx';
+import { action, computed, decorate, observable } from 'mobx';
 import { css, html } from 'lit-element';
 import { formatNumber, formatPercent } from '@brightspace-ui/intl';
 import { RECORD, USER } from '../consts';
@@ -173,10 +173,16 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		// map to a 2D userData array, with column 0 as a sub-array of [lastFirstName, username - id]
 		// then sort by lastFirstName
 		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
-		return this.data.users
+		const userData = this.data.users
 			.map(this._preProcessData.bind(this))
 			.sort(sortFunction.bind(this))
 			.map(this._formatDataForDisplay.bind(this));
+		this.setDataForExport(userData);
+		return userData;
+	}
+
+	setDataForExport(userData) {
+		this.data.dataForExport = userData;
 	}
 
 	get columnInfo() {
@@ -264,5 +270,6 @@ decorate(UsersTable, {
 	userDataForDisplay: computed,
 	_sortColumn: observable,
 	_sortOrder: observable,
+	setDataForExport: action
 });
 customElements.define('d2l-insights-users-table', UsersTable);
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -29,6 +29,7 @@ import { Localizer } from './locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { OverdueAssignmentsFilter } from './components/overdue-assignments-card';
 import { TimeInContentVsGradeFilter } from './components/time-in-content-vs-grade-card';
+import { toJS } from 'mobx';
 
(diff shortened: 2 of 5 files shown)


model/exportData.js
```diff
@@ -0,0 +1,55 @@
+import { action, decorate, observable, toJS } from 'mobx';
+import { TABLE_USER } from '../components/users-table';
+
+export class ExportData {
+	constructor() {
+		this.data = [];
+	}
+
+	getData() {
+		return toJS(this.data).map(item => {
```
bemailloux, line 10: TIL: `toJS` exists. Cool!

model/exportData.js
```diff
@@ -0,0 +1,55 @@
...
+				`"${item[TABLE_USER.COURSES]}"`,
+				`"${item[TABLE_USER.AVG_GRADE]}"`,
+				`"${item[TABLE_USER.AVG_TIME_IN_CONTENT]}"`,
+				`"${item[TABLE_USER.LAST_ACCESSED_SYS]}"`];
+		});
+	}
+
+	arrayToCsv() {
+		const csvRows = [];
+		const headers = ['FirstName', 'LastName', 'UserName', 'UserID', 'Courses', 'AvgGrade', 'AvgTimeInContent', 'lastAccessedSys'];
```
bemailloux, line 27: You should ask Megan or Kevin if these column names are ok

model/exportData.js
```diff
@@ -0,0 +1,55 @@
...
+		csvRows.push(headers.join(','));
+
+		for (const row of this.getData()) {
+			csvRows.push(row.join(','));
+		}
+		return csvRows.join('\n');
```
bemailloux, line 33: In general it's a good idea to use a CSV library instead of making the CSV yourself. The [CSV spec](https://tools.ietf.org/html/rfc4180#section-2) has lots of edge cases you wouldn't think of, like "what happens if there's a comma or double-quotes in the field". 

I don't know if D2L chose a CSV library to use for front-end. Probably worth checking with team Gaudi.

model/exportData.js
```diff
@@ -0,0 +1,55 @@
...
+	downloadCsv() {
+		const blob = new Blob([this.arrayToCsv()], { type: 'text/csv' });
+		const url = window.URL.createObjectURL(blob);
+		const a = document.createElement('a');
+		a.setAttribute('hidden', '');
+		a.setAttribute('href', url);
+		a.setAttribute('download', 'engagement.csv');
+		document.body.appendChild(a);
+		a.click();
+		document.body.removeChild(a);
```
bemailloux, line 45: Instead of adding this element using JS, you should just put it inside the export button.

model/exportData.js
MykolaGalian: I asked, but they have no solution from d2l. I will do it manually, the main thing is to remove all quotes found in the element, and then wrap each element in quotes.

model/exportData.js
johngwilkinson: And should they be localized?

MykolaGalian: Already done

model/exportData.js
johngwilkinson: Have you looked for a module in npm that does this?

components/users-table.js
```diff
@@ -173,10 +175,29 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		// map to a 2D userData array, with column 0 as a sub-array of [lastFirstName, username - id]
 		// then sort by lastFirstName
 		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
-		return this.data.users
+		const userData = this.data.users
 			.map(this._preProcessData.bind(this))
 			.sort(sortFunction.bind(this))
 			.map(this._formatDataForDisplay.bind(this));
+		this.setDataForExport(userData);
```
johngwilkinson, line 182: Getters should not have side-effects. Instead, add a getter `get dataForExport()` which uses `this.userDataForDisplay` and does the formatting.

Then, instead of passing an `ExportData` object to this component, give a reference to this component to your export button; if that doesn't go well, make a `UserTableData` class in this file that takes the `FitleredData` object and has the necessary getters, and pass it to both this component and the export button.

components/users-table.js
```diff
@@ -88,11 +93,16 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		this._sortOrder = 'desc';
 		this._sortColumn = TABLE_USER.NAME_INFO;
 		this.selectedUserIds = [];
+		this.dataForExport = [];
```
johngwilkinson, line 96: convert these to getters please: there's no need for `_setExportData()`

engagement-dashboard.js
```diff
@@ -200,6 +203,13 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		`;
 	}
 
+	_exportToCsv() {
+		const usersTable = this.shadowRoot.querySelector('d2l-insights-users-table');
+		const formattedData = ExportData.dataFormatter(usersTable.dataForExport);
+		const arrayToCsv = ExportData.arrayToCsv(formattedData, usersTable.headersForExport);
```
johngwilkinson, line 209: This looks good. Maybe move this line into `dataFormatter` (and rename something like `userDataToCsv`).

model/exportData.js
```diff
@@ -0,0 +1,31 @@
+import exportFromJSON from 'export-from-json';
+import { TABLE_USER } from '../components/users-table';
+import { toJS } from 'mobx';
+
+export class ExportData {
+
+	static userDataToCsv(data, headers) {
+		const formattedData = toJS(data).map(item => {
+			const lastFirstName = item[TABLE_USER.NAME_INFO][0].split(', ');
```
johngwilkinson, line 9: This is looking pretty good. One thing I'd do is rearrange the user-table code a little so that this method doesn't have to parse out the user name and id. This is error-prone - consider, for example, what happens if a user has a ' - ' in their name. The best way to do this safely is to have a method on user-table that returns the actual values *before* they get concatenated into strings. Then the table-display code can do the concatenation last instead of early on.

model/exportData.js
```diff
@@ -0,0 +1,30 @@
...
+		});
+
+		return this.exportToCsv(formattedData, headers);
+	}
+
+	static exportToCsv(formattedData, headers) {
+		const jsonArrayData = formattedData.map((data) => {
+			return headers.reduce((acc, val, i) => ({ ...acc, [val]: data[i] }), {});
+		});
+		exportFromJSON({ data: jsonArrayData, fileName: 'engagement', exportType: exportFromJSON.types.csv });
```
johngwilkinson, line 28: Do the columns appear in the csv in the desired order? I noticed there was an issue about this on the export-from-json lib, but I didn't look into it.

MykolaGalian: No, it's all ok with sorting

johngwilkinson: I'm asking about column order, not row order. https://github.com/zheeeng/export-from-json/issues/60

MykolaGalian: As you can see from the link the column order does not change

components/users-table.js
```diff
@@ -53,7 +55,9 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 			_pageSize: { type: Number, attribute: false },
 			_sortColumn: { type: Number, attribute: false },
 			_sortOrder: { type: String, attribute: false },
-			selectedUserIds: { type: Array, attribute: false }
+			selectedUserIds: { type: Array, attribute: false },
+			dataForExport: { type: Array, attribute: false },
```
johngwilkinson, line 59: These two getters are not used in rendering, so they don't need to be declared here.

johngwilkinson (commented):
Can you unit test the new getters?

Also looking for cross-browser testing.

components/users-table.js
```diff
@@ -215,15 +218,50 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		// map to a 2D userData array, with column 1 as a sub-array of [id, FirstName, LastName, UserName]
+		// then sort by 'lastName, FirstName'
 		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
 		return this.data.users
 			.map(this._preProcessData, this)
 			.sort(sortFunction)
 			.map(this._formatDataForDisplay, this);
 	}
 
+	get userDataForDisplayFormatted() {
```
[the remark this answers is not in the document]
MykolaGalian: We need to do a larger refactor around preprocessing/sorting in the next PRs

PR #119, opened 2020-10-28
Enable all card animations

NicholasHallman:
- Changed easing function to quadric easeOut
- Added reduced motion accessibility
- Converted histogram chart to column chart to fix animation problems (may be due to multiple data series?)

The grade card was converted from a histogram to a column chart because of animation troubles. Couldn't get them working for most of yesterday and decided to use Highcharts second example for histograms. 

Testing
 - Verified using unit tests 
- And Manual testing to make sure that converting the table didn't break any of the mobx state management.
 Performance
 - N/A css and svg animations are hardware accelerated, and they don't run very often
UX
- Pending Demo, showed kevin propositions for different easing functions.

@johngwilkinson @anhill-D2L @bemailloux @rohitvinnakota @hyehayes

### DIFF
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -6,6 +6,11 @@ import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 import { css, html, LitElement } from 'lit-element';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
+const easeOut = (t) => (1 - (--t) * t * t * t);
+const defaultAnimationOptions = {
+	duration: 500,
+	easing: easeOut,
+};
 // use default highcharts's template but with <h3> instead of <h4> to fix axe heading-order error
 export const BEFORE_CHART_FORMAT = '<h3>{chartTitle}</h3>' +
 	'<div>{typeDescription}</div>' +
@@ -57,6 +62,11 @@ class Chart extends SkeletonMixin(LitElement) {
 		 */
 		this.immutable = false;
 		/**
+		 * Checks accessibility preferences and enables or disables them accordingly.
+		 */
+		this.showAnimations = !window.matchMedia('(prefers-reduced-motion: reduce)').matches ?
+			defaultAnimationOptions : false;
+		/**
 		 * Array of update()'s function optional arguments.
 		 * Parameters should be defined in the same order like in
 		 * native Highcharts function: [redraw, oneToOne, animation].
@@ -65,7 +75,7 @@ class Chart extends SkeletonMixin(LitElement) {
 		this.updateArgs = [
 			true,
 			true,
-			true
+			this.showAnimations
 		];
 		this.globalOptions = null;
 	}
@@ -80,6 +90,10 @@ class Chart extends SkeletonMixin(LitElement) {
 				display: none;
 			}
 		`;
+	}
+
+	defaultOptions() {
+		this.options.plotOptions.series.animation = this.showAnimations
 	}
 
 	render() {
@@ -116,6 +130,8 @@ class Chart extends SkeletonMixin(LitElement) {
 			console.warn('The "options" property was not passed.');
 		}
 		else {
+			// accessible options overload
+			this.defaultOptions();
 			// Create a chart
 			H.setOptions({
 				lang: {
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -93,7 +93,14 @@ class Chart extends SkeletonMixin(LitElement) {
 	}
 
 	defaultOptions() {
-		this.options.plotOptions.series.animation = this.showAnimations
+		// check for series and plotOptions because we could have a column instead.
+		if (this.options.plotOptions === undefined) return;
+		else if (this.options.plotOptions.series !== undefined) {
+			this.options.plotOptions.series.animation = this.showAnimations;
+		}
+		else if (this.options.plotOptions.column !== undefined) {
+			this.options.plotOptions.column.animation = this.showAnimations;
+		}
 	}
 
 	render() {
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -92,13 +92,15 @@ class Chart extends SkeletonMixin(LitElement) {
 		`;
 	}
 
-	defaultOptions() {
+	setDefaultOptions() {
 		// check for series and plotOptions because we could have a column instead.
-		if (this.options.plotOptions === undefined) return;
-		else if (this.options.plotOptions.series !== undefined) {
+		if (!this.options.plotOptions) {
+			return;
+		}
+		else if (this.options.plotOptions.series) {
 			this.options.plotOptions.series.animation = this.showAnimations;
 		}
-		else if (this.options.plotOptions.column !== undefined) {
+		else if (this.options.plotOptions.column) {
 			this.options.plotOptions.column.animation = this.showAnimations;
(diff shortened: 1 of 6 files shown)


components/chart/chart.js
```diff
@@ -56,6 +61,11 @@ class Chart extends SkeletonMixin(LitElement) {
 		 * useful in some cases but slower than a regular update.
 		 */
 		this.immutable = false;
+		/**
+		 * Checks accessibility preferences and enables or disables them accordingly.
+		 */
+		this.showAnimations = !window.matchMedia('(prefers-reduced-motion: reduce)').matches ?
+			defaultAnimationOptions : false;
```
Vitalii-Misechko, line 68: nice!

components/chart/chart.js
```diff
@@ -6,6 +6,11 @@ import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 import { css, html, LitElement } from 'lit-element';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
+const easeOut = (t) => (1 - (--t) * t * t * t);
```
Vitalii-Misechko, line 9: `curious`: what does this math mean?

components/chart/chart.js
```diff
@@ -82,6 +92,17 @@ class Chart extends SkeletonMixin(LitElement) {
 		`;
 	}
 
+	defaultOptions() {
```
Vitalii-Misechko, line 95: `suggestion`: defaultOptions => setDefaultOptions

components/chart/chart.js
```diff
@@ -82,6 +92,17 @@ class Chart extends SkeletonMixin(LitElement) {
...
+		// check for series and plotOptions because we could have a column instead.
+		if (this.options.plotOptions === undefined) return;
```
Vitalii-Misechko, line 97: `nit` curly brases would be good here due to this statement can be confused with guard clause which does not have `else` part.

Vitalii-Misechko: Or this statment can bu turned into a guard clause

Vitalii-Misechko: `nit`: ckekning truthiness or falsiness can make statements more concise. 
```JS
if (!this.options.plotOptions) return;

if (this.options.plotOptions.series) {...}
```

components/course-last-access-card.js
```diff
@@ -286,7 +285,6 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 				series: {
 					minPointLength: 2, // visualize 0 points
 					pointStart: 0,
-					animation: false,
 					pointWidth: 16,
```
Vitalii-Misechko, line 288: Interesting, what kind of animation it was preventing and why? I dont see any difference in behaviour between master and PR branch.

components/current-final-grade-card.js
```diff
@@ -86,11 +86,14 @@ class CurrentFinalGradeCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const bins = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+		this.data.withoutFilter(filterId).records
 			.filter(record => record[RECORD.CURRENT_FINAL_GRADE] !== null && record[RECORD.CURRENT_FINAL_GRADE] !== undefined)
 			.map(record => [record[RECORD.TIME_IN_CONTENT], record[RECORD.CURRENT_FINAL_GRADE]])
 			.filter(item => item[0] || item[1])
-			.map(item => gradeCategory(item[1]));
+			.map(item => gradeCategory(item[1]))
+			.forEach(gradeCategory => bins[gradeCategory / 10] += 1);
+		return bins;
 	}
```
Vitalii-Misechko, line 97: Could you double check if code in `_colours` is valid for column chars as well? I see the comment at line 105 that makes some assumption regarding coloring histogram chart. `Math.min` does not return minimal category anymore. I suspect we might see unexpected results when each bin has at least 10 records

components/chart/chart.js
NicholasHallman: The default animation timing is linear, like y = x, this feels unnatural because in the real world nothing starts and stops with no change in momentum. This math changes the relationship so that the momentum starts quickly, and deaccelerates as the value t approaches one, creating a quadratic curve. This eases out the motion and makes it look more natural [Example](https://d2lmail-my.sharepoint.com/personal/nhallman_desire2learn_com/Documents/Documents/ShareX/Screenshots/2020-10/easeout.gif)

components/course-last-access-card.js
NicholasHallman: the series animation is the initial render, which may not show if the card is too far down the page

components/chart/chart.js
Vitalii-Misechko: cool!

components/current-final-grade-card.js
johngwilkinson: Yes, _colours can follow the pattern in last course access now, I think.

I actually wanted to convert to a column chart when doing that fix: the histogram module in HC really doesn't buy us much, and because the rendered series is calculated after card render, it makes it awkward to handle things like colours. So I'm glad to see this. :)

components/chart/chart.js
```diff
@@ -92,13 +92,15 @@ class Chart extends SkeletonMixin(LitElement) {
 		`;
 	}
 
-	defaultOptions() {
+	setDefaultOptions() {
 		// check for series and plotOptions because we could have a column instead.
-		if (this.options.plotOptions === undefined) return;
-		else if (this.options.plotOptions.series !== undefined) {
+		if (!this.options.plotOptions) {
```
NicholasHallman, line 97: this caused the regression during standup, forgot the not, fixed with this commit

components/current-final-grade-card.js
```diff
@@ -102,12 +103,9 @@ class CurrentFinalGradeCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 		}
 
-		// the histogram module only renders zeroes for points between non-zero points,
-		// so we provide colours for all points starting with the minimum non-zero value
-		// (extra colours will be ignored)
-		const min = Math.min(...data);
+		// Go through all of the bins and assign the correct color.
 		return [0, 10, 20, 30, 40, 50, 60, 70, 80, 90].map(category =>
-			(this.category.has(category + min) ?
+			(this.category.has(category) ?
```
NicholasHallman, line 108: _colors now simply goes through all the bins and checks if it's in the filter

test/components/current-final-grade-card.test.js
```diff
@@ -41,7 +41,7 @@ describe('d2l-insights-current-final-grade-card', () => {
 			filter.selectCategory(50);
 			const el = await fixture(html`<d2l-insights-current-final-grade-card .data="${data}"></d2l-insights-current-final-grade-card>`);
 			// only the first 8 colours matter (from 20 to 90)
-			expect(el._colours.slice(0, 8)).to.deep.equal([
+			expect(el._colours.slice(2, 10)).to.deep.equal([
```
Vitalii-Misechko, line 44: > _colors now simply goes through all the bins and checks if it's in the filter

Having this in mind `slice` looks rather confusing now. Could you just remove it?

Vitalii-Misechko (approved):

Vitalii-Misechko (approved):

MykolaGalian (approved):

PR #121, opened 2020-10-29
US120050: screen reader support for summary cards

johngwilkinson:
Quality Notes:
* functional
  * w. Narrator and keyboard, navigate page
    * tabbing to clickable summary number reads all the card text
    * using up and down arrows to reach both clickable and non-clickable summary numbers reads all the card text
    * screen reader does not read out any duplicate text in the summary body
    * summary cards with filters can be enabled using the keyboard
* devops, perf, ux/docs, security: n/a

FYI @anhill-D2L @MykolaGalian @NicholasHallman @rohitvinnakota

### DIFF
--- a/components/summary-card.js
+++ b/components/summary-card.js
@@ -1,6 +1,8 @@
+import '@brightspace-ui/core/components/offscreen/offscreen.js';
 import './overlay';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
+import { Localizer } from '../locales/localizer';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
 
 /**
@@ -9,13 +11,13 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
  * @property {string} message
  * @fires d2l-labs-summary-card-value-click
  */
-class SummaryCard extends SkeletonMixin(LitElement) {
+class SummaryCard extends SkeletonMixin(Localizer(LitElement)) {
 	static get properties() {
 		return {
 			title: { type: String, attribute: 'card-title' },
 			value: { type: String, attribute: 'card-value' },
 			message: { type: String, attribute: 'card-message' },
-			isValueClickable: { type: Boolean, attribute: 'is-value-clickable' }
+			isValueClickable: { type: Boolean, attribute: 'value-clickable' }
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


components/summary-card.js
```diff
@@ -9,13 +11,13 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
...
  */
-class SummaryCard extends SkeletonMixin(LitElement) {
+class SummaryCard extends SkeletonMixin(Localizer(LitElement)) {
 	static get properties() {
 		return {
 			title: { type: String, attribute: 'card-title' },
 			value: { type: String, attribute: 'card-value' },
 			message: { type: String, attribute: 'card-message' },
-			isValueClickable: { type: Boolean, attribute: 'is-value-clickable' }
+			isValueClickable: { type: Boolean, attribute: 'value-clickable' }
```
johngwilkinson, line 20: just updating this attribute name to match d2l standard

bemailloux: `out of scope` I'm wondering why we don't just make the whole card clickable

bemailloux (approved):

components/summary-card.js
johngwilkinson: Yeah, that's what I did in my prototype, but this is the design. Maybe it's harder for the user to know it's clickable.

bemailloux: I'd argue having big buttons is a good thing for accessibility reasons. Some of the other cards have small hitboxes though, so it doesn't really matter. I think the normal behaviour for someone who has trouble clicking precisely is to make the whole page bigger, which engagement dashboard can handle pretty well.

components/summary-card.js
```diff
@@ -104,18 +108,35 @@ class SummaryCard extends SkeletonMixin(LitElement) {
...
 			<div class="d2l-insights-summary-card-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this.title}</div>
 			<div class="d2l-insights-summary-card-body" aria-hidden="${this.skeleton}">
 				<div class="d2l-skeletize">
-					${this.isValueClickable ?
-						html`<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field" ?is-value-clickable=${this.isValueClickable} @click=${this._valueClickHandler}>${this.value}</span>` :
-						html`<span class="d2l-insights-summary-card-value d2l-insights-summary-card-field">${this.value}</span>`}
+					${this.isValueClickable ? html`<button
+ 						class="d2l-insights-summary-card-button d2l-insights-summary-card-field"
+ 						@click=${this._valueClickHandler}
+ 					>
```
Vitalii-Misechko, line 125: Have you chance to check skeleton view?

Vitalii-Misechko (approved):

components/summary-card.js
johngwilkinson: It looks okay to me. Is there anything in particular I should check for?

johngwilkinson: image

Vitalii-Misechko: Looks good!

Vitalii-Misechko: I see that the second card has a different skeleton (bottom margin is smaller than in the first card). I don't think this was caused by this PR.

PR #123, opened 2020-10-29
Fix Edge Legacy Discussion Summary

NicholasHallman:
fix

Solution: 
It looks like lit elements will make up for the lack of shadow dom in edge by applying the elements name as a css style to scope the styles to the component. However, if the element name overlaps with a class name that class (in this case, the card border) get's applied to all of the elements in the component! fun!


johngwilkinson (approved):

bemailloux (approved):
Ah yes, this issue! That's why there's a class called `d2l-insights-table-table` in table.js XD

PR #125, opened 2020-10-29
Add margin to expander button

bemailloux:
So the text doesn't get too close to the button.

Before:
image

After: 
image

(Yes the language in the rest of the dialog is wrong - I was just testing quickly)

FYI @NicholasHallman @Vitalii-Misechko @MykolaGalian @rohitvinnakota

### DIFF
--- a/components/expander-with-control.js
+++ b/components/expander-with-control.js
@@ -28,6 +28,10 @@ class ExpanderWithControl extends Localizer(LitElement) {
 				margin: 1em 0 0.5em 0;
 			}
 
+			.d2l-insights-expand-collapse-control-button {
+				margin-left: 10px;
+			}
+
 			.d2l-insights-expand-collapse-control-text {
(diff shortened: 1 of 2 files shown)


components/expander-with-control.js
```diff
@@ -28,6 +28,10 @@ class ExpanderWithControl extends Localizer(LitElement) {
 				margin: 1em 0 0.5em 0;
 			}
 
+			.d2l-insights-expand-collapse-control-button {
+				margin-left: 10px;
```
anhill-D2L, line 32: Do we need to add a margin to both sides for RTL?

bemailloux: Yes we do! Great catch

bemailloux: Fixed: 
image

anhill-D2L (approved):

johngwilkinson (approved):

PR #126, opened 2020-10-30
Us121479/email popup

bemailloux:
Opens an email popup with the email addresses of users selected in the table.

Behind the scenes we're making 2 calls to the email API. 
1. Send a list of userIds and get back a list of email addresses. Relies on "See External Email Address for `<role>`" (or "See Internal Email Address for `<role>`" permission.
2. Send the email. Relies on "Send emails to external email addresses" (or "Send emails to internal email addresses") permission.

Users will be expected to have these permissions at the root orgUnit level to be able to view and send emails from the engagement dashboard. (I'll include a writeup in the story notes as to why I chose this approach over alternatives).

image

Quality notes
* Testing
  * [x] Browsers (Windows x Chrome, FF, Edgium, Legacy)
  * [x] Mobile (it opens the email window in a new tab)
  * [x] Fetching email addresses for user ids
    * If user has permission, email address is added to BCC list
    * If user does not have permission, email popup provides a toast message: "The email addresses of <N> users could not be added"
  * [ ] Sending emails
    * Untested. I can't send emails from my dev LMS because I don't have an SMTP server running. Once this is merged I'm planning to test using Draco LMS.
* Security
  * Users are required to have view / send email permissions on the root OU
* Docs, perf, devops: N/A
* UX
  * [x] Demo to Kevin
  * [x] Screen reader: NVDA will automatically start reading the content of the email popup when it appears.
  * [x] Localized, RTL-supported


engagement-dashboard.js
```diff
@@ -295,8 +298,8 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 			const noUsersSelectedDialog = this.shadowRoot.querySelector('#no-users-selected-dialog');
 			noUsersSelectedDialog.opened = true;
 		} else {
-			// (out-of-scope) show email edit dialog
-			console.log(selectedUserIds);
+			// we use the root OU id because that's where we expect users to have email permissions
+			createComposeEmailPopup(toJS(selectedUserIds), this._data._data.orgUnitTree.rootId);
```
rohitvinnakota, line 302: _aside_: Is there anything we can do to change the attribute names as we dive into the tree structure? `this._data._data` is a little confusing at first.

bemailloux: Haha I totally agree :)

Actually, I think I can just use `this._serverData().orgUnitTree`. 

What confused me is that in engagement-dashboard.js there's a `get _serverData` which returns a Data object, and a `get _data` which returns a FilteredData object, and also the Data class has a property called `serverData`. So some renaming is probably a good idea

johngwilkinson: What is definitely a code smell and is only possible because javascript is bad at actual encapsulation, is accessing any member starting with `_` from outside the class.

johngwilkinson (approved):

rohitvinnakota (approved):

PR #127, opened 2020-10-30
Add no results error message

rohitvinnakota:
image

[US117901](https://rally1.rallydev.com/#/detail/userstory/399910079036?fdp=true): [Engagement] Errors - No data in filtered range

Based on earlier discussion during our meetings, I have the cards and charts not rendering when 0 users are returned. ~~It seems that we have some default functionality for the user table when no users are returned so I left that as is.~~ 
Demoed to Kevin and we agreed upon rendering just the message and no user table when there are 0 users returned. 

**Quality notes**
Tested on edge, firefox, chrome
Changing filters removes or recreates dialog as expected
LTR to RTL works as expected
Responsiveness works as expected

Accessibility: Text picked up by screen reader

Perf testing: n/a
Devops: n/a

- [x]  Demo to Kevin

fyi: @Vitalii-Misechko @MykolaGalian @NicholasHallman @hyehayes @anhill-D2L

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -158,33 +158,31 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 						?demo="${this.isDemo}"
 					></d2l-insights-role-filter>
 				</div>
-				<d2l-insights-message-container .data="${this._data}"></d2l-insights-message-container>
-				<span class="${this._isNoUserResults ? 'd2l-insights-noDisplay' : 'd2l-insights-display'}">
-					<h2 class="d2l-heading-3">${this.localize('components.insights-engagement-dashboard.summaryHeading')}</h2>
-					<div class="d2l-insights-summary-container-applied-filters">
-						<d2l-insights-applied-filters .data="${this._data}" ?skeleton="${this._isLoading}"></d2l-insights-applied-filters>
-					</div>
-					<div class="d2l-insights-summary-container">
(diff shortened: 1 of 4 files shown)


engagement-dashboard.js
```diff
@@ -224,6 +229,10 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		return this.__data;
 	}
 
+	get _isNoUserResults() {
+		return this.__data.users.length === 0 && !this._data.isLoading;
```
johngwilkinson, line 233: So, is the desired behaviour
a) show this message when there is no data matching all current filters? or
b) show when there is no data matching the ou, semester, and role filters?

You've implemented (a) here, but you've also hidden the "applied filters" component, so there's no way for the user to remove a card filter! That makes me think you intended (b). Another problem with (a) is that the screen becomes kind of "jumpy": if I click a "0" on a card, all the cards suddenly disappear.

For (b), you want to look at `this._systemData.records.length` I believe (either way: one underscore is better than two: only the getter should ever reference the "hidden" backing variable).

rohitvinnakota: This is meant to be a page level error, (case when all cards have 0 results), so it would probably be a) with the applied filters still visible?

johngwilkinson: This is probably something to walk through with Kevin. Take the discussion filters, for example: they are toggle-able, so you can click, then click again - but if most of the page suddenly vanishes when you click, that experience seems bad.

Does the a.c. require hiding the cards? Maybe it's best to just show the message and leave the cards. Anyway, I think it's Kevin's call.

PR #130, opened 2020-10-30
Fix flaky test for Safari

Vitalii-Misechko:
Fix flaky test for Safari

FYI @rohitvinnakota @MykolaGalian @johngwilkinson @NicholasHallman

### DIFF
--- a/karma.sauce.conf.js
+++ b/karma.sauce.conf.js
@@ -28,12 +28,6 @@ const customLaunchers = {
 module.exports = config => {
 	config.set(
 		merge(createDefaultConfig(config), {
-			// from https://support.saucelabs.com/hc/en-us/articles/225104707-Karma-Tests-Disconnect-Particularly-When-Running-Tests-on-Safari
-			// to avoid DISCONNECTED messages
-			browserDisconnectTimeout: 10000, // default 2000
-			browserDisconnectTolerance: 1, // default 0
-			browserNoActivityTimeout: 4 * 60 * 1000, //default 10000
-			captureTimeout: 4 * 60 * 1000, //default 60000
 			files: [
 				// runs all files ending with .test in the test folder,
 				// can be overwritten by passing a --grep flag. examples:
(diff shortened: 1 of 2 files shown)


karma.sauce.conf.js
```diff
@@ -28,6 +28,12 @@ const customLaunchers = {
 module.exports = config => {
 	config.set(
 		merge(createDefaultConfig(config), {
+			// from https://support.saucelabs.com/hc/en-us/articles/225104707-Karma-Tests-Disconnect-Particularly-When-Running-Tests-on-Safari
+			// to avoid DISCONNECTED messages
+			browserDisconnectTimeout: 10000, // default 2000
+			browserDisconnectTolerance: 1, // default 0
+			browserNoActivityTimeout: 4 * 60 * 1000, //default 10000
+			captureTimeout: 4 * 60 * 1000, //default 60000
```
anhill-D2L, line 36: Thank you for the comments and link 👍

anhill-D2L (approved):

karma.sauce.conf.js
Vitalii-Misechko: Oh.. this shouldn't be here. It does not work AFAIK. I copied it from `karma.conf.js` in hope it will fix our timeouts.

anhill-D2L: Is it to fix a different issue from the flakey timeouts we've been seeing lately?

Vitalii-Misechko: When I started worked on this flaky-test I could not figure out the route cause of `sends timings with a performance event`-test. So, I tried to fix timeouts that you mentioned. Now, when I reverted these lines I see that introducing them had some positive effect. Thus, I'm reverting my revert ))

anhill-D2L (approved):

test/model/telemetry-helper.test.js
```diff
@@ -36,11 +36,19 @@ describe('TelemetryHelper', () => {
...
 
-			console.dir(fetchSandbox.lastCall());
 			const lastCall = fetchSandbox.lastCall();
 
 			expect(lastCall[0]).to.equal('http://example.com/');
-			const event = JSON.parse(await lastCall[1].body);
+
+			const body = await lastCall[1].body;
+
+			if (body === '[object ReadableStream]') {
```
bemailloux, line 45: If this is a `ReadableStream`, can we just read the stream to get the contents?

Vitalii-Misechko: Unfortunately, It's not a `ReadableStream` it's string with value that equals to`[object ReadableStream]` which is returned in body when `await lastCall[1].body`-promise is resolved

PR #132, opened 2020-11-03
US121093 [Engagement] Explain when no courses in default view

Vitalii-Misechko:
[US121093](https://rally1.rallydev.com/#/detail/userstory/439372243588?fdp=true): [Engagement] Explain when no courses in default view

LMS PR: https://github.com/Brightspace/lms/pull/3043

### Functional Testing
* Unit tests
* Manual test
  * Navigate to Engagement Dashboard. Confirm that popup windows appears.
    * Confirm that pop-up has courses list if users is enrolled to courses
    * Confirm that pop-up does not have courses list if user is not enrolled to any course within default semesters
    * Confirm that pop-up window does not appear if any further fetch from LMS occurs
### UX and docs - N/A

- [x] Demo

image


components/default-view-popup.js
```diff
@@ -51,26 +58,13 @@ class DefaultViewPopup extends RtlMixin(Localizer(LitElement)) {
 					width="615"
 					@d2l-dialog-close="${this._closeDialog}">
 
-				<p class="d2l-insights-default-view-popup-description">
-					${this.localize('components.insights-default-view-popup.defaultViewDescription1', { numDefaultCourses: displayData.length })}
-				</p>
+				${this._renderDescription1({ displayData, numDefaultCourses: displayData.length, numDefaultSemesters: this._defaultSemestersCount })}
```
johngwilkinson, line 61: nit: is the `1` at the end of the method name needed?

johngwilkinson (approved):

components/default-view-popup.js
```diff
@@ -79,6 +73,41 @@ class DefaultViewPopup extends RtlMixin(Localizer(LitElement)) {
 		`;
 	}
 
+	_renderDescription1({ displayData, numDefaultCourses, numDefaultSemesters }) {
+		if (displayData.length) {
+			return html`
+				<p class="d2l-insights-default-view-popup-description">
+				${this.localize('components.insights-default-view-popup.defaultViewDescription1', { numDefaultCourses })}
```
bemailloux, line 80: `defaultViewDescription1` and similar names probably should be more descriptive now.

model/data.js
```diff
@@ -137,6 +138,14 @@ export class Data {
 		});
 	}
 
+	get numDefaultSemesters() {
+		return this.serverData.numDefaultSemesters;
+	}
+
+	get isDefaultView() {
+		return this.serverData.isDefaultView;
```
bemailloux, line 146: This won't work as expected: there is a situation where `serverData.isDefaultView` can be true but no orgUnitIds were preselected. In that case we should not show the default view popup.

Please see https://github.com/Brightspace/insights-engagement-dashboard/pull/132/files#diff-eee75df07aac258797ab7469564820082ce3f54d26a380d5998fed4a5bf130d4R122

If you're going to make a getter for `isDefaultView`, it should be
```
get isDefaultView() {
  return this.defaultViewPopupDisplayData.length > 0;
}
```

johngwilkinson: Isn't that the case this story is about?

Vitalii-Misechko: > `serverData.isDefaultView` can be true but no orgUnitIds were preselected. In that case we should not show the default view popup.

I thought this is exactly case when the code should show pop-up. Have a look at [thread here ](https://d2l.slack.com/archives/CMNGJ22G0/p1604066573346400?thread_ts=1604064374.344100&cid=CMNGJ22G0)

@johngwilkinson 
> "If semesters exist" means defaultView is true and the server sends back a value for selectedSemestersIds, but no org units. This will happen if the user does not have permission to view data for any courses in the selected (by the server) semesters.

Sorry, what should the link you provided show?

bemailloux: No, what I'm talking about is a distinct case.

This story: show the popup if no data is returned at all, given the semester preselection

The case I'm talking about: data is returned and it happens to be the same as if you had loaded using the `GeneralDataProvider`

Looks like the way to differentiate now is to check if `serverData.records.length > 0`.

bemailloux: Maybe we don't care that much about the case I'm talking about - it was never part of the formal AC for the original default view popup.

components/default-view-popup.js
Vitalii-Misechko: done

components/default-view-popup.js
Vitalii-Misechko: fixed

model/data.js
Vitalii-Misechko: `GeneralDataProvider` is not a part of default view flow, AFAIK. 
Anyway, let me know if any action from my side is required

components/default-view-popup.js
```diff
@@ -73,17 +73,17 @@ class DefaultViewPopup extends RtlMixin(Localizer(LitElement)) {
 		`;
 	}
 
-	_renderDescription1({ displayData, numDefaultCourses, numDefaultSemesters }) {
+	_renderDescriptiveEmptyResultsMessage({ displayData, numDefaultCourses, numDefaultSemesters }) {
 		if (displayData.length) {
 			return html`
 				<p class="d2l-insights-default-view-popup-description">
-				${this.localize('components.insights-default-view-popup.defaultViewDescription1', { numDefaultCourses })}
+				${this.localize('components.insights-default-view-popup.emptyResultsFromNRecentCourses', { numDefaultCourses })}
```
bemailloux, line 80: `emptyResultsFromNRecentCourses` should just be `resultsFromNRecentCourses`, because this is the message that displays if the results are _not_ empty

bemailloux (approved):

model/data.js
bemailloux: No action required, I think it's ok if we just show the default view in the case I described. It means that all clients will now see the default view popup on load, which might be annoying to users until we decide to implement logic for saving the client-side state.

locales/en.js
```diff
@@ -128,8 +128,9 @@ export default {
...
 
 	"components.insights-default-view-popup.title": "Engagement Dashboard Default View",
-	"components.insights-default-view-popup.defaultViewDescription1": "This dashboard is designed to look at portions of your organization's engagement. Results showing are from {numDefaultCourses} recently accessed courses to get you started.",
-	"components.insights-default-view-popup.defaultViewDescription2": "Use the dashboard filters to change the results displayed.",
+	"components.insights-default-view-popup.emptyResultsFromNRecentCourses": "This dashboard is designed to look at portions of your organization's engagement. Results showing are from {numDefaultCourses} recently accessed courses to get you started.",
+	"components.insights-default-view-popup.promptUseFilters": "Use the dashboard filters to change the results displayed.",
+	"components.insights-default-view-popup.emptyResultsFromNRecentSemesters": "This dashboard is designed to look at portions of your organization's engagement. You do not have permission to review data in any courses in the most recently created {numDefaultSemesters} semesters",
```
bemailloux, line 133: There is a period `.` missing at the end of the sentence

components/default-view-popup.js
Vitalii-Misechko: fixed

locales/en.js
Vitalii-Misechko: fixed

Vitalii-Misechko:
During manual test found a defect introduced by this PR. 
image

Now it's fixed

components/default-view-popup.js
```diff
@@ -37,6 +37,10 @@ class DefaultViewPopup extends RtlMixin(Localizer(LitElement)) {
 			numDefaultSemesters: 0
 		};
 		this.opened = false;
+		// when the pop-up is rendered the first time it is hidden and it can't properly calculate the height
+		// of the pop-up due to data is not received yet and not all components placed into the content.
+		// In order to calculate pop-up's height right, we should render all possible components for the first time
+		this._firstRender = true;
```
bemailloux, line 43: `Out of scope` I wonder if this can also be used to fix the resizing when the courses list is expanded

test/components/default-view-popup.test.js
```diff
@@ -53,7 +53,7 @@ describe('d2l-insights-default-view-popup', () => {
 		});
 	});
 
-	describe('when there are no selected courses', () => {
+	describe.skip('when there are no selected courses', () => {
```
bemailloux, line 56: Why is this skipped?

components/default-view-popup.js
Vitalii-Misechko: It worth trying. However, I'm not sure if it'll work

test/components/default-view-popup.test.js
Vitalii-Misechko: Thank for noticing it! I thought I removed it. 
This `skip` is used to check if tests on master work.

bemailloux (approved):

PR #135, opened 2020-11-05
DE41152 Fix cards skeleton

Vitalii-Misechko:
[DE41152](https://rally1.rallydev.com/#/detail/defect/450306771184?fdp=true): [Engage] Discussion Activity shows spinner shifted to the left edge of the card

This PR introduces `d2l-insights-card-overlay`. This way we can show consistent skeletons for different card types

### Functional Testing
* Unit tests pass
* Verify look of the dashboard when it's loading
* Verify look of the dashboard when it's finished loading
### UX and docs
* Responsive UI works
* Accessible (screen reader / keyboard nav)
* Browsers
  * Chrome
  * FF
  * Edgium
  * ~Edge Legacy N/A~

Before
image

After
image

FYI @anhill-D2L @MykolaGalian @rohitvinnakota @hyehayes @johngwilkinson

### DIFF
--- a/components/card-overlay.js
+++ b/components/card-overlay.js
@@ -27,7 +27,7 @@ class CardOverlay extends SkeletonMixin(LitElement) {
 				display: none;
 			}
 
-			.d2l-insights-card-overlay {
+			.d2l-insights-card-overlay-container {
 				background-color: white;
 				border-radius: 15px;
 				display: flex;
@@ -76,7 +76,7 @@ class CardOverlay extends SkeletonMixin(LitElement) {
 		}
 
 		return html`
-			<div class="d2l-insights-card-overlay" aria-hidden="true">
+			<div class="d2l-insights-card-overlay-container" aria-hidden="true">
 				<div class="d2l-insights-card-overlay-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">A title. Do not localize.</div>
 				<div class="d2l-insights-card-overlay-body">
 					<div class="d2l-skeletize"></div>
--- a/components/card-overlay.js
+++ b/components/card-overlay.js
@@ -32,10 +32,10 @@ class CardOverlay extends SkeletonMixin(LitElement) {
 				border-radius: 15px;
 				display: flex;
 				flex-direction: column;
-				height: calc(100% - 10px);
+				height: calc(100% - 20px);
 				justify-content: center;
 				padding-left: 10px;
-				padding-top: 10px;
+				padding-top: 20px;
 				width: calc(100% - 15px);
 			}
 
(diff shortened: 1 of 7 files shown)


components/discussion-activity-card.js
```diff
@@ -153,15 +153,15 @@ class DiscussionActivityCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 			<div class="d2l-insights-discussion-activity-card-body">
 				<d2l-labs-chart
-					class="d2l-insights-discussion-activity-card-body"
 					.options="${this.chartOptions}"
 					.globalOptions="${this.globalHighchartsOptions}"
-					?skeleton="${this.skeleton}">
+				>
 				</d2l-labs-chart>
 			</div>
+			<d2l-insights-card-overlay ?skeleton="${this.skeleton}"></d2l-insights-card-overlay>
```
bemailloux, line 164: `Wondering` How do you make sure the overlay always renders on top of the card?

bemailloux (commented):
This looks really nice!

Have you checked:
- different browsers?
- different page sizes?

components/discussion-activity-card.js
Vitalii-Misechko: In short, a div that is closer to the end of the page will overlay all the div's that are higher in a page's markup. Here is a [picture from MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index). 

image.

Of course, applying z-index can ruin that. However, I've been told that using z-index is a code smell.

Vitalii-Misechko:
> This looks really nice!
> 
> Have you checked:
> 
> * different browsers?
> * different page sizes?

Thanks! Added a few notes into _UX and docs_ section of _Quality Notes_

Vitalii-Misechko:
Minor improvement. Decreased left padding
Before
image

After

image

components/discussion-activity-card.js
bemailloux: I had no idea about this, but of course it makes sense because elements lower down get painted after (so "on top"). Thanks for sharing!

components/card-overlay.js
```diff
@@ -0,0 +1,89 @@
...
+			}
+			:host([skeleton]) {
+				height: 100%;
+				width: 100%;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+
+			.d2l-insights-card-overlay {
```
bemailloux, line 30: I just remembered: having a CSS class that is the same as the element name causes problems on Edge Legacy. (@NicholasHallman ran into this problem as well)

image

Solution: change the class name to something else. (I've checked that this works)

Vitalii-Misechko: Thank you a lot!

bemailloux: All good now 👍 

image

bemailloux (approved):

NicholasHallman (approved):

Vitalii-Misechko:
I've updated some padding after demo to Kevin

image

PR #140, opened 2020-11-13
US122025 Check if a user can send emails for selected students via LMS call

Vitalii-Misechko:
[US122025](https://rally1.rallydev.com/#/detail/userstory/450314090320?fdp=true): [Engagement] Email permission error message

### Functional Testing
* Manual tests
  * Confirm that click on Email fires a call to `/d2l/api/ap/unstable/insights/data/canSendEmail`
  * Confirm that if response code other than `200` dialog with following message appear `Additional permissions are required to be able to send emails to these users.  Please contact your system administrator for access`
### UX and docs - N/A

image

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -17,6 +17,8 @@ import './components/discussion-activity-card.js';
 import './components/message-container.js';
 import './components/default-view-popup.js';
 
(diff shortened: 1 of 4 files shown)


Vitalii-Misechko:
We don't need client-side message. The compose dialog should show this message

PR #141, opened 2020-11-16
US121459: fix broken osx test

johngwilkinson:
The test used to create an event-handler after a setTimeout call which attempted to make sure the click happened after the event-handler was attached. This looks like a possible race condition, and straightening it out by creating the promise before clicking (then awaiting it after) did fix the issue (see earlier commit). But there's also a utility, `oneEvent` we use elsewhere for this purpose, so I updated the code to use that.

Quality Notes: test change only

### DIFF
--- a/test/components/expander-with-control.test.js
+++ b/test/components/expander-with-control.test.js
@@ -73,7 +73,7 @@ describe('d2l-insights-expander-with-control', () => {
 
 		// This test fails on Chrome/OSX and it seems to sometimes crash Safari as well. I don't understand why, but
 		// collapsing and expanding does work on a real LMS
-		it.skip('should fire collapsed event if element is expanded and control is clicked', async function() {
+		it('should fire collapsed event if element is expanded and control is clicked', async function() {
 			this.timeout(3000);
 
 			const controlDiv = elExpanded.shadowRoot.querySelector('div');
--- a/test/components/expander-with-control.test.js
+++ b/test/components/expander-with-control.test.js
@@ -1,6 +1,6 @@
 import '../../components/expander-with-control';
 
-import { expect, fixture, html } from '@open-wc/testing';
+import { expect, fixture, html, oneEvent } from '@open-wc/testing';
 import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper';
 
 describe('d2l-insights-expander-with-control', () => {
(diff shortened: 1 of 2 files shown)


johngwilkinson:
Oh sure, it passes twice in a row then fails when I update the README. 😂

test/components/expander-with-control.test.js
```diff
@@ -73,24 +73,28 @@ describe('d2l-insights-expander-with-control', () => {
 
 		// This test fails on Chrome/OSX and it seems to sometimes crash Safari as well. I don't understand why, but
 		// collapsing and expanding does work on a real LMS
-		it.skip('should fire collapsed event if element is expanded and control is clicked', async function() {
+		it('should fire collapsed event if element is expanded and control is clicked', async function() {
 			this.timeout(3000);
 
+			const listener = oneEvent(elExpanded, 'd2l-insights-expander-with-control-collapsed');
```
bemailloux, line 79: Interesting: the whole reason I moved to the "alternate" way of creating an event handler was specifically _because_ oneEvent wasn't working for me.

test/components/expander-with-control.test.js
```diff
@@ -73,24 +73,28 @@ describe('d2l-insights-expander-with-control', () => {
 
 		// This test fails on Chrome/OSX and it seems to sometimes crash Safari as well. I don't understand why, but
 		// collapsing and expanding does work on a real LMS
```
bemailloux, line 75: If the flakiness has been fixed then this comment isn't relevant anymore

bemailloux (approved):

NicholasHallman (approved):

test/components/expander-with-control.test.js
johngwilkinson: Seems to be a bit flaky now. Maybe I'll try the promise-click-await pattern again to see if that helps. To clarify: it failed 100% of the time before, right?

bemailloux: Yes, it was failing 100% of the time on Chrome/OSX with both oneEvent and the alternative

johngwilkinson:
Sorry for the noise. It's failing about 50% of the time with the oneEvent version, and more than that with the promises (based on a small sample, so...). It's possibel the 100% failure rate originally seen was a fluke and this change hasn't accomplished anything.

Trying some other things...

johngwilkinson:
Well, the answer is pretty boring: it just takes a long time sometimes, so I increased the timeout.

(The d2l component we're using waits for the next animation frame; this can take more than 10s in the test environment - found by adding logging to a copy of the component and using that to run Sauce ci on another branch)

bemailloux:
> The d2l component we're using waits for the next animation frame; this can take more than 10s in the test environment

This is surprising. Any ideas what it is about the test env that causes this?

johngwilkinson:
> > The d2l component we're using waits for the next animation frame; this can take more than 10s in the test environment
> 
> This is surprising. Any ideas what it is about the test env that causes this?

Nothing educated. But maybe the video recording interacts with the browsers notion of animation frames.

PR #144, opened 2020-11-18
US121773 [UserDrill] Create the page

Vitalii-Misechko:
[US121773](https://rally1.rallydev.com/#/detail/userstory/449082079828?fdp=true): [Engagement][UserDrill] Create the page

Minimal working solution that would allow other developers to work on UserDrill in demo page.

What is in-progress (separate PR/PRs):
 * ImmersiveNav as a part of engagement dashboard
 * Navigation to the new page by clicking on user name
 * Routing based on `page.js` that supports browser's Back button
 * Page design refining

### Functional Testing
* Unit tests pass
* Manual tests
  * The new page is available by default in demo page.
  * `d2l-insights-engagement-dashboard` behavior is not changed when it's invoked from LMS
### UX and docs - N/A

image

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -16,6 +16,7 @@ import './components/course-last-access-card.js';
 import './components/discussion-activity-card.js';
 import './components/message-container.js';
 import './components/default-view-popup.js';
+import './components/user-drill-view.js';
 
 import { css, html } from 'lit-element/lit-element.js';
 import { getPerformanceLoadPageMeasures, TelemetryHelper } from './model/telemetry-helper';
@@ -38,6 +39,7 @@ import { toJS } from 'mobx';
 
 /**
  * @property {Boolean} isDemo - if true, use canned data; otherwise call the LMS
+ * @property {String} currentView - is the name of supported view. Valid values: home, user, settings
  * @property {String} telemetryEndpoint - endpoint for gathering telemetry performance data
  * @property {String} telemetryId - GUID that is used to group performance metrics for each separate page load
  */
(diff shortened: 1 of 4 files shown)


engagement-dashboard.js
```diff
@@ -46,11 +48,17 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
...
 			isDemo: { type: Boolean, attribute: 'demo' },
+			currentView: { type: String, attribute: 'view', reflect: true },
 			telemetryEndpoint: { type: String, attribute: 'telemetry-endpoint' },
 			telemetryId: { type: String, attribute: 'telemetry-id' },
 		};
 	}
 
+	constructor() {
+		super();
+		this.currentView = 'home';
```
bemailloux, line 59: `out of scope` there are a bunch of initializations in each of the getter functions (e.g. [here](https://github.com/Brightspace/insights-engagement-dashboard/pull/144/files#diff-da16422f38b47f7e8f1ecb041f786890b12adf07da4b62f2ab5a7d539b44a9e3R285)) that could go in the constructor. Not sure if we want to do that, but I personally find it cleaner to initialize in the constructor

locales/en.js
```diff
@@ -135,5 +135,9 @@ export default {
 	"components.insights-default-view-popup.emptyResultsFromNRecentSemesters": "This dashboard is designed to look at portions of your organization's engagement. You do not have permission to review data in any courses in the most recently created {numDefaultSemesters} semesters.",
 	"components.insights-default-view-popup.expandDefaultCourseList": "Expand to see the courses included in your default view",
 	"components.insights-default-view-popup.collapseDefaultCourseList": "Collapse the list of courses included in your default view",
-	"components.insights-default-view-popup.buttonOk": "Ok"
+	"components.insights-default-view-popup.buttonOk": "Ok",
+
+	"components.insights-user-drill-view.exportToCsv": "Export to CSV",
+	"components.insights-user-drill-view.print": "Print",
```
bemailloux, line 141: Maybe controls on the user page should use the same lang terms as the main page? That way
- there can't be any inconsistencies between the main page and the user page
- we don't have to pay for extra translations (I think we pay for each term)

Worth asking Kevin, I guess

bemailloux (approved):

johngwilkinson (approved):

locales/en.js
Vitalii-Misechko: I'm fine with using the same langterm in distinct pages. If some page has different wording we will create a new term. I'll updated terms in a new PR. I think these are technical details and they would not be interesting to Kevin.

PR #145, opened 2020-11-18
[WIP] US120701 Settings page - initial

bemailloux:
TODO: responsive footer buttons

image


bemailloux:
Making the footer look like it's "floating" depends on the position of the header, so it'll be easier if I move the header inside the engagement dashboard first. FYI @Vitalii-Misechko

Vitalii-Misechko:
> Making the footer look like it's "floating" depends on the position of the header, so it'll be easier if I move the header inside the engagement dashboard first. FYI @Vitalii-Misechko

No objection from my site.

PR #148, opened 2020-11-19
Us120701/move immersive nav

bemailloux:
Moves immersive nav component from the MVC component into the engagement dashboard itself.

Corresponding LMS changes: https://github.com/Brightspace/lms/pull/3671

**Quality notes**
* Testing
  * [x] Responsive
  * [x] Accessibility (keyboard nav / screen reader, NVDA + Chrome)
  * [x] Test cases (regressions)
    * It links back to the Insights Portal with the correct OU
      * Note: it always links back to insights portal with `ou=<rootOUId>`. This isn't a regression
    * On large screens it displays the full text: "Back to Insights Portal"
    * On small screens it displays short text: "Back"
    * Opt in/out fly-out component is still visible
* Security, perf, devops: N/A
* UX
  * Not demoing to Kevin - from a UX standpoint nothing has changed

**Screenshots**
\> 615px:
image

< 615px:
image

FYI @NicholasHallman @johngwilkinson @MykolaGalian @anhill-D2L

### DIFF
--- a/components/immersive-nav.js
+++ b/components/immersive-nav.js
@@ -1,0 +1,77 @@
+import 'd2l-navigation/d2l-navigation-immersive';
+import 'd2l-navigation/d2l-navigation-link-back';
+
+import { css, html, LitElement } from 'lit-element/lit-element';
+
+// Extends the standard immersive nav to resize the back link on small screens
+class InsightsImmersiveNav extends LitElement {
+	static get properties() {
+		return {
+			href: { type: String, attribute: true },
+			mainText: { type: String, attribute: 'main-text' },
+			backText: { type: String, attribute: 'back-text' },
+			backTextShort: { type: String, attribute: 'back-text-short' } // optional - will default to backText if unspecified
+		};
+	}
+
+	static get styles() {
+		return [css`
+			.d2l-insights-immersive-nav-title {
+				align-items: center;
+				display: flex;
+			}
+
+			.d2l-insights-link-back-default {
+				display: inline-block;
+			}
+
+			.d2l-insights-link-back-responsive {
+				display: none;
+			}
+
+			@media screen and (max-width: 615px) {
+				.d2l-insights-link-back-default {
+					display: none;
+				}
+
+				.d2l-insights-link-back-responsive {
+					display: inline-block;
+				}
+			}
+		`];
+	}
+
+	constructor() {
+		super();
+		this.href = '';
+		this.mainText = '';
+		this.backText = '';
+		this.backTextShort = '';
+	}
+
+	render() {
+		return html`
+			<d2l-navigation-immersive width-type="fullscreen">
+
+				<div slot="left">
+					<d2l-navigation-link-back
+						text="${this.backText}"
+						href="${this.href}"
+						class="d2l-insights-link-back-default">
+					</d2l-navigation-link-back>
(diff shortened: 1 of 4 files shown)


bemailloux:
@Vitalii-Misechko here's what the user drill view looks like with the immersive nav:

image

bemailloux:
FYI: the immersive nav webcomponent does not support RTL and there isn't a way for us to fix it without fixing the component itself, which is still using old Polymer. This isn't a regression from the current state though (image below).

image

components/immersive-nav.js
```diff
@@ -0,0 +1,77 @@
...
+		this.backText = '';
+		this.backTextShort = '';
+	}
+
+	render() {
+		return html`
+			<d2l-navigation-immersive width-type="fullscreen">
+
+				<div slot="left">
+					<d2l-navigation-link-back
```
bemailloux, line 57: Putting this in a link may cause the page to reload when going from the user drill page to the main dashboard page. I'm not sure how the routing with page.js will work - would it cause the page to not fully reload?

Vitalii-Misechko: The idea of using client-side routing is to prevent page reloading and preserve client app's state between views. 
Client-side routing can handle browser `Back` button without reloading. I think it should work. Thanks for calling it out

Vitalii-Misechko:
Looks great!

Vitalii-Misechko (approved):
Loo

Vitalii-Misechko:
Have you chance to check if the immersive nav behaves good when the page is scrolled?  I expect the immersive nav stays on top and content is behind the nav.

components/immersive-nav.js
bemailloux: Ok, I think we can switch from using `d2l-navigation-back-link` to a regular button or something if we need to avoid page reloads.

Vitalii-Misechko: That's good. I've just figured out that we eventually need this. probably all `back` navigation should happen with `window.history.back()` instead of url in href

bemailloux: Cool I can change it in this PR

rohitvinnakota (approved):

bemailloux:
> Have you chance to check if the immersive nav behaves good when the page is scrolled? I expect the immersive nav stays on top and content is behind the nav.

Thanks for asking this, there is a small problem but otherwise it loads correctly:

image

components/immersive-nav.js
johngwilkinson: Except that if there's no history, there should be a default. I guess something like
if(window.history is empty) window.history.pushState(the dashboard url);
window.history.back()

bemailloux:
> > Have you chance to check if the immersive nav behaves good when the page is scrolled? I expect the immersive nav stays on top and content is behind the nav.
> 
> Thanks for asking this, there is a small problem but otherwise it loads correctly:
> 
> image

It looks like I can't solve this problem from the engagement dashboard - the underlying components would have to be modified. I have opened an issue on d2l-table: https://github.com/BrightspaceUI/table/issues/219

components/immersive-nav.js
bemailloux: I'm being interrupted to work on something else so I will merge this PR as-is. I'll come back to this issue later, or someone else can pick it up.

PR #151, opened 2020-11-20
US121773 [UserDrill] Create the page. Routing via UrlState

Vitalii-Misechko:
[US121773](https://rally1.rallydev.com/#/detail/userstory/449082079828?fdp=true): [Engagement][UserDrill] Create the page

FYI @anhill-D2L @MykolaGalian @rohitvinnakota @hyehayes @bemailloux 

### Functional Testing
* Unit tests
* Manual tests
  * Clicking on user name navigates user to the user view as [per design](https://xd.adobe.com/view/2b998c7a-800e-472b-b41d-d2e14658bff7-7eca/)
  * On dashboard user page, browser's `Back` button navigates back to the default/home view
  * On dashboard user page, clicking on `Back to Engagement Dashboard` navigates to the home view
    * Even if a user navigates to the user page using browser's bookmark or manually typed url
  * The default view popup appears exactly once. It does not appear the second time when a user returns back from the user view
  * On dashboard home page, back button goes to Insights Portal
  * Dashboard page stores its view in URL query parameter. For instance `/ap/visualizations/dashboards/engagement?v=home,0`

### UX and docs
- [x] Browsers (Chrome, FF, Edgium, ~Edgacy~)
- [ ] Responsive
- [ ] Accessibility (keyboard nav / screen reading: NVDA + Chrome)
      * ability to navigate link with keyboard
- [ ] Demo

- [x] Unit tests
- [x] Manual tests
~- [ ] Add filters to user view~

image


NicholasHallman:
I think we should change urlState so we can chose where we want to store the data in the url. For this case it would be nice to store the current view state in the path or href, instead of in the search parameters.

so we get something like   
localhost/d2l/ap/visualizations/dashboards/engagement/<viewState>?<filterState>

johngwilkinson:
> I think we should change urlState so we can chose where we want to store the data in the url. For this case it would be nice to store the current view state in the path or href, instead of in the search parameters.
> 
> so we get something like
> localhost/d2l/ap/visualizations/dashboards/engagement/?

I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.

Vitalii-Misechko:
> > I think we should change urlState so we can chose where we want to store the data in the url. For this case it would be nice to store the current view state in the path or href, instead of in the search parameters.
> > so we get something like
> > localhost/d2l/ap/visualizations/dashboards/engagement/?
> 
> I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.

Yeah... this way we will need url-rewrite configuration in IIS config.

components/table.js
```diff
@@ -324,9 +329,12 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 				</td>
 			`;
 		} else if (columnType === COLUMN_TYPES.TEXT_SUB_TEXT) {
```
johngwilkinson, line 331: It may not matter for this dashboard, but it looks like this will only make a cell clickable if it is of text-subtext type, even though any cell type can be styled as clickable (above). It seems like it should be possible to make any cell clickable just by setting the clickable attribute - I wonder if a little refactor would make that possible.

engagement-dashboard.js
```diff
@@ -158,18 +160,30 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 	}
 
 	render() {
-		switch (this.currentView) {
+		// if a user entered other than home view in url should we show home's view loading page?
```
johngwilkinson, line 163: No, I think it should be possible to bookmark a user's page or the settings page. (I guess that means the user page will need to handle loading conditions)

engagement-dashboard.js
```diff
@@ -394,6 +409,15 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		});
 	}
 
+	_userTableCellClicked(event) {
+		const nameCellIdx = 1;
+		if (event.detail.columnIdx === nameCellIdx) {
+			this._userId = event.detail.userId;
+			this.currentView = 'user';
```
johngwilkinson, line 416: Not sure if it's better or worse, but if `currentView` were an observable, perhaps in `Data`, the `save()` would not be needed here.

engagement-dashboard.js
```diff
@@ -394,6 +409,15 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		});
 	}
 
+	_userTableCellClicked(event) {
+		const nameCellIdx = 1;
+		if (event.detail.columnIdx === nameCellIdx) {
+			this._userId = event.detail.userId;
```
johngwilkinson, line 415: `this._userId` affects rendering, so should it be a lit property or a mobx observable?

NicholasHallman:
>>I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.

> Yeah... this way we will need url-rewrite configuration in IIS config.

I don't think it would require any server side code since pushing and popping the history doesn't send a get request to the server, it would only be observed locally. I agree there's not a huge need and the only reason to do this would be to keep with conventions.

Vitalii-Misechko:
> > > I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.
> 
> > Yeah... this way we will need url-rewrite configuration in IIS config.
> 
> I don't think it would require any server side code since pushing and popping the history doesn't send a get request to the server, it would only be observed locally. I agree there's not a huge need and the only reason to do this would be to keep with conventions.

When I was working with `page.js` to implement routing with url segments like `http://localhost/demo/user/100`. I discovered that  LMS responds with 404 then I found [this article](https://www.mattburkedev.com/iis-rewrite-rule-for-single-page-apps/)

>  since pushing and popping the history doesn't send a get request to the server

that's right until user tries to post url into browser address bar

engagement-dashboard.js
Vitalii-Misechko: I would vote for observable. `userId` does not belong to `home`-view. `engagement-dashboard.js` is a mix of hosting master page, application state and `home`-view. I think we should move in the direction where these pieces are separated.

engagement-dashboard.js
Vitalii-Misechko: Agree. I think this is outside the scope of this story. I'll create one

components/table.js
```diff
@@ -286,21 +287,25 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
+		const clickable = this.columnInfo[idx].clickable;
 		const styles = {
 			'd2l-insights-table-cell': true,
 			'd2l-insights-table-cell-first': idx === 0,
-			'd2l-insights-table-cell-last': idx === this._numColumns - 1
+			'd2l-insights-table-cell-last': idx === this._numColumns - 1,
+			'd2l-insights-table-cell-clickable': clickable
 		};
 
+		const clickHandler = this._clickHandler.bind(this, rowIdx, idx);
```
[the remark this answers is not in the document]
Vitalii-Misechko: done

components/table.js
Vitalii-Misechko: I've added support of clickable to `NORMAL_TEXT`. I've skipped `SUB_COLUMNS`-column type. I think this type of column require some more work.  I guess it is better to postpone of adding clickable for this column till we need to. However, let me know if you think otherwise

engagement-dashboard.js
Vitalii-Misechko: done

engagement-dashboard.js
Vitalii-Misechko: done!

engagement-dashboard.js
Vitalii-Misechko: I've created a story [US122721](https://rally1.rallydev.com/#/detail/userstory/459367003504?fdp=true): [Engagement][UserDrill] Skeleton page while loading

Vitalii-Misechko:
The PR is not really completed yet. There are a few accessibility and responsive issues that I'm working on.
This PR is dependent on https://github.com/Brightspace/insights-engagement-dashboard/pull/147. Therefore, it is hard to review this PR while the dependency is not in sync with master

Vitalii-Misechko:
Hooray...I've updated this PR so it is readable and can be reviewed. It's quite big though. Still working on accessibility and responsive design

components/immersive-nav.js
```diff
@@ -73,5 +94,16 @@ class InsightsImmersiveNav extends LitElement {
 			</d2l-navigation-immersive>
 		`;
 	}
+
+	_backLinkClickHandler(e) {
+		if (this.view === 'home') {
+			return true;
+		}
+
+		window.history.back();
```
johngwilkinson, line 103: Scenario: user goes directly to a user view url, e.g. by bookmark. There's no history. Does this link work?

johngwilkinson: Scenario 2, just to complicate things: same, but it was a link in another web page. So there is browser history, but not from this page.

Vitalii-Misechko: In both cases it will be equal to browser's `Back` button. But we should navigate to home view. I'll fix it. Thank you!

NicholasHallman (approved):

components/immersive-nav.js
Vitalii-Misechko: > Scenario: user goes directly to a user view url, e.g. by bookmark. There's no history. Does this link work?
> Scenario: user goes directly to a user view url, e.g. by bookmark. There's no history. Does this link work?

Fixed. Updated manual tests

johngwilkinson (approved):

Vitalii-Misechko:
@johngwilkinson @NicholasHallman 
Are there any concerns if I merge this PR and continue working on accessibility and responsive UI issues in a separate PR?

bemailloux:
This is a bit late but it looks like the back link text is always `Back to Insights Portal` if you load the home page first then go to the user drill page:

image

I think it's because navigating to the user drill page doesn't cause a re-render of the immersive nav

Vitalii-Misechko:
> This is a bit late but it looks like the back link text is always `Back to Insights Portal` if you load the home page first then go to the user drill page:

Thanks! I'll take a look

NicholasHallman:
> > > I think we should change urlState so we can chose where we want to store the data in the url. For this case it would be nice to store the current view state in the path or href, instead of in the search parameters.
> > > so we get something like
> > > localhost/d2l/ap/visualizations/dashboards/engagement/?
> > 
> > 
> > I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.
> 
> Yeah... this way we will need url-rewrite configuration in IIS config.

Doing some research into the routing framework I think this assumptions is incorrect. [According to these docs](https://docs.dev.d2l/index.php/Routing_Framework) the routing is decoupled from IIS's file serving. I experimented with this idea by adding a `/*` to the dashboard pattern and can add anything to the url after engagement/ and it works great. This would let us add something like page.js for managing the views like in a real SPA with the benefit of human readable routes AND not needing to mess around in IIS.

Vitalii-Misechko:
> > > > I think we should change urlState so we can chose where we want to store the data in the url. For this case it would be nice to store the current view state in the path or href, instead of in the search parameters.
> > > > so we get something like
> > > > localhost/d2l/ap/visualizations/dashboards/engagement/?
> > > 
> > > 
> > > I believe putting it in the href will hit a different LMS route. We could add the extra route(s) on the LMS side, but I'm not sure it's worth extra code.
> > 
> > 
> > Yeah... this way we will need url-rewrite configuration in IIS config.
> 
> Doing some research into the routing framework I think this assumptions is incorrect. [According to these docs](https://docs.dev.d2l/index.php/Routing_Framework) the routing is decoupled from IIS's file serving. I experimented with this idea by adding a `/*` to the dashboard pattern and can add anything to the url after engagement/ and it works great. This would let us add something like page.js for managing the views like in a real SPA with the benefit of human readable routes AND not needing to mess around in IIS.
> 
> Good dragon slayer material perhaps

Cool finding! 👍  Thanks for sharing.

PR #153, opened 2020-11-23
US120705: show cards according to user prefs

johngwilkinson:
Quality Notes:

* functional: demo & e2e (show and hide cards using prefs API)
* security: n/a
* perf: n/a
* ux/docs: n/a (layout is wonky, but there's another story for that)
* devops: requires latest LMS change or all the cards will be hidden

Still to come: user table columns, last access threshold, and roles

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -4,6 +4,8 @@
 		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
 		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 		<script type="module">
+			/* eslint-disable no-console */
+
 			import '@brightspace-ui/core/components/demo/demo-page.js';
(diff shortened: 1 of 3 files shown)


NicholasHallman:
Does this impact filtering in any way? 

For example, if I have a url with the current grade filter set but the current grade card hidden, will the filter still apply?

johngwilkinson:
> Does this impact filtering in any way?
> 
> For example, if I have a url with the current grade filter set but the current grade card hidden, will the filter still apply?

I don't have the url stuff in this branch to test that, but I believe this does not affect filtering: the filters are constructed the same way, and this this just determines whether they are shown.

Something we might want to think about further: if I share a link with you, and you have different preferences set, you could see something very different from me. Especially due to role filters and last access threshold. Maybe we need to include those values in the url, even if they aren't affected by regular interactions.

bemailloux (approved):

NicholasHallman (approved):

PR #154, opened 2020-11-24
US120701 Immersive nav back doesn't reload page

bemailloux:
Modify the immersive nav back link to be a button on dashboard subpages, so it doesn't fully reload the page when clicked.

Quality notes
* Testing
  * [x] Responsive
  * [x] Accessibility (keyboard nav / accessibility: tested with NVDA/Chrome)
  * [x] Test cases
    * (Default size + responsive) On dashboard home page, back button still goes to Insights Portal
    * (Default size + responsive) On dashboard user drill page, back button goes to home page
    * Default view popup is only displayed once
* Security, perf, devops: N/A
* UX: not demoing to Kevin as this is not really something "new"

@johngwilkinson @rohitvinnakota @anhill-D2L @MykolaGalian

### DIFF
--- a/components/immersive-nav.js
+++ b/components/immersive-nav.js
@@ -1,12 +1,20 @@
 import 'd2l-navigation/d2l-navigation-immersive';
 import 'd2l-navigation/d2l-navigation-link-back';
+import 'd2l-navigation/d2l-navigation-button';
+import '@brightspace-ui/core/components/icons/icon';
 
 import { css, html, LitElement } from 'lit-element/lit-element';
 
-// Extends the standard immersive nav to resize the back link on small screens
+// Extends the standard immersive nav to
+// - resize the back button on small screens
+// - allow use of a button instead of a link
 class InsightsImmersiveNav extends LitElement {
 	static get properties() {
 		return {
+			/**
+			 * type: either 'button' or 'link'
+			 */
+			backButtonType: { type: String, attribute: 'back-button-type' },
 			href: { type: String, attribute: true },
 			mainText: { type: String, attribute: 'main-text' },
 			backText: { type: String, attribute: 'back-text' },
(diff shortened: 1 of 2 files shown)


NicholasHallman (approved):
LGTM! my only thought is that instead of rendering two different elements ( a link or a button ) we only render the link and move the on click for the button to the link. If we detect that it's a "button" type we prevent default and push the new href into the url history. I'm not sure if this is better or just different but it would be one fewer functions?

Vitalii-Misechko:
@bemailloux 
Whoops :) I thought that the nav in master is final solution and made some changes in my PR https://github.com/Brightspace/insights-engagement-dashboard/pull/151

Commits:
https://github.com/Brightspace/insights-engagement-dashboard/pull/151/commits/2712f7c3bcc1e1e066e2ceed964dcbc6dddb872a
https://github.com/Brightspace/insights-engagement-dashboard/pull/151/commits/5a0821def2b1d771f05c043a9fc40f0cb6704510

bemailloux:
> @bemailloux
> Whoops :) I thought that the nav in master is final solution and made some changes in my PR #151
> 
> Commits:
> [2712f7c](https://github.com/Brightspace/insights-engagement-dashboard/commit/2712f7c3bcc1e1e066e2ceed964dcbc6dddb872a)
> [5a0821d](https://github.com/Brightspace/insights-engagement-dashboard/commit/5a0821def2b1d771f05c043a9fc40f0cb6704510)
> 
> EDIT: If you don't mind I would preserve mine version.

Sounds good to me - yours uses UrlState which mine does not, so yours is probably better. Although, can you please include the code from this PR which prevents the default view popup from displaying more than once?

bemailloux:
> LGTM! my only thought is that instead of rendering two different elements ( a link or a button ) we only render the link and move the on click for the button to the link. If we detect that it's a "button" type we prevent default and push the new href into the url history. I'm not sure if this is better or just different but it would be one fewer functions?

I'll be closing this in favour of Vitalii's solution (#151) which does what you suggest here.

Vitalii-Misechko:
> Although, can you please include the code from this PR which prevents the default view popup from displaying more than once?

Sure

PR #156, opened 2020-11-25
Fix: Popping from history doesn't break mobx state tracking

NicholasHallman:
A brief history...

### Problem 1

The urlstate created a loop so when we saved the the persistenceValue it would call autorun to save the new state. Mobx is smart enough to know that the outcome will be the same as the url that was just set. However, when we set the persistent state for category filters (like the course access chart) the state was set by clearing all the categories and then turing on all the ones received from the url. autorun would get called for each of the individual category changes, which polluted the history and prevented the user from using the forward button in the browser.

My initial solution was simple, if we are popping the state, don't allow urlstate to save the state.

### Problem 2

When mobx called the autorun it would change the dependent observable to the popping boolean. Because this variable wasn't an observable mobx would simply stop tracking the persistentValue... darn.

The second sultion, was to create a new function in the category filters class that would clear the categories and set the categories given but all wrapped in an actio. This way mobx only called autorun to save the urlstate AFTER all the intermediate states had collapsed. 

TLDR mobx actions are useful

### DIFF
--- a/components/course-last-access-card.js
+++ b/components/course-last-access-card.js
@@ -66,8 +66,7 @@ export class CourseLastAccessFilter extends CategoryFilter {
 			return;
 		}
 		const categories = value.split(',').map(category => Number(category));
-		this.selectedCategories.clear();
-		categories.forEach(category => this.selectCategory(category));
+		this.setCategories(categories);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


johngwilkinson (approved):
`action` wins!

Vitalii-Misechko (approved):
LGTM

PR #158, opened 2020-11-25
US120701 Settings page

bemailloux:
Sets up the settings page so that we can start working on the content. ~~It does not yet work with URL state.~~ It now works with URL state

Quality notes
* Testing
  * [x] Responsive
    * It scrolls correctly when there is enough content on the page to cause scrolling
    * Footer is still at the bottom of the page even if there isn't enough content to cause scrolling
  * [x] Browsers
    * Tested w/ Chrome+NVDA
    * D2L menus have interesting behaviour: you can't tab through the items but you can go through them with arrow keys. This is [expected behaviour](https://d2l.slack.com/archives/C0PHG3QB0/p1606404317237600). (@johngwilkinson you were right) 
  * [x] RTL
  * [x] Test cases
    * Settings page -> home page
      * via browser back button
      * via immersive nav back button
      * via cancel / save and close buttons
    * Home page -> settings page -> home page
      * via browser back button
      * via immersive nav back button
      * via cancel / save and close buttons
    * Browser forward button
* Security, perf, devops: N/A
* UX
  * [x] Demo to Kevin

@anhill-D2L @rohitvinnakota @johngwilkinson @NicholasHallman

### DIFF
--- a/components/dashboard-settings.js
+++ b/components/dashboard-settings.js
@@ -1,0 +1,153 @@
+import { css, html, LitElement } from 'lit-element';
+import { Localizer } from '../locales/localizer';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
+
+class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
+	static get styles() {
+		return [css`
+			:host {
+				display: flex;
+				flex-direction: column; /* required so the footer actually appears on-screen */
+				height: 100%;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+
+			.d2l-insights-settings-page-main-container {
+				height: 100%;
+				overflow-y: auto;
+			}
+
+			.d2l-insights-settings-page-main-content {
+				background-color: white;
+				margin: 0 auto;
+				max-width: 1230px;
+				padding-bottom: 72px; /* height of footer */
+				width: 100%;
+			}
+
+			footer {
+				background-color: white;
+				bottom: 0;
+				box-shadow: 0 -2px 4px rgba(73, 76, 78, 0.2); /* ferrite */
+				height: 42px; /* height of a d2l-button */
+				left: 0;
+				padding: 0.75rem 30px;
+				position: fixed;
+				right: 0;
+				z-index: 2; /* ensures the footer box-shadow is over main areas with background colours set */
+			}
+
+			.d2l-insights-settings-page-footer {
+				margin: 0 auto;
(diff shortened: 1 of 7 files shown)


bemailloux:
Known issue: https://github.com/Brightspace/insights-engagement-dashboard/pull/151#issuecomment-733909439

model/view-state.js
```diff
@@ -28,6 +28,15 @@ export class ViewState {
 		if (this._urlState) this._urlState.save();
 	}
 
+	setSettingsView() {
```
Vitalii-Misechko, line 31: Could you add unit test into `test\model\view-state.test.js`?

components/dashboard-settings.js
```diff
@@ -0,0 +1,160 @@
+import { css, html, LitElement } from 'lit-element';
+import { heading1Styles, heading2Styles } from '@brightspace-ui/core/components/typography/styles.js';
+import { Localizer } from '../locales/localizer';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
+
+class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
```
Vitalii-Misechko, line 6: Could you add some basic unit tests for this component? For instance:
* constructor test
* axe test
* event test

Vitalii-Misechko (approved):
LGTM!

test/model/view-state.test.js
```diff
@@ -60,5 +60,19 @@ describe('ViewState', () => {
 			const searchParams = new URLSearchParams(window.location.search);
 			expect(searchParams.get(KEY)).equals('user,321');
 		});
+
+		it('should save settings view to the url', async() => {
+			setStateForTesting(KEY, 'settings,0');
+
+			const sut = new ViewState();
+
+			sut.setSettingsView();
```
Vitalii-Misechko, line 69: you probably should change params in `setStateForTesting`-call because we can't say if the test checks proper deserialization from url or checks the update of its state via this method. Two separate tests would be optimal. Have a look at `should **load** user view from the url` and `'should **save** user view to the url'`

Vitalii-Misechko (approved):
LGTM

test/model/view-state.test.js
bemailloux: Thanks for pointing this out, I missed that there were 2 sets of tests

Vitalii-Misechko (approved):

NicholasHallman (approved):
Looks great!

PR #159, opened 2020-11-25
Fix: Explicit tree filter appearing in url state

NicholasHallman:
This is a revert of my changes


NicholasHallman:
@bemailloux this fixes the problem you are having with reloads

bemailloux (commented):
Oops my bad 😅

bemailloux (approved):

NicholasHallman:
> Oops my bad 😅

Nope it's on me I should have tested it lmao

johngwilkinson (approved):
Aha! I saw that, then it went away, so I thought maybe it was just a weird demo page thing. Thanks for the fix.

PR #161, opened 2020-11-26
Us121775/add active courses table

anhill-D2L:
functional
Tabbing through the table headers and using the spacebar to sort works
Tabbing into the table with NVDA reads "Active courses table with x rows and y columns course name sorted descending button". This is slightly different than in the users list table because the first row there is a checkbox which acts differently.
Checked on my local LMS and the demo page that inactive courses are filtered out

ux:
demo'd to Kevin
the table and "Active Courses" title are responsive just like in the users table

perf, security, devops, docs: n/a
image

FYI: @bemailloux @hyehayes @johngwilkinson @NicholasHallman @MykolaGalian


components/active-courses-table.js
```diff
@@ -0,0 +1,223 @@
...
+			user[TABLE_COURSES.DISCUSSION_ACTIVITY],
+			lastSysAccessFormatted
+		];
+	}
+
+	// @computed
+	get userDataForDisplay() {
+		// map to a 2D userData array, with column 1 as a sub-array of [id, FirstName, LastName, UserName]
+		// then sort by the selected sorting function
+		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
```
rohitvinnakota, line 154: To populate this with the data we want, we should check the bool value in the last element in the `orgunits` data attribute. If it's true, then it is an active course, and not if it is false.

components/user-drill-view.js
```diff
@@ -129,8 +133,16 @@ class UserDrill extends Localizer(MobxLitElement) {
...
 
+			<h2 class="d2l-heading-3">${this.localize('components.insights-active-courses-table.title')}</h2>
+
 			<div class="d2l-insights-user-drill-view-content">
-				<!-- put your tables here -->
+				
+				<d2l-insights-active-courses-table
+					.data="${this.data}"
+					.orgUnitTree="${this.orgUnitTree}"
+					?skeleton="${this.skeleton}">
```
Vitalii-Misechko, line 143: I don't see `skeleton` property you should  add one or better use `SkeletonMixin`

engagement-dashboard.js
```diff
@@ -252,6 +253,9 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		return html`
 			<d2l-insights-user-drill-view
 				.user="${user}"
+				.data="${this._data.recordsByUser.get(user.userId)}"
+				.orgUnitTree="${this._serverData.orgUnitTree}"
+				.skeleton="${this._isLoading}"
```
Vitalii-Misechko, line 258: If you add `SkeletonMixin` to `UserDrill` then you'll need change this line as well to pass value into boolean attribute

components/active-courses-table.js
```diff
@@ -0,0 +1,223 @@
...
+			></d2l-insights-table>
+		`;
+	}
+
+}
+decorate(CoursesTable, {
+	userDataForDisplay: computed,
+	userDataForDisplayFormatted: computed,
+	_sortColumn: observable,
+	_sortOrder: observable,
```
Vitalii-Misechko, line 220: `thought`: I guess `_sortColumn` and `_sortOrder` can be removed from the class properties. Now `lit-element` framework and `MobX` framework both track changes. If they don't fight with each other it's good )

components/active-courses-table.js
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * The mobx data object is doing filtering logic
+ *
+ * @property {Object} data - an instance of Data from model/data.js filtered to only one user
+ * @property {Object} orgUnitTree - an instance of orgUnitTree from model/data.js
+ * @property {Number} _sortColumn - The index of the column that is currently sorted
+ * @property {String} _sortOrder - either 'asc' or 'desc'
+ */
+class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
```
Vitalii-Misechko, line 33: Adding basic unit tests ( constructor/axe ) would be good

locales/en.js
```diff
@@ -144,5 +144,15 @@ export default {
 	"components.insights-default-view-popup.emptyResultsFromNRecentSemesters": "This dashboard is designed to look at portions of your organization's engagement. You do not have permission to review data in any courses in the most recently created {numDefaultSemesters} semesters.",
 	"components.insights-default-view-popup.expandDefaultCourseList": "Expand to see the courses included in your default view",
 	"components.insights-default-view-popup.collapseDefaultCourseList": "Collapse the list of courses included in your default view",
-	"components.insights-default-view-popup.buttonOk": "Ok"
+	"components.insights-default-view-popup.buttonOk": "Ok",
+
+	"components.insights-active-courses-table.title": "Active Courses",
```
johngwilkinson, line 149: see my recent OSLO PR: these should now be named like `activeCoursesTable.loadingPlaceholder`.

components/active-courses-table.js
```diff
@@ -0,0 +1,230 @@
...
+		};
+	}
+
+	_formatDataForDisplay(user) {
+		const lastSysAccessFormatted = user[TABLE_COURSES.COURSE_LAST_ACCESS]
+			? formatDateTime(new Date(user[TABLE_COURSES.COURSE_LAST_ACCESS]), { format: 'medium' })
+			: this.localize('usersTable:null');
+		const currentGrade = user[TABLE_COURSES.CURRENT_GRADE]
+			? formatPercent(user[TABLE_COURSES.CURRENT_GRADE] / 100, numberFormatOptions)
+			: this.localize('activeCoursesTable:noGrade');
```
anhill-D2L, line 145: I checked with keven, this is so when there are no grades, it matches with what is in Class Progress

components/active-courses-table.js
```diff
@@ -0,0 +1,230 @@
...
+		const reads = userRecords.get(RECORD.DISCUSSION_ACTIVITY_READS);
+		const replies = userRecords.get(RECORD.DISCUSSION_ACTIVITY_REPLIES);
+		const lastCourseAccess = userRecords.get(RECORD.COURSE_LAST_ACCESS) ? new Date(userRecords.get(RECORD.COURSE_LAST_ACCESS)) : undefined;
+
+		return [
+			this.localize('treeFilter:nodeName', { orgUnitName, id: orgUnitId }),
+			finalGrade,
+			predictedGrade,
+			timeInContent,
+			[threads, reads, replies],
```
rohitvinnakota, line 108: Do we want to use `Math.round` here?

rohitvinnakota (approved):

components/active-courses-table.js
anhill-D2L: I believe the threads/reads/replies should already be whole numbers as there is no averaging happening

NicholasHallman (approved):
Maybe for a future refactor it would be viable to take the sorting functions out of all the tables (since they are basically just copy pasted with minor changes) and into a separate hooks file to cleanup a bit. But overall this LGTM

PR #163, opened 2020-11-26
No Results Undo (US121746)

NicholasHallman:
## Implementation Notes

One of the issues I ran into involved a user trying to undo if they received a link that returned no filtered results. Hitting the undo button would simply redirect them back to the new tab page. To fix this I added a static count variable to the UrlState class so that it can keep track of where in the history the new popped state is. This way, if the user tries to undo the filters when they have no active filters in their history it will reset them to the default view. 

### Functional Testing
- Manual testing (chrome, edge, ff) (Also works in local lms)
  - In demo page, selecting org unit 9 returns no results, button shows, hitting the undo button unselects org unit 9
  - Copying the org unit 9 state into a new tab and hitting the undo button will whip the url state and refresh instead of going back to new tab.

- Unit tests
  - Moving through the history is covered by the urlstate tests, 
  - Checking the state whip isn't really possible since it causes a page refresh
### UX
 - Demoed to kevin

FYI @anhill-D2L @rohitvinnakota @johngwilkinson @bemailloux


engagement-dashboard.js
```diff
@@ -362,45 +346,78 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 	}
 
 	get _courseAccessCard() {
-		if (!this.showCourseAccessCard) return '';
+		if (!this.showCourseAccessCard || this._isNoUserResults) return '';
```
Vitalii-Misechko, line 349: Just curious why the code hides each component individually instead of hiding some parent `div` that contains all these components?

engagement-dashboard.js
```diff
@@ -362,45 +346,78 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
...
+				<d2l-button-subtle
+					icon="d2l-tier1:email"
+					text="${this.localize('components.insights-engagement-dashboard.emailButton')}"
+					@click="${this._handleEmailButtonPress}">
+				</d2l-button-subtle>
+			</d2l-action-button-group>
+
+			<d2l-insights-users-table
+				.data="${this._data}"
+				?skeleton="${this._isLoading}"
```
Vitalii-Misechko, line 397: you should restore `@d2l-insights-users-table-cell-clicked="${this._userTableCellClicked}"` line

Vitalii-Misechko (approved):
LGTM other than mentioned in comments

engagement-dashboard.js
Vitalii-Misechko: It breaks user drill navigation from main view table

engagement-dashboard.js
NicholasHallman: I didn't want to introduce a dive because it didn't add any semantic value to the page. These functions for showing and hiding the elements based on preferences were already there so I figured this could piggy back off of that and still be understandable

components/message-container.js
```diff
@@ -1,3 +1,4 @@
+import 'd2l-button';
```
bemailloux, line 1: Please use the button from `@brightspace-ui/core` directly: 
```
import '@brightspace-ui/core/components/button/button';
```
(No other changes are required because you _are_ already using it, just [in a roundabout way](https://github.com/BrightspaceUI/button/blob/master/d2l-button.js))

bemailloux (approved):

components/message-container.js
NicholasHallman: Ah! too funny thank you

Vitalii-Misechko (approved):

PR #164, opened 2020-11-26
Us120705/sa threshold pref

johngwilkinson:
Quality Notes
- functional: test different values via demo and LMS; with some data munging to set last access time to N days ago, confirmed both card count and filtering of table rows works as expected, with both higher and lower values of the preference
- devops/perf/security: n/a
- ux/docs: split langterm so the message is grammatical if the preference is set to "1"

FYI @anhill-D2L @bemailloux @MykolaGalian @rohitvinnakota


NicholasHallman (approved):

Vitalii-Misechko (approved):
LGTM

PR #169, opened 2020-11-30
User filters - US121776

NicholasHallman:
## Implementation Notes (optional)

Some interesting quirks of the structure is that the semester filter and the org filter selectors need to pass their event listeners up through the user drill page and then again to engagement. If anybody knows of a better way that would be 💯.

### Functional Testing (WIP)
 - Manual tests
   - When a semmester or orgunit is changed the records in the data object will change. This shows that the filters are being 
applied properly
 - Integration test
   - TODO check that something on the page or the data structures are changed as a result of a selection
### UX
 - Requires demo to kevin

fyi @anhill-D2L @MykolaGalian @rohitvinnakota @johngwilkinson @bemailloux

### DIFF
--- a/components/user-drill-view.js
+++ b/components/user-drill-view.js
@@ -1,9 +1,13 @@
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/button/button.js';
+import './semester-filter';
+import './ou-filter';
 import { bodySmallStyles, heading2Styles } from '@brightspace-ui/core/components/typography/styles.js';
 import { css, html } from 'lit-element/lit-element.js';
 import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
+import { RECORD } from '../consts';
+import { SemesterSelectorFilter } from '../model/selectorFilters';
 
 /**
  * @property {Object} user - {firstName, lastName, username, userId}
@@ -12,6 +16,8 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 class UserDrill extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
+			data: { type: Object, attribute: true },
+			isDemo: { type: Boolean, attribute: 'demo' },
 			user: { type: Object, attribute: false }
 		};
 	}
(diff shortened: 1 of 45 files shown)


components/user-drill-view.js
```diff
@@ -131,6 +148,19 @@ class UserDrill extends Localizer(MobxLitElement) {
 				@click="${this._composeEmailHandler}"
 			>${this.localize('components.insights-engagement-dashboard.emailButton')}</d2l-button>
 
+			<div class="d2l-insights-view-filters-container">
+				<d2l-insights-ou-filter
+					.data="${this.data}"
```
NicholasHallman, line 153: I'm not sure if we should only send in the data the user drill need, like in @anhill-D2L [pr](https://github.com/Brightspace/insights-engagement-dashboard/pull/161/files#diff-da16422f38b47f7e8f1ecb041f786890b12adf07da4b62f2ab5a7d539b44a9e3R256). We can do something similar here but the ou-filter expects the whole data object, so it would be confusing to pass in a filtered object. If later somebody made changes to ou-filter there's the chance it may work in the engagement dashboard but not in the user-drill page.

johngwilkinson: I think of it this way: we have a model, and we have some views of it (the home view and, potientially, a view for each userid). I don't think it clarifies anything to build a separate model just for the user view.

On the other hand, I'm slightly surprised we ended up instantiating each of these filter components twice. This is related to the need to bubble up the events: I wonder if a little re-organization would straighten this out. Instead of having two completely distinct pages, what if the structure were something like:
<maybe a header, like a title and the user profile - depends on view>
<ou filter><semester filter>
<body (cards, tables) - depends on view>

Then there would only be one instance of each filter component, owned directly by the dashboard component.

NicholasHallman: That makes sense to me, Having a base page that hosts the serverdata and view which then draws the views for either the home dashboard or the user drill makes a lot of sense. This could be called Routes (because it controls the view) or Base

Vitalii-Misechko: JFYI @NicholasHallman 
you probably have seen `EngagementDashboard::render` and `ViewState`. These two have basic implementation of the base page and routing respectively

NicholasHallman: Instead of having a shared header I used a slot to inject the shared filters into the user-drill. Much easier than moving everything out of the user drill page. And because the headers didn't have that much in common it made sense to keep the code separate instead of having one big engagement-dashboard file

johngwilkinson (approved):
The use of slots is very elegant!

Vitalii-Misechko (approved):
Neat!

PR #170, opened 2020-11-30
Us120707/save settings

johngwilkinson:
Quality Notes
- functional
  - LMS call works
  - update a value (hardcoded), and saving updates the home view
- devops/perf/security/ux/docs: n/a

FYI @anhill-D2L @MykolaGalian @NicholasHallman @rohitvinnakota

### DIFF
--- a/test/model/lms.test.js
+++ b/test/model/lms.test.js
@@ -306,16 +306,11 @@ describe('Lms', () => {
 
 			await saveSettings(settings);
 
-			const actualOptions = fetchMock.lastOptions();
-			expect(actualOptions).to.deep.include({
-				method: 'PUT',
-				headers: {
-					'content-type': 'application/json',
-					'x-csrf-token': 'token'
-				}
-			});
-			// fetchMock wraps the request body in a promise
-			expect(await actualOptions.body).to.equal(JSON.stringify(settings));
+			const request = fetchMock.lastCall().request; // assumes fetch was called with a Request
+			expect(request.method).to.equal('PUT');
+			expect(request.headers.get('content-type')).to.equal('application/json');
+			expect(request.headers.get('x-csrf-token')).to.equal('token');
+			expect(await request.json()).to.deep.equal(settings);
 		});
 	});
 });
--- a/test/model/lms.test.js
+++ b/test/model/lms.test.js
@@ -298,6 +298,7 @@ describe('Lms', () => {
 				showDiscussionsCol: true,
 				showLastAccessCol: true
 			};
+			window.localStorage.setItem('XSRF.Token', 'token');
 			fetchMock.put(
 				'end:/d2l/api/ap/unstable/insights/mysettings/engagement',
 				200
@@ -305,13 +306,16 @@ describe('Lms', () => {
 
 			await saveSettings(settings);
 
-			expect(fetchMock.lastOptions()).to.deep.equal({
+			const actualOptions = fetchMock.lastOptions();
+			expect(actualOptions).to.deep.include({
 				method: 'PUT',
 				headers: {
-					'Content-Type': 'application/json'
-				},
-				body: JSON.stringify(settings)
+					'content-type': 'application/json',
+					'x-csrf-token': 'token'
+				}
 			});
+			// fetchMock wraps the request body in a promise
+			expect(await actualOptions.body).to.equal(JSON.stringify(settings));
 		});
 	});
 });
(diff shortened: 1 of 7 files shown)


test/model/lms.test.js
```diff
@@ -298,20 +298,24 @@ describe('Lms', () => {
 				showDiscussionsCol: true,
 				showLastAccessCol: true
 			};
+			window.localStorage.setItem('XSRF.Token', 'token');
```
bemailloux, line 301: Oh - so the only thing d2l-fetch does is it grabs the XSRF token from local storage and adds it to the request?

EDIT: ah, I guess d2l-fetch-auth is responsible for making sure that XSRF token is there then

bemailloux (approved):

test/model/lms.test.js
johngwilkinson: Yup It can also do things like get an oauth token if the call is to a trusted off-stack url. I could have written the line of code that gets the token from localstorage, but I didn't want to break if somebody changes the name some day.

johngwilkinson:
Huh. Test timeout on the event-firing test. I was seeing this locally, then it went away, so I'm guessing it's a race of some sort.

johngwilkinson:
> Huh. Test timeout on the event-firing test. I was seeing this locally, then it went away, so I'm guessing it's a race of some sort.

Oh, it's because it's trying to contact the server to save the settings. Standby while I refactor to make the lms call mockable.

johngwilkinson:
Drat: ```Safari 13.0.1 (Mac OS X 10.13.6) Lms saveSettings should send the settings to the lms FAILED
	expected '[object ReadableStream]' to equal '{"showResultsCard":true...}```
That might even be a real issue on mac, but most likely just another fetch-mock glitch.

bemailloux:
> Drat: `Safari 13.0.1 (Mac OS X 10.13.6) Lms saveSettings should send the settings to the lms FAILED expected '[object ReadableStream]' to equal '{"showResultsCard":true...}`
> That might even be a real issue on mac, but most likely just another fetch-mock glitch.

This rings a bell! See https://github.com/Brightspace/insights-engagement-dashboard/pull/130#discussion_r516004176

johngwilkinson:
> > Drat: `Safari 13.0.1 (Mac OS X 10.13.6) Lms saveSettings should send the settings to the lms FAILED expected '[object ReadableStream]' to equal '{"showResultsCard":true...}`
> > That might even be a real issue on mac, but most likely just another fetch-mock glitch.
> 
> This rings a bell! See [#130 (comment)](https://github.com/Brightspace/insights-engagement-dashboard/pull/130#discussion_r516004176)

Thanks - looks like the same issue. That code is also using d2l-fetch, which calls `fetch(<Request>)` instead of `fetch(url, options)`. Something about the way fetchmock infers the options from the Request object is quite wonky.

Vitalii-Misechko (approved):
LGTM

Vitalii-Misechko:
> > > Drat: `Safari 13.0.1 (Mac OS X 10.13.6) Lms saveSettings should send the settings to the lms FAILED expected '[object ReadableStream]' to equal '{"showResultsCard":true...}`
> > > That might even be a real issue on mac, but most likely just another fetch-mock glitch.
> > 
> > 
> > This rings a bell! See [#130 (comment)](https://github.com/Brightspace/insights-engagement-dashboard/pull/130#discussion_r516004176)
> 
> Thanks - looks like the same issue. That code is also using d2l-fetch, which calls `fetch(<Request>)` instead of `fetch(url, options)`. Something about the way fetchmock infers the options from the Request object is quite wonky.

Probably we should try out `msw` It looks promising https://kentcdodds.com/blog/stop-mocking-fetch

johngwilkinson:
> > > > Drat: `Safari 13.0.1 (Mac OS X 10.13.6) Lms saveSettings should send the settings to the lms FAILED expected '[object ReadableStream]' to equal '{"showResultsCard":true...}`
> > > > That might even be a real issue on mac, but most likely just another fetch-mock glitch.
> > > 
> > > 
> > > This rings a bell! See [#130 (comment)](https://github.com/Brightspace/insights-engagement-dashboard/pull/130#discussion_r516004176)
> > 
> > 
> > Thanks - looks like the same issue. That code is also using d2l-fetch, which calls `fetch(<Request>)` instead of `fetch(url, options)`. Something about the way fetchmock infers the options from the Request object is quite wonky.
> 
> Probably we should try out `msw` It looks promising https://kentcdodds.com/blog/stop-mocking-fetch

That might be interesting if we can replace `fakeLms` with the same code.

For now though, I found the right way to get json from a Request object (request.body is actually supposed to be a ReadableStream; and request.json() is a Promise), and have updated the two tests that had this issue.

PR #171, opened 2020-12-01
US120706 Settings page: card selection list

bemailloux:
I expect this will conflict with other ongoing work. Happy to either merge first or resolve conflicts.

Quality notes
* Testing
  * [x] Accessibility (keyboard nav / screen reader - tested with NVDA/Chrome)
  * [x] RTL
  * [x] Responsive
  * [x] Test cases
    * Loading: if no settings exist yet for this user, it uses the defaults
    * Loading: if settings exist for this user, it loads with those
    * Display: after save and close, engagement dashboard uses the new settings
    * Display: if cancel button is pressed, dashboard uses old settings
      * Special case: last system access card: it uses the new threshold value (even if the card is hidden)
    * Reload after saving: saved settings are persisted
* Security, devops, perf: N/A
* UX
  * [x] Demo to Kevin

Screenshots (90% zoom so everything would fit)
image
image

Responsive layout
image

@MykolaGalian @rohitvinnakota @anhill-D2L @NicholasHallman


Vitalii-Misechko:
I like this structure of the quality notes. 👍   I've borrowed as template for my further PRs

components/svg/course-access-thumbnail.svg.js
```diff
@@ -0,0 +1,76 @@
+import { css, html, LitElement } from 'lit-element/lit-element';
+import { Localizer } from '../../locales/localizer';
+import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
+
+class CourseAccessThumbnailSvg extends RtlMixin(Localizer(LitElement)) {
+	static get styles() {
+		return [css`
```
Vitalii-Misechko, line 7: It seems that styles can be shared. Mixin might be good option here

components/svg/course-access-thumbnail.svg.js
```diff
@@ -0,0 +1,76 @@
...
+			.d2l-insights-course-access-thumbnail-title {
+				font-size: smaller;
+				font-weight: bold;
+				margin-left: 16px;
+				margin-right: 0;
+				margin-top: 5px;
+				position: absolute;
+				z-index: 1;
```
Vitalii-Misechko, line 15: Have you checked how `z-index` interact with immersive nav? Is it possible to get rid of `z-index`?

engagement-dashboard.js
```diff
@@ -392,6 +398,10 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 
 	get _lastAccessCard() {
 		if (!this.showSystemAccessCard) return '';
+
+		// update LastSystemAccess filter's threshold, as it may have changed (e.g. if new settings were saved)
+		this._data.getFilter(lastAccessFilterId).thresholdDays = this.lastAccessThresholdDays;
```
Vitalii-Misechko, line 403: I think this code should reside inside of `_backToHomeHandler`-function. Not sure if this change can be done in this PR though

Vitalii-Misechko (approved):
Screenshot looks awesome!

engagement-dashboard.js
Vitalii-Misechko: It seems that your PR accomplishes goals in this story 
[US120708](https://rally1.rallydev.com/#/detail/userstory/437184259152?fdp=true): [Engage][Config] Set system access threshold

components/svg/course-access-thumbnail.svg.js
bemailloux: Immersive nav has z-index: 2 and this has z-index: 1. So it works... for now. 

I'll see if there is a way to get rid of it

engagement-dashboard.js
bemailloux: Oh! I had no idea this was a separate story

bemailloux: > I think this code should reside inside of _backToHomeHandler-function.

That's funny, I talked to John about this and he thought it should go somewhere in _render, because that seems like the place where it makes sense to handle property changes. 

There's a possibility that putting this code in `_backToHomeHandler` will cause an extra re-render - I'll check if this is the case. But otherwise I don't have a preference either way

Vitalii-Misechko: This is a good topic to discuss. My preferences are based on the idea of placing all changes that the Settings page does in one place.

locales/localizer.js
```diff
@@ -90,4 +90,9 @@ export const Localizer = superclass => class extends LocalizeMixin(superclass) {
 			};
 		}
 	}
+
+	truncateString(str, maxLength) {
```
johngwilkinson, line 94: I was wondering which mixin this was defined in, and my guess was the RTL one, since the truncation would happen at the other end of the string for RTL languages. I wonder if D2L has a component or lib for this.

components/card-selection-list.js
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	_renderSystemAccessListContents() {
+		const summaryCardMessage = this.localize(
+			'components.insights-engagement-dashboard.lastSystemAccessMessage',
+			{ thresholdDays: this.lastAccessThresholdDays }
+		);
+
+		// hack to get 2 parts of a localized version of a string
+		const editTextParts = this.localize('components.insights-settings-view.systemAccessEdit', { num: '{num}' }).split('{num}');
```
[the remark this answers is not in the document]
bemailloux: Because splitting a sentence the way we do in English doesn't apply in other languages. Case in point: this string in particular:

English: `Show users who have not accessed the system in the last {num} days.`
French (rough translation): `Montrer les utilisateurs qui n'ont pas utilisé le système dans les {num} derniers jours.`

If you had a before and after, the French would put the word "derniers" (last) in the wrong spot. Which isn't wrong in this case, just awkward. But in other cases it could literally be wrong.

locales/localizer.js
bemailloux: I don't think of this as an RTL thing because the end of the string is still the end of the string in RTL languages. As long as the formatting is symmetrical truncation in RTL should work exactly the same as truncation in LTR.

That being said, I haven't looked for any existing utils - I'll do that.

engagement-dashboard.js
johngwilkinson: Another possible solution: make the threshold property on the dashboard `observable`, initialize the filter with `this` instead of passing `this.lathreshold` (call it `config` maybe), and have the filter reference `config.lathreshold`. Then setting the threshold is the same as setting any of the card properties.

components/card-selection-list.js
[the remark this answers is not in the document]
johngwilkinson: I think we're making different assumptions about how the translators work. I see a before and after string in both languages - but I think you're suggesting the translations will happen fairly literally and in isolation, which may be true in practice. Given that, I'm okay with this solution.

locales/localizer.js
johngwilkinson: The characters are rendered RTL in those languages - so the langterms are all "backwards"? This is the kind of question I'd rather leave to a shared utility. :)

bemailloux: TIL: CSS `text-overflow: ellipsis` is a thing

https://d2l.slack.com/archives/C0PHG3QB0/p1606838071257600?thread_ts=1606838016.257500&cid=C0PHG3QB0

johngwilkinson: 😮 Cool!

components/card-selection-list.js
```diff
@@ -0,0 +1,203 @@
...
+			showDiscussionsCard: this.showDiscussionsCard,
+			showSystemAccessCard: this.showSystemAccessCard,
+			showGradesCard: this.showGradesCard,
+			showTicGradesCard: this.showTicGradesCard,
+			showCourseAccessCard: this.showCourseAccessCard,
+			lastAccessThresholdDays: this.lastAccessThresholdDays
+		};
+	}
+}
+customElements.define('d2l-card-selection-list', CardSelectionList);
```
johngwilkinson, line 203: minor: this component sounds more general than it is: maybe `d2l-insights-engagement-card-selection`?

bemailloux: Done

components/svg/course-access-thumbnail.svg.js
bemailloux: It wasn't needed at all. Not sure why the title wasn't appearing before, but it is appearing now 🤷

johngwilkinson:
🍿 (watching the race to merge first...)

engagement-dashboard.js
bemailloux: It looks like Mobx doesn't want to allow the threshold property on the dashboard to be `observable`. 

I don't really know why, but am suspecting that it might not be a good idea to make a property that is also reflected as an attribute mobx-observable (lots of things would get updated at the same time). 

I think I'd prefer to stick to LitElement updates here.

@Vitalii-Misechko looks like putting the update inside `_backToHomeHandler` does not cause an extra render, so I'll probably move this code over there so all the updates can be in the same place.

Vitalii-Misechko (approved):

locales/en.js
```diff
@@ -150,6 +150,19 @@ export default {
 
 	"components.insights-settings-view.title": "Settings",
 	"components.insights-settings-view.description": "Set which metrics display in the Summary and Result Detail section of the Engagement Dashboard.",
+	"components.insights-settings-view.tabTitleSummaryMetrics": "Summary Metrics",
+	"components.insights-settings-view.tabTitleResultsTableMetrics": "Results Table Metrics",
+
+	"components.insights-settings-view.currentGradeDesc": "The Current Grade card shows the current grade for each enrollment per user. Data shown will appear both with the card and in the Result Detail table.",
```
johngwilkinson, line 156: Can we change all instances of `.insights-settings-view.` to `.insights-engagement-settings.`? There will hopefully be other insights dashboards with settings.

bemailloux: Different PR? This change will affect a bunch of the already-translated terms and would clutter up this PR

johngwilkinson (commented):
Is it white or blue? The design has a white background, as does your responsive screenshot, but the other ones have blue, which seems to be the d2l-list default.

bemailloux:
> Is it white or blue? The design has a white background, as does your responsive screenshot, but the other ones have blue, which seems to be the d2l-list default.

Each list item is blue when selected and white when unselected

johngwilkinson:
Does this have a white or a blue background? The design has white but d2l-list defaults to blue.

> > Is it white or blue? The design has a white background, as does your responsive screenshot, but the other ones have blue, which seems to be the d2l-list default.
> 
> Each list item is blue when selected and white when unselected

Aha! Thanks. UX is okay with that? (I assume my story will be the same)

bemailloux:
> Aha! Thanks. UX is okay with that? (I assume my story will be the same)

I'll ask in my demo to Kevin. Should be fine since it's default behaviour from d2l-list.

bemailloux:
I changed the responsive view so that titles would disappear on small screens, as requested by UX.

image

components/card-selection-list.js
```diff
@@ -98,29 +108,38 @@ class CardSelectionList extends RtlMixin(Localizer(LitElement)) {
...
 				<d2l-list-item key="showGradesCard" selectable ?selected="${this.showGradesCard}">
 					<div class="d2l-insights-list-flex-container">
-						<d2l-insights-current-grade-thumbnail aria-hidden="true" class="d2l-demo-card"></d2l-insights-current-grade-thumbnail>
+						<d2l-insights-current-grade-thumbnail class="d2l-demo-card" aria-hidden="true"></d2l-insights-current-grade-thumbnail>
 						<div class="d2l-card-selection-text">
-							<h3 class="d2l-heading-3">${this.localize('components.insights-current-final-grade-card.currentGrade')}</h3>
+							<h3 class="d2l-heading-3 d2l-card-selection-title" aria-hidden="true">
+								${this.localize('components.insights-current-final-grade-card.currentGrade')}
+							</h3>
+							<d2l-offscreen>${this.localize('components.insights-current-final-grade-card.currentGrade')}</d2l-offscreen>
```
johngwilkinson, line 116: Is this so that the title is always available to the screen-reader, or is there another reason for the offscreen?

bemailloux: Yes, it's so the title is always read out. What happens is this (and is subject to change once Carin gets her hands on this page):

* Visual users
  * Large screens: Visualization w/ title, title, description
  * Small screens: Visualization w/ title, description
* Screen reader users
  * Any size screen: Title (from d2l-offscreen), description

PR #172, opened 2020-12-01
Add contact Support message

MykolaGalian:
[US120035](https://rally1.rallydev.com/#/detail/userstory/423589167812?fdp=true): [Engagement] Error when we fail -- contact Support

Functional:
Local App and LMS work as expected with test data.

Accessibility: n/a
Security: n/a
Devops: n/a
UX: demo - done

FYI @rohitvinnakota @bemailloux  @Vitalii-Misechko @anhill-D2L


model/lms.js
```diff
@@ -24,7 +28,8 @@ export async function fetchData({ roleIds = [], semesterIds = [], orgUnitIds = [
 	}
 	url.searchParams.set('defaultView', defaultView ? 'true' : 'false');
 	const response = await fetch(url.toString());
-	return await response.json();
+	if (response.ok) return await response.json();
+	else {return Promise.resolve({	...emptyServerData });}
```
johngwilkinson, line 32: Exceptions are generally the best way to propagate errors. This can just `throw new Exception('query failure')`; then `data.loadData` can catch it and set the `isQueryFailed` property. This will be clearest if you make that method `async`. Then something like (pseudocode):
```
async loadData() {
   try {
     const data = await this.recordProvider();
     this.isQueryFailed = false;
     this.onServerDataReload(data);
   } catch (ignored) {
     this.isQueryFailed = true;
   }
   
```

model/data.js
```diff
@@ -41,15 +44,22 @@ export class Data {
...
 			orgUnitIds: newOrgUnitIds || this._selectorFilters.orgUnit.selected,
 			defaultView
 		};
-		this.recordProvider(filters).then(data => this.onServerDataReload(data));
+		try {
+			const data = await this.recordProvider(filters);
+			this.onServerDataReload(data);
+			this.isQueryError = false;
+		} catch (ignored) {
+			this.onServerDataReload(emptyData);
```
johngwilkinson, line 60: maybe it would be better to do `this.onServerDataReload(this.serverData)` - that way if an error occurs when some data was already showing, it won't vanish

components/message-container.js
```diff
@@ -51,6 +51,16 @@ class MessageContainer extends Localizer(MobxLitElement) {
 				width: 73vw;
 			}
 
+			.d2l-insights-message-container-body-queryFails {
```
NicholasHallman, line 54: I think we should have a default base class and add more classes to specify the differences between the messages, as this one is identical to "tooManyResults"

maybe 
`.d2l-insights-message-container-body`
and if changes are needed we make it 
`.d2l-insights-message-container-body.d2l-insights-message-queryFails`

NicholasHallman (approved):

components/message-container.js
MykolaGalian: We will fix CSS in the next PR

johngwilkinson (approved):

PR #176, opened 2020-12-02
Use MSW instead of fetch-mock

Vitalii-Misechko:
Use MSW instead of fetch-mock

- [ ] Serve `mockServiceWorker.js` from `demo` folder or exclude the file from deploying to prod. See [es-dev-server](https://www.npmjs.com/package/es-dev-server) plugins
- [ ] Update tests

image


mockServiceWorker.js
```diff
@@ -0,0 +1,241 @@
+/**
```
Vitalii-Misechko, line 1: auto generated file

Vitalii-Misechko:
trying out on AQ repo

PR #182, opened 2020-12-02
User Drill Email - US121803

NicholasHallman:
This pr makes the email callback window appear when you click on the email button in the user drill

### Functional Testing
  - Verified it works by going into my local LMS, selecting a user, and clicking the button
  - The result is the window opening and the user's email appearing in the Bcc.

### All others - N/A

FYI @MykolaGalian @anhill-D2L @rohitvinnakota @johngwilkinson

### DIFF
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -260,6 +260,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		return html`
 			<d2l-insights-user-drill-view
 				.user="${user}"
+				.data="${this._serverData}"
 				@d2l-insights-user-drill-view-back="${this._backToHomeHandler}"
 			>
 				<div slot="filters">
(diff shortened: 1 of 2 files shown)


engagement-dashboard.js
```diff
@@ -260,6 +260,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		return html`
 			<d2l-insights-user-drill-view
 				.user="${user}"
+				.data="${this._serverData}"
```
bemailloux, line 263: I think passing the whole _serverData object just to get the org unit id is overkill. The same id just happens to be stored in the `orgUnitId` property on `engagement-dashboard` - it would probably be better to use that instead.

Vitalii-Misechko (approved):
LGTM

bemailloux (approved):

PR #184, opened 2020-12-03
US120707: update title and descriptions for column preferences

johngwilkinson:
Text requested by PM and approved via screen shot.

(BM: I also got approval via screenshot)


locales/en.js
```diff
@@ -169,12 +169,12 @@ export default {
...
-	"settings:lastAccessedSystemDescription": "The System Last Access indicator displays the last session date and local time that a user has accessed the system and any subsequent courses in the system.",
+	"settings:avgGradeDescription": "The Average Grade indicator presents the current average grade for the user across all the courses included in the applied filters.",
+	"settings:avgTimeInContentDescription": "The Average Time in Content indicator shows the average time spent in content, as an average of total time per course, for the user across all the courses included in the applied filters.  The metric is reported in minutes.",
+	"settings:avgDiscussionActivityDescription": "The Average Discussion Participation indicator presents user statistics for how often the user creates a thread, reads a post or replies to a post across all the courses included in the applied filters.  This metric averages the total count per course.",
+	"settings:lastAccessedSystemDescription": "The System Last Access indicator displays the timestamp of the last time the user accessed the system in any way.",
```
anhill-D2L, line 179: What time zone is the timestamp in?

locales/en.js
```diff
@@ -169,12 +169,12 @@ export default {
...
-	"settings:avgTimeInContentDescription": "The Average Time in Content indicator shows the average time spent in content across all the applied filtered fields. The metric is tracked in minutes.",
-	"settings:avgDiscussionActivityDescription": "The Average Discussion Participation Summary indicator presents user statistics for average reading, posting, and responding to discussions across all the applied filtered fields.",
-	"settings:lastAccessedSystemDescription": "The System Last Access indicator displays the last session date and local time that a user has accessed the system and any subsequent courses in the system.",
+	"settings:avgGradeDescription": "The Average Grade indicator presents the current average grade for the user across all the courses included in the applied filters.",
+	"settings:avgTimeInContentDescription": "The Average Time in Content indicator shows the average time spent in content, as an average of total time per course, for the user across all the courses included in the applied filters.  The metric is reported in minutes.",
```
anhill-D2L, line 177: I think it should be "This metric.... "
That will also match the wording in `settings:avgDiscussionActivityDescription`

johngwilkinson: Instead of "The ... indicator...", you mean?

locales/en.js
bemailloux: If I understand correctly, it's whatever timezone is set by `new Date()` which IIRC is whatever timezone is set on the user's PC. 

It goes through a `formatDateTime` function from `@BrightspaceUI/intl` but I don't think that has any effect on the timezone, just how the date is formatted.

johngwilkinson: Actually, if you drill in to that method you'll see it reads the timezone from the html tag, same as it does the language.

johngwilkinson: Talking with PM now about wording for that.

bemailloux: > it reads the timezone from the html tag, same as it does the language

Yes, but it doesn't use the html timezone to determine what time it is. It's only used to put at the end of the time string as a replacement for "ZZZ".

https://github.com/BrightspaceUI/intl/blob/c3a5165c442c40f7630a343718dffe93bb6b4cfb/lib/dateTime.js#L948

johngwilkinson: Oh! Then I suspect we have a bug to fix, separate from this PR. I'll take a look.

bemailloux: What bug?

johngwilkinson: We're supposed to be displaying the date in the configured LMS timezone. I'll confirm what we're actually doing and check with the PM if we should change it, but I know that's her current expectation.

bemailloux: Oh. That could be weird for users who aren't located in the same place as their school / organization. Probably doesn't happen often, but I'd expect it happens enough to be a problem.

locales/en.js
```diff
@@ -159,22 +159,22 @@ export default {
...
+	"settings:currentGradeDesc": "The Current Grade card shows the current grade for each enrollment per user for courses that are filtered on.",
+	"settings:courseAccessDesc": "The Course Access card shows the last access in a course for each enrollment per user for courses that are filtered on.",
+	"settings:ticVsGradeDesc": "The Time in Content versus Grade card shows the total time spent in a course relative to the current grade for each enrollment per user for courses that are filtered on. The chart is mapped into quadrants based on the average grade or time spent for the users in view.",
+	"settings:overdueAssignmentsDesc": "The Overdue Assignments card shows the number of users who have one or more assignments overdue in the courses that are filtered on.",
+	"settings:systemAccessDesc": "The System Access card shows the last access the user had in any way.",
+	"settings:discActivityDesc": "The Discussion Activity card shows passive and active social engagement in each course that is filtered on. The metric reports when a user creates a thread, replies to an existing post, or reads a post.",
```
NicholasHallman, line 167: I think this should also be "**This metric** reports when a user creates..."

NicholasHallman (approved):

anhill-D2L (approved):

johngwilkinson:
Merging now: the date description is currently incorrect, but there is a defect for that which is currently underway.

PR #185, opened 2020-12-03
US118252 [Engagement][Filters] Org Unit Pruning the Tree

Vitalii-Misechko:
[US118252](https://rally1.rallydev.com/#/detail/userstory/402789989336?fdp=true): [Engagement][Filters] Org Unit Pruning the Tree

* Functional Testing
  * OrgUnit filter
    * A user enrolled into a course offering only: the user sees only course offerings
    * A user enrolled into a course template: the user sees only course offerings
    * A user enrolled into a department: the user sees a tree that starts from nodes of course template type
    * A user enrolled into multiple departments: the user sees a tree that starts from nodes of department type
    * A user enrolled into two departments: the user filters out one department with the semester filter then the user sees a pruned tree as if the user had only one department that left from the start.
    * Search mode: does not show pruned nodes
* Security, devops: N/A
* Performance
   * I don't expect any impact of performance. Filtering is based on preprocessing of relatively small org unit tree with caching results to a map.
* UX and docs
  * [x] Demo

Pic 1

image

Pic 2

image

Pic 3

image


Vitalii-Misechko:

components/tree-filter.js
```diff
@@ -49,6 +50,7 @@ export class Tree {
 		this._open = oldTree ? new Set(oldTree.open) : new Set();
 		// null for no filter, vs. empty Set() when none match
 		this._visible = null;
+		this._pruned = new Map();
```
johngwilkinson, line 53: in general, a `Set` makes more sense than a `Map` with boolean values.

components/tree-filter.js
```diff
@@ -230,7 +239,7 @@ export class Tree {
 
 	getMatchingIds(searchString) {
 		return this.ids
-			.filter(x => this._isVisible(x))
+			.filter(x => this._isVisible(x) && !this._isPruned(x))
```
johngwilkinson, line 242: Is it important to omit pruned nodes from the search? Might it confuse a user if they search for their department and it doesn't show up?

If we don't need to affect the search, then I think we can drop `this._pruned()`, and just make a smaller change to `getChildIdsForDisplay(id, pruning)`: `if ((pruning || this._isRoot(id)) && children.length === 1) return this.getChildIdsForDisplay(children[0], true);`. Are there other issues with this approach?

Vitalii-Misechko: > Is it important to omit pruned nodes from the search? Might it confuse a user if they search for their department and it doesn't show up?

I'll ask Kevin or Megan. It looked confusing to see pruned departments in search to me.

> If we don't need to affect the search, then I think we can drop this._pruned(), and just make a smaller change to getChildIdsForDisplay(id, pruning): if ((pruning || this._isRoot(id)) && children.length === 1) return this.getChildIdsForDisplay(children[0], true);. Are there other issues with this approach?

I believe this approach would prune some intermediate nodes. Imagine a tree with few departments, each department has exactly one course template and many courses for each template. I assume opening a department node will output courses, skipping a course template for the chosen department. Code is neat though

johngwilkinson: It shouldn't prune intermediate nodes because `pruning` will be false. But they UX question for Megan is most important anyway.

Vitalii-Misechko: > Is it important to omit pruned nodes from the search? Might it confuse a user if they search for their department and it doesn't show up?

We should show pruned node when a user searches. I'll update the code

components/tree-filter.js
Vitalii-Misechko: good point. thanks

Vitalii-Misechko:
> A user enrolled into two departments: the user filters out one department with the semester filter then the user sees a tree that starts from nodes of the pruned tree as if it root started from the department that left

This case works after the last code change. I've added unit test for it as well

bemailloux (approved):

components/tree-filter.js
```diff
@@ -215,10 +216,13 @@ export class Tree {
...
 	}
 
-	getChildIdsForDisplay(id) {
-		return this.getChildIds(id)
-			.filter(x => this._isVisible(x))
-			.sort((a, b) => this._nameForSort(a).localeCompare(this._nameForSort(b)));
+	getChildIdsForDisplay(id, pruning) {
+		const children = this.getChildIds(id).filter(x => this._isVisible(x));
+
+		const isPrunning = (pruning || this._isRoot(id)) && children.length === 1 && this.getType(children[0]) !== COURSE_OFFERING;
```
johngwilkinson, line 222: nit: one 'n' in `isPruning`

johngwilkinson (approved):

johngwilkinson (approved):

PR #188, opened 2020-12-04
DE41591: use lms timezone setting for last accessed date

johngwilkinson:
Quality Notes
- functional
  - demo page: change timezone in html tag and check that times adjust
  - lms: change timezone in account settings and verify times change (including a tz that pushed a date across a day boundary)
- devops: relies on new version of intl; will tag a release after this merges so that BSI can pick it up
- security/perf/ux: n/a

FYI @MykolaGalian @NicholasHallman @rohitvinnakota @Vitalii-Misechko

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@adobe/lit-mobx": "0.0.x",
     "@brightspace-ui-labs/pagination": "^1",
     "@brightspace-ui/core": "^1.102.8",
-    "@brightspace-ui/intl": "^3.0.11",
+    "@brightspace-ui/intl": "^3.2.0",
     "@webcomponents/webcomponentsjs": "^2",
     "array-flat-polyfill": "^1.0.1",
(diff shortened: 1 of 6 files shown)


anhill-D2L (approved):

package.json
```diff
@@ -47,7 +47,7 @@
     "@adobe/lit-mobx": "0.0.x",
     "@brightspace-ui-labs/pagination": "^1",
     "@brightspace-ui/core": "^1.102.8",
-    "@brightspace-ui/intl": "^3.0.11",
+    "@brightspace-ui/intl": "^3.2.0",
```
bemailloux, line 50: what happened to 3.1.x? (out of curiosity)

bemailloux (approved):

package.json
johngwilkinson: It existed, we just didn't upgrade to it.

PR #189, opened 2020-12-07
Add Inactive Courses Table

rohitvinnakota:
[US122294](https://rally1.rallydev.com/#/detail/userstory/455008985292?fdp=true): [Engagement][UserDrill] Inactive Courses Table

image

I refactored both the active and inactive courses table to directly take in `Data` so it works with our filters. I think there is room for condensing it into one component but it would be outside the scope of this story. Currently, if there are no active/inactive courses within the filters, the corresponding table disappears but the header remains. After speaking with Kevin, the plan is to add error message dialogs when that is the case(also out of scope). There will be a separate story created for pagination. Currently the tables render all the courses in one page.

**Quality notes**
Tested on edge, firefox, chrome
Changing filters removes or regenerates tables as expected
Responsiveness works as expected
Accessibility: Text picked up by screen reader
Perf testing: n/a
Devops: n/a

- [ ] Add pagination to inactive courses table

fyi: @Vitalii-Misechko @MykolaGalian @NicholasHallman @hyehayes @bemailloux


johngwilkinson:
I just noticed that changing the ou or semester filter while in user view has no effect (because the data is calculated before render instead of being a calculated property on `Data`, I believe). I suspect inactive courses will have the same problem - do you think you could fix both in this PR?

(aside: it seems like the two tables could be instances of the same component with just a flag or two different - not sure, just a thought)

rohitvinnakota:
> I just noticed that changing the ou or semester filter while in user view has no effect (because the data is calculated before render instead of being a calculated property on `Data`, I believe). I suspect inactive courses will have the same problem - do you think you could fix both in this PR?
> 
> (aside: it seems like the two tables could be instances of the same component with just a flag or two different - not sure, just a thought)

I'll take a look

components/active-courses-table.js
```diff
@@ -84,13 +83,13 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 
 	_activeCourses(userRecords) {
 		const orgUnitId =  userRecords.get(RECORD.ORG_UNIT_ID);
-		const orgUnit = this.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
+		const orgUnit = this.data._data.serverData.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
 		return orgUnit[ORG_UNIT.IS_ACTIVE];
 	}
 
 	_preProcessData(userRecords) {
```
johngwilkinson, line 90: is the split between pre-process and format data doing anything here? Looks like the format data method doesn't really do anything.

johngwilkinson, line 90: again, this takes a single record, so the parameter name is confusing

components/active-courses-table.js
```diff
@@ -84,13 +83,13 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 		const orgUnitId =  userRecords.get(RECORD.ORG_UNIT_ID);
-		const orgUnit = this.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
+		const orgUnit = this.data._data.serverData.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
```
johngwilkinson, line 92: again, this is O(n*m)-ish, and a direct lookup of the org unit is better. I think org unit tree might even have a `getName(id)` method.

components/user-drill-view.js
```diff
@@ -182,15 +181,22 @@ class UserDrill extends Localizer(MobxLitElement) {
 				<slot name="filters"></slot>
 			</div>
 
-			<h2 class="d2l-heading-3">${this.localize('activeCoursesTable:title')}</h2>
-
 			<div class="d2l-insights-user-drill-view-content">
+				<!-- put your tables here -->
+				<h2 class="d2l-heading-3">${this.localize('activeCoursesTable:title')}</h2>
```
anhill-D2L, line 186: If you're wanting the `Active/Inactive Courses` titles to disappear along with the tables, you could move these titles into the render functions in the active/inactive-courses-table.js files

components/active-courses-table.js
```diff
@@ -84,13 +83,13 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 
 	_activeCourses(userRecords) {
 		const orgUnitId =  userRecords.get(RECORD.ORG_UNIT_ID);
-		const orgUnit = this.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
+		const orgUnit = this.data._data.serverData.orgUnits.find(x => x[ORG_UNIT.ID] === orgUnitId);
```
johngwilkinson, line 86: Underscore indicates a private member, so referencing `this.data._data` is suspicious, btw - generally that should never happen.

components/active-courses-table.js
```diff
@@ -158,7 +157,8 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		// map to a 2D userData array, with column 1 as a sub-array of [id, FirstName, LastName, UserName]
 		// then sort by the selected sorting function
 		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
-		return this.userCourses
+		const recordsMap = this.data.recordsByUser.get(this.user.userId);
```
johngwilkinson, line 160: naming: does `recordsByUser` return a map or an array? I'm guessing it's an array, based on the array operations below.

components/active-courses-table.js
```diff
@@ -207,18 +207,22 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-			<d2l-insights-table
-				title="${this.localize('activeCoursesTable:title')}"
-				@d2l-insights-table-sort="${this._handleColumnSort}"
-				sort-column="0"
-				.columnInfo=${this.columnInfo}
-				.data="${this._displayData}"
-				?skeleton="${this.skeleton}"
-			></d2l-insights-table>
-		`;
+		if (this.data.recordsByUser.get(this.user.userId) === undefined) {
```
johngwilkinson, line 210: `data.recordsByUser` goes through all the data, so it would be nice to avoid doing this an extra time; could this protective check go somewhere in the getter (or helper methods) for `_displayData` instead?

components/user-drill-view.js
anhill-D2L: > table disappears ...... the plan is to add error message dialogs when that is the case

Is this error message different from the `No data in filtered range` message [here](https://xd.adobe.com/view/2b998c7a-800e-472b-b41d-d2e14658bff7-7eca/screen/6a784cdd-22f9-4368-849e-ef43ea5a3858)

johngwilkinson (commented):
Although I commented on the active courses file, I believe all the same comments apply to the inactive courses one. Which underscores that there is too much duplication happening here, and this should be refactored. That's definitely not out of scope (code quality, including reasonably amounts of refactoring, is always in scope, I'd say 😄 ) - I could see doing it in a separate PR, but I'd suggest doing that PR (separate from this one) _first_: the challenge (I suspect) is just dealing with the slightly different columns, so if you can figure out how to handle that (probably similar to how the users table on the main page handles column selection configs), along with passing in an active/inactive flag (or something to cover that need), the PR to actually add the second table will be pretty tiny.

Also: there should probably be more unit tests for these tables (see users table tests) - and there's no point in writing those tests twice.

(The code for sorting is also getting duplicated here and there, which suggests another refactor so all three tables can share most of their code - but that seems less pressing to me.)

components/user-drill-view.js
rohitvinnakota: Yeah, I think so. It would be referencing the active/inactive tables specifically.

components/inactive-courses-table.js
```diff
@@ -0,0 +1,221 @@
...
+		const sortFunction = this._choseSortFunction(this._sortColumn, this._sortOrder);
+		const recordsMap = this.data.recordsByUser.get(this.user.userId);
+		return recordsMap
+			.filter(this._inactiveCourses, this)
+			.map(this._preProcessData, this)
+			.sort(sortFunction)
+			.map(this._formatDataForDisplay, this);
+	}
+
+	get userDataForDisplayFormatted() {
```
MykolaGalian, line 159: This is an unnecessary function, instead of it you can use `userDataForDisplay`

components/active-courses-table.js
rohitvinnakota: I think in this case, `this.data._data` is the only way to access `orgUnitTree`. Unless there is another way?

johngwilkinson: You could pass in `serverData`. Or make a getter for `orgUnitTree`.

rohitvinnakota:
Creating new PR for combined component

PR #191, opened 2020-12-07
Profile Images for User Drill - US122293

NicholasHallman:
# User Profile Image in the User Drill

## Implementation Notes 

**Profile Image**

**User had no image or token is invalid (initials)** 

**Profile image is loading or bad UserID**

### Functional Testing
  - Unit Test
    - Give the profile component a fake user entity and check if the component reflects the right name and last name
  - Manual Test
    - Check that image appears with proper token and user id
    - check that initials appear with no profile image, with proper token and id
    - check that initials appear with profile image but invalid token, with valid user id
    - check that user placeholder appears while fetching user entity
    - check that user placeholder appears if user id is invalid

### Security
  - Fetches the oauth token using D2L.LP.Oauth, same as is done on the front page.
### UX
  - requires demo to Kevin  

### Rest N/A

FYI @johngwilkinson @Vitalii-Misechko @rohitvinnakota @MykolaGalian @rohitvinnakota


bemailloux:
I approve!... your profile picture 😆

bemailloux (commented):
Do you need any specific LMS permissions to be able to view a user's profile picture? Have you tested with a non-d2lsupport user who has / doesn't have that permission?

components/user-drill-view.js
```diff
@@ -104,13 +110,32 @@ class UserDrill extends Localizer(MobxLitElement) {
...
+	}
+
+	get userProfile() {
+		return until(this.token.then(
+			token => html`
+				<d2l-profile-image
+					class="d2l-insights-user-drill-view-profile-pic"
+					href="${this.userEntity}"
+					token="${token}" x-large>
+				</d2l-profile-image>`), html`<d2l-icon class="d2l-insights-user-drill-view-profile-pic" icon="tier3:profile-pic"></d2l-icon>
```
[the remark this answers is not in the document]
NicholasHallman: yep!

NicholasHallman:
> Do you need any specific LMS permissions to be able to view a user's profile picture? Have you tested with a non-d2lsupport user who has / doesn't have that permission?

I tested this with an admin user enrolled in all courses. I will create a teacher user and get back to you

components/user-drill-view.js
```diff
@@ -55,8 +58,11 @@ class UserDrill extends Localizer(MobxLitElement) {
 			}
 
 			.d2l-insights-user-drill-view-profile-pic {
-				--d2l-icon-height: 100px;
-				--d2l-icon-width: 100px;
+				height: 100px;
+				width: 100px;
```
anhill-D2L, line 62: Does this scale as the page is resized?

NicholasHallman: No! good catch thank you

NicholasHallman:
> Do you need any specific LMS permissions to be able to view a user's profile picture? Have you tested with a non-d2lsupport user who has / doesn't have that permission?



> > Do you need any specific LMS permissions to be able to view a user's profile picture? Have you tested with a non-d2lsupport user who has / doesn't have that permission?
> 
> I tested this with an admin user enrolled in all courses. I will create a teacher user and get back to you

Admin
 - can see user profiles
Instructor
 - can see user profiles ( even if they are a higher role like admin )
Student 
 - can see user profiles ( but why would a student ever need access to the engagement dashboard haha )

package.json
```diff
@@ -57,7 +57,7 @@
     "d2l-navigation": "BrightspaceUI/navigation#semver:^4",
     "d2l-table": "BrightspaceUI/table#semver:^2",
     "d2l-telemetry-browser-client": "Brightspace/d2l-telemetry-browser-client#semver:^1",
-    "d2l-users": "github:BrightspaceHypermediaComponents/users",
+    "d2l-users": "BrightspaceHypermediaComponents/users#semver:^2.4",
```
bemailloux, line 60: You'd probably have to change it to `^2` because that's what BSI has. https://github.com/Brightspace/brightspace-integration/blob/master/package.json#L92

Basically leaving ^2.4 here would mean that if someone publishes 2.5.x, BSI would try to get 2 different versions of d2l-users and break the build

johngwilkinson: `^` matches more recent minor versions, so this should be fine actually

bemailloux: TIL: there is a semver calculator on https://semver.npmjs.com/. `lodash@^1.2` does in fact match 1.3.x as well

bemailloux (approved):
Do you need any special LMS permissions to be able to

bemailloux:
> Do you need any special LMS permissions to be able to

Oops I think this was left over from a previous review lol

anhill-D2L (approved):

PR #192, opened 2020-12-07
fix: semester filter state on reload data from server

johngwilkinson:
Three small changes here:
1. fix handling of null semester filter in server response.
2. add `computed` to some getters in `Data` - not strictly necessary, but good practice, especially for the ones with corresponding setters (makes those into `action`s).
3. add `persistEmpty` to the semester filter so we can distinguish the default state from an all-filters-cleared state

Quality Notes
- functional (on local LMS)
  - load default, check url
  - clear all filters, check url
  - clear all filters, refresh page (no default view popup or filters)
  - clear query string and load (default view popup and filters)
  - back and forward buttons with various filter settings including cleared semester filter
- devops/security/perf/ux: n/a

FYI @bemailloux @MykolaGalian @rohitvinnakota @Vitalii-Misechko

### DIFF
--- a/model/urlState.js
+++ b/model/urlState.js
@@ -40,6 +40,8 @@ export function clearUrlState() {
  * set persistenceValue({String}) - called when the component should set the given state
  * If persistenceValue references one or more mobx observables, mobx will track changes. Otherwise,
  * it is necessary to call save() when the state changes.
+ * The wrapped object may optionally implement:
+ * {Boolean} get persistEmpty() - if true, the key will not be removed from the url even if the value is ''
  */
 export class UrlState {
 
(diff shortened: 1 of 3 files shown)


model/urlState.js
```diff
@@ -60,7 +62,7 @@ export class UrlState {
 	save() {
 		const url = new URL(window.location.href);
 		const valueToSave = this.value;
-		if (valueToSave === '' && url.searchParams.has(this.key)) {
+		if (!this.persistEmpty && valueToSave === '' && url.searchParams.has(this.key)) {
```
NicholasHallman, line 65: Nice!

NicholasHallman (approved):

anhill-D2L (approved):

PR #193, opened 2020-12-07
Fix: Applied Filters isn't Keyboard Accessible

NicholasHallman:
This is one way we can fix it, but it may be better if we submit a pr directly to the component. Thoughts?

https://rally1.rallydev.com/#/23525809118d/workviews?detail=%2Fdefect%2F459133385892&fdp=true&view=349baf7a-c63a-4fa4-a43b-2bb09274aad7

### DIFF
--- a/components/applied-filters.js
+++ b/components/applied-filters.js
@@ -28,7 +28,10 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	_handleKeyEvents(e) {
-		e.path[0].shadowRoot.querySelector('d2l-icon').click();
+		if (e.keyCode === 32 || e.keyCode === 13) {
+			e.preventDefault();
+			e.path[0].shadowRoot.querySelector('d2l-icon').click();
+		}
 	}
 
 	render() {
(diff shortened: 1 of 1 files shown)


johngwilkinson:
General question: why are we using this component at all? It also creates a "bug" where if there are more than N filters, it adds an unwanted "clear all". I'm wondering if it's the right component for our design.

NicholasHallman:
> General question: why are we using this component at all? It also creates a "bug" where if there are more than N filters, it adds an unwanted "clear all". I'm wondering if it's the right component for our design.

I think this is what we should be using https://github.com/BrightspaceUILabs/multi-select

johngwilkinson:
> > General question: why are we using this component at all? It also creates a "bug" where if there are more than N filters, it adds an unwanted "clear all". I'm wondering if it's the right component for our design.
> 
> I think this is what we should be using https://github.com/BrightspaceUILabs/multi-select
> 
> EDIT: It has the same axe problem... nvm

These look like a handful of buttons, and are supposed to behave much like that - so could we just draw them as `d2l-button`?

Not that I'm set against Bryan's fix, which is very quick and simple.

Vitalii-Misechko (approved):

anhill-D2L (approved):

PR #195, opened 2020-12-08
Us121775/hide active courses

anhill-D2L:
Functional:
Tested on the demo page and my local LMS
-when there are some active course(s) the title and table are 
-when there are no active courses, the title, and table are both not shown

Accessibility: 
-NVDA didn't mention anything related to active courses when they were hidden

Ux: 
-double checked with PM that we want to be hiding the active table/title if there are no active courses (and similarly for inactive coruses)

Security: n/a
Devops: n/a

FYI: @Vitalii-Misechko @MykolaGalian @hyehayes @bemailloux @NicholasHallman


anhill-D2L:
This is covered in https://github.com/Brightspace/insights-engagement-dashboard/pull/189/files

PR #199, opened 2020-12-09
Export from user table

MykolaGalian:
[US121804](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F449519683764&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement] Export from user table

Local

LMS

In progress: unit tests
PR depending from "Add active and inactive courses table" PR

### DIFF
--- a/components/active-courses-table.js
+++ b/components/active-courses-table.js
@@ -18,6 +18,8 @@ export const TABLE_COURSES = {
 	COURSE_LAST_ACCESS: 5
 };
 
+export const IS_ACTIVE_COURSE = 6;
+
 const numberFormatOptions = { maximumFractionDigits: 2 };
 
 const DEFAULT_PAGE_SIZE = 20;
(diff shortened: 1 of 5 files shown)


components/active-courses-table.js
```diff
@@ -88,6 +90,16 @@ class CoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return orgUnit[ORG_UNIT.IS_ACTIVE];
 	}
 
+	_activeCourseById(id) {
+		const orgUnit = this.orgUnits.find(x => x[ORG_UNIT.ID] === id);
```
johngwilkinson, line 94: You might want to work off of @rohitvinnakota's branch, as some of this will have to change (no `this.orgUnits` anymore, for one). If he is refactoring, that will change things, too, so you might need to coordinate.

PR #211, opened 2020-12-14
User Drill Active Filters - US123291

NicholasHallman:
Added the applied filters to the user drill page using a slot. This component shares the code used in the main view.

Quality Notes

Functional Testing
  - In the demo, selected the 5 - 7 course access filter, selected student B from the user table (only one course shown) removed the filter ( now 3 courses shown ), exited from the users view, verified that the cleared filter is also cleared in the main view.
  - In the demo, select any filter, select any student from the user table, return to the main view (using the header back button) verified that the filter remained applied. 

Performance - N/A  

Security - N/A  

Dev-Ops - N/A

Cost - N/A

Documentation - N/A

UX
Demoed to Kevin, received 👍 

5YZlEkjyTV


components/user-drill-view.js
```diff
@@ -184,17 +184,21 @@ class UserDrill extends Localizer(MobxLitElement) {
...
-
+				${ this.userCourses ? html`
 				<d2l-insights-active-courses-table
 					.userCourses="${this.userCourses}"
 					.orgUnits="${this.orgUnits}"
 					.isStudentSuccessSys="${this.isStudentSuccessSys}"
 				>
 				</d2l-insights-active-courses-table>
-
+				` : ''}
```
bemailloux, line 201: `Just a thought` I wonder if there's a difference between `''` and ```html`` ```

bemailloux (commented):
Does it also work if no (card) filters are applied?

components/user-drill-view.js
NicholasHallman: I don't think so... because this function renders into an html`` template literal the empty string would appear the same, it's probably just faster since it doesn't have to call the html templating function.

If it was being called by the render() lit-html function or the lit-element render function then yes? I think it would make a difference

NicholasHallman:
> Does it also work if no (card) filters are applied?

Yep! No filters returns all records (doesn't apply any filters)

bemailloux (approved):

components/user-drill-view.js
Vitalii-Misechko: Just came across [recently](https://lit-html.polymer-project.org/guide/writing-templates#rendering-nothing)
> Sometimes, you may want to render nothing at all. The values **undefined**, **null** and the **empty string** ('') in a text binding all render an empty text node. In most cases, that's exactly what you want:

In order to render nothing you should use directive `nothing` )

Vitalii-Misechko (approved):

components/user-drill-view.js
NicholasHallman: Very cool!

PR #220, opened 2020-12-16
Add semester column to inactive courses table

MykolaGalian:
[US122803](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F459588582524&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement] Add semester column to inactive courses table

Export
Local (isStudentSuccessSys = false)
false1

Local (isStudentSuccessSys = true)
true1

LMS
lms1

Functional:
Local App and LMS work as expected with test data.

Accessibility: n/a
Security: n/a
Devops: n/a
UI: demo for Kevin - done

FYI @rohitvinnakota @NicholasHallman @Vitalii-Misechko @anhill-D2L


components/user-drill-courses-table.js
```diff
@@ -97,7 +98,7 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		if (this._itemsCount) {
 			const start = this._pageSize * (this._currentPage - 1);
 			const end = this._pageSize * (this._currentPage); // it's ok if this goes over the end of the array
-			const numCols = this.isActiveTable && this.isStudentSuccessSys ? Array.from(Array(6), (_, x) => x) : Array.from(Array(5), (_, x) => x);
+			const numCols = this.isActiveTable && this.isStudentSuccessSys ? Array.from(Array(6), (_, x) => x) : Array.from(Array(6), (_, x) => x);
```
johngwilkinson, line 101: I thought we got rid of this line in another code review. Anyway, it was odd before, and makes no sense now: it's a conditional with the same value on both branches! And it's used below to map an array to itself, which doesn't appear to serve any purpose.

components/user-drill-courses-table.js
```diff
@@ -143,11 +145,20 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-				lastCourseAccess
+				lastCourseAccess,
+				semesterOrgUnitName
 			];
 		}
 	}
 
+	_getSemesterNameByOrgUnitId(orgUnitId) {
+		const ancestors = toJS(this.data.orgUnitTree.getAncestorIds(orgUnitId));
+		const semesterTypeId =  this.data._data.serverData.semesterTypeId;
```
johngwilkinson, line 156: `data._data` is private; you can put a getter for `semesterTypeId` on `data` instead.

components/user-drill-courses-table.js
```diff
@@ -143,11 +145,20 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const firstSemesterOrgUnitId = ancestors.find(id => this.data.orgUnitTree.getType(id) === semesterTypeId);
+		const semesterOrgUnitName = this.data.orgUnitTree.getName(firstSemesterOrgUnitId);
+		return this.localize('semesterFilter:semesterName', { orgUnitName: semesterOrgUnitName, orgUnitId: firstSemesterOrgUnitId });
```
johngwilkinson, line 159: This looks right, except both these values could be null or undefined - not sure what we should display in that case.

components/user-drill-courses-table.js
```diff
@@ -143,11 +145,20 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 				[threads, reads, replies],
-				lastCourseAccess
+				lastCourseAccess,
+				semesterOrgUnitName
 			];
 		}
 	}
 
+	_getSemesterNameByOrgUnitId(orgUnitId) {
+		const ancestors = toJS(this.data.orgUnitTree.getAncestorIds(orgUnitId));
```
johngwilkinson, line 155: I don't think `toJs` is needed here, is it?

MykolaGalian: I think so because the value is an observable Set

johngwilkinson: Looking at the docs for toJs, I don't see that we should be using it. I think what's happening here is that `getAncestorIds` returns an es6 `Set` (yes, it's observable, but that shouldn't matter), and you are using `toJs` to convert it to an array so you can do `find` on it. Please try using `[...this.data.orgUnitTree.getAncestorIds(orgUnitId)]` or `this.data.orgUnitTree.getAncestorIds(orgUnitId).values()` instead.

components/user-drill-courses-table.js
```diff
@@ -97,10 +99,16 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		if (this._itemsCount) {
 			const start = this._pageSize * (this._currentPage - 1);
 			const end = this._pageSize * (this._currentPage); // it's ok if this goes over the end of the array
-			const numCols = this.isActiveTable && this.isStudentSuccessSys ? Array.from(Array(6), (_, x) => x) : Array.from(Array(5), (_, x) => x);
-			return this.userDataForDisplay
-				.slice(start, end)
-				.map(user => numCols.map(column => user[column]));
+
+			const dataForDisplay = JSON.parse(JSON.stringify(this.userDataForDisplay));
```
johngwilkinson, line 103: Why is a deep copy needed here?

components/user-drill-courses-table.js
```diff
@@ -97,10 +99,16 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			if (this.isActiveTable) {
+				return dataForDisplay.map(data => {
+					data.pop();  // removed semester name column
```
johngwilkinson, line 106: `slice()` is better than `pop()` here. Notice that `slice()` makes a shallow copy instead of modifying the original. There's functional code:
```
return dataForDisplay.map(x => x.slice(0, lastcol))
...
```
and code with side-effects:
```
dataForDisplay.forEach(x => x.pop());
return dataForDisplay
...
```
It's best not to mix them (`pop` inside `map`), and generally functional is cleaner and safer (might eliminate the need for the deep copy just above).

components/user-drill-courses-table.js
MykolaGalian: Removed

johngwilkinson (commented):
Looks good. I'll approve after the release has been tagged.

johngwilkinson:
Probably need to merge in master then get the new visual diff.

johngwilkinson:
There are some merge conflicts from my last PR, which changed the column logic a little.

components/user-drill-courses-table.js
```diff
@@ -123,17 +132,27 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			semesterOrgUnitName
 		];
 	}
 
+	_getSemesterNameByOrgUnitId(orgUnitId) {
+		const ancestors = [...this.data.orgUnitTree.getAncestorIds(orgUnitId)];
+		const semesterTypeId =  this.data.semesterTypeId;
+		const firstSemesterOrgUnitId = ancestors.find(id => this.data.orgUnitTree.getType(id) === semesterTypeId);
+		const semesterOrgUnitName = this.data.orgUnitTree.getName(firstSemesterOrgUnitId);
+		return firstSemesterOrgUnitId !== undefined ? this.localize('semesterFilter:semesterName', { orgUnitName: semesterOrgUnitName, orgUnitId: firstSemesterOrgUnitId }) : '';
```
bemailloux, line 153: It looks like this can be simplifed to `return firstSemesterOrgUnitId ? this.localize(...) : ''`

test/components/user-drill-courses-table.test.js
```diff
@@ -59,11 +62,11 @@ const expected = {
...
-		['Course 11 (Id: 11)', '80 %', '110', [3, 3, 3], getLocalDateTime(1607979700000)],
-		['Course 12 (Id: 12)', '85 %', '90', [2, 2, 2], 'NULL'],
-		['Course 13 (Id: 13)', '75 %', '70', [5, 5, 5], getLocalDateTime(1607979820000)],
-		['Course 14 (Id: 14)', 'No grade', '80', [7, 7, 7], getLocalDateTime(1607979640000)],
-		['Course 15 (Id: 15)', '88 %', '120', [4, 4, 4], getLocalDateTime(1607979880000)],
+		['Course 11 (Id: 11)', '80 %', '110', [3, 3, 3], getLocalDateTime(1607979700000), 'Course 100 (Id: 100)'],
+		['Course 12 (Id: 12)', '85 %', '90', [2, 2, 2], 'NULL', 'Course 100 (Id: 100)'],
+		['Course 13 (Id: 13)', '75 %', '70', [5, 5, 5], getLocalDateTime(1607979820000), 'Course 100 (Id: 100)'],
+		['Course 14 (Id: 14)', 'No grade', '80', [7, 7, 7], getLocalDateTime(1607979640000), 'Course 100 (Id: 100)'],
+		['Course 15 (Id: 15)', '88 %', '120', [4, 4, 4], getLocalDateTime(1607979880000), 'Course 100 (Id: 100)'],
```
bemailloux, line 69: For the semesters it would be nice if the mock org unit name was "Semester" instead of "Course"

bemailloux (approved):

components/user-drill-courses-table.js
```diff
@@ -92,8 +93,16 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 		if (this._itemsCount) {
 			const start = this._pageSize * (this._currentPage - 1);
 			const end = this._pageSize * (this._currentPage); // it's ok if this goes over the end of the array
+
 			const visibleColumns = this._visibleColumns;
-			return this.userDataForDisplay
+			const dataForDisplay = this.userDataForDisplay;
+			if (this.isActiveTable) {
+				const columnsWithoutSemester = visibleColumns.slice(0, visibleColumns.length - 1);
```
johngwilkinson, line 100: This works so long as there is a semester column and it is last, but another column might get added in future, and we might add configuration to hide the semester.

I believe a better solution is to filter in `get _visibleColumns` - something like `return this.isActiveTable ? columns.filter(x => x !== TABLE_COLUMNS.SEMESTER_NAME) : columns;`. This should eliminate the need to make any changes in `get _displayData()`, and also ensures that sorting works if other columns are added in future.

Vitalii-Misechko (approved):

PR #222, opened 2020-12-16
User Drill Skeleton US122721

NicholasHallman:
Adds the skeleton mixin to the user drill and the profile image. Creates a promise chain for loading the data provided by the engagement-dashboard-component and the profile entity (which is a separate fetch) and waits for this promise chain to resolve. 

The skeleton is most visible when you load right into a user drill from a bookmark, when clicking from the engagement dashboard to the user-drill all data except for the user profile will already be available and the page will load almost instantly.

Functional Testing
  - Extending the fake lms timeout and loading right into a user drill shows the skeleton
  - All components update after skeleton period is over
  - Filtering tables works after skeleton period.

Performance - N/A

Security - N/A

Dev-Ops - N/A

Cost - N/A

Documentation - N/A

UX
 - Demoed to Kevin received 👍 
 
 
skeleton

 
FYI @bemailloux @hyehayes @johngwilkinson @anhill-D2L @MykolaGalian


bemailloux:
We're going to conflict a tiny bit so I'll rebase on top of your branch

components/user-drill-view.js
```diff
@@ -126,35 +142,62 @@ class UserDrill extends Localizer(MobxLitElement) {
...
+		return this.skeleton || !this.hasToken;
+	}
+
+	get skeletonClass() {
+		return this.isLoading ? 'd2l-skeletize' : '';
+	}
+
 	get token() {
+		// use the cached token so we don't call for it everytime the ui reloads
+		if (this._token) return this._token;
```
bemailloux, line 155: IIRC tokens expire after 30 min - might want to add the ability to get a new token if trying to use it resulted in 401 or 403

components/user-drill-courses-table.js
```diff
@@ -205,6 +205,7 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 	get userDataForDisplay() {
 		// map to a 2D userData array, with column 1 as a sub-array of [id, FirstName, LastName, UserName]
 		// then sort by the selected sorting function
+		if (this.skeleton) return [];
```
Vitalii-Misechko, line 208: nice guard condition!

engagement-dashboard.js
```diff
@@ -230,24 +230,28 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 	}
 
 	_renderUserDrillView() {
-		const userId = this._viewState.userViewUserId;
-		const userData = this._serverData.userDictionary.get(userId);
-
-		if (!userData) {
-			return;
+		let user = {};
+		if (!this._isLoading) {
```
Vitalii-Misechko, line 234: `nit` this can be converted to a guard condition
```
if (this._isLoading) return user;
```

components/user-drill-view.js
```diff
@@ -215,4 +259,10 @@ class UserDrill extends Localizer(MobxLitElement) {
 	}
 }
 
+decorate(UserDrill, {
+	hasToken: observable,
+	skeleton: observable,
+	isLoading: computed,
+});
```
Vitalii-Misechko, line 266: `thought` All these properties affects rendering. I believe we could declare them as UserDrill's properties without any change in the component's behavior. My current approach is to avoid mixing MobX with Lit-element if this can be avoided. These technologies are new to me though.

Vitalii-Misechko (approved):
LGTM!

engagement-dashboard.js
NicholasHallman: It's a little different, This function doesn't return the user it returns the user drill view template. We want to pass an empty user into the drill if the user hasn't loaded yet (for example when you enter a user drill from a bookmarked url)

components/user-drill-view.js
NicholasHallman: After experimenting it looks like it doesn't need these decorators. Good catch thank you

components/user-drill-courses-table.js
NicholasHallman: It's a hack lol, I needed the skeleton observable to be in the function for it to re calculate after loading finished. In reality the guard is never executed since there's an earlier guard that prevents this function from being run when loading

rohitvinnakota (approved):

components/user-drill-view.js
Vitalii-Misechko: Interesting. Only `skeleton` is declared as a property `hasToken` is not UserDrill's property. Therefore, `token` getter should not cause any rendering after promise resolving (or I miss something :) )

Vitalii-Misechko (approved):

PR #225, opened 2020-12-16
Settings page - error when saving fails(Last System Access threshold value)

rohitvinnakota:
Adds a d2l-alert-toast message when trying to save and close an invalid set of settings. Currently, the message works for the last system access threshold value case but the functionality can be extended.

[US122868](https://rally1.rallydev.com/#/23525809118d/workviews/collection?detail=%2Fuserstory%2F459991179776%2Ftasks&itemRef=%2Fhierarchicalrequirement%2F459991179776&previousTokens=%2F23525809118d%2Fcustom%2F12155517225%3F&project=%2Fproject%2F23525809118&ref=%2Fhierarchicalrequirement%2F459991179776&relationship=tasks&fdp=true?fdp=true): [Engagement][Config] Settings page - error when saving fails

image
image

(The value in the screenshot's input is 233, not 23)

Screen reader: Clicking Save and close on an invalid system access value immediately triggers the alert and reads it out.
Responsiveness: Alert centers around screen when window is resized. 
Tested on edge, chrome, firefox

Security: n/a
Devops: n/a

- [x] Demo to Kevin

fyi: @johngwilkinson @hyehayes @NicholasHallman @anhill-D2L @MykolaGalian

### DIFF
--- a/components/card-selection-list.js
+++ b/components/card-selection-list.js
@@ -235,11 +235,11 @@ class CardSelectionList extends RtlMixin(Localizer(LitElement)) {
 
 	_handleThresholdFieldChange(event) {
 		const newValue = Number(event.target.value);
-		if (isNaN(newValue) || newValue < lastSysAccessThresholdMinDays || newValue > lastSysAccessThresholdMaxDays) {
-			return;
-		}
-
 		this.lastAccessThresholdDays = Math.floor(newValue);
+	}
+
+	isInvalidSystemAccessValue() {
+		return (isNaN(this.lastAccessThresholdDays) || this.lastAccessThresholdDays < lastSysAccessThresholdMinDays || this.lastAccessThresholdDays > lastSysAccessThresholdMaxDays);
 	}
 
 	get settings() {
--- a/components/custom-toast-message.js
+++ b/components/custom-toast-message.js
@@ -5,21 +5,22 @@ import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
 
 class CustomToastMessage extends RtlMixin(Localizer(LitElement)) {
 
+	static get properties() {
+		return {
+			toastMessageText: { type: String, attribute: false },
+		};
+	}
+
 	render() {
 		return html`
-			<d2l-alert-toast type="critical"></d2l-alert-toast>
+			<d2l-alert-toast type="critical" subtext="${this.toastMessageText}"></d2l-alert-toast>
 		`;
 	}
 
-	systemLastAccessError() {
-		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:invalidSystemAccessValueToast');
+	open() {
 		this.shadowRoot.querySelector('d2l-alert-toast').open = true;
 	}
 
-	failedServerResponseError() {
-		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:serverSideErrorToast');
-		this.shadowRoot.querySelector('d2l-alert-toast').open = true;
-	}
 }
 
 customElements.define('d2l-insights-custom-toast-message', CustomToastMessage);
--- a/components/custom-toast-message.js
+++ b/components/custom-toast-message.js
@@ -15,6 +15,11 @@ class CustomToastMessage extends RtlMixin(Localizer(LitElement)) {
 		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:invalidSystemAccessValueToast');
 		this.shadowRoot.querySelector('d2l-alert-toast').open = true;
 	}
+
+	failedServerResponseError() {
+		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:serverSideErrorToast');
+		this.shadowRoot.querySelector('d2l-alert-toast').open = true;
+	}
 }
 
 customElements.define('d2l-insights-custom-toast-message', CustomToastMessage);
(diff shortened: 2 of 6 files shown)


components/custom-toast-message.js
```diff
@@ -0,0 +1,20 @@
...
+		`;
+	}
+
+	systemLastAccessError() {
+		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:invalidSystemAccessValueToast');
+		this.shadowRoot.querySelector('d2l-alert-toast').open = true;
+	}
+}
+
+customElements.define('d2l-insights-custom-toast-message', CustomToastMessage);
```
Vitalii-Misechko, line 20: `thought` This component does not provide any additional behavior on top of `d2l-alert-toast`. Should we delete it?

components/dashboard-settings.js
```diff
@@ -249,6 +250,11 @@ class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
 			includeRoles: this._selectedRoleIds
 		};
 
+		if (cardSelectionList.isInvalidSystemAccessValue()) {
+			this.shadowRoot.querySelector('d2l-insights-custom-toast-message').systemLastAccessError();
+			return;
```
Vitalii-Misechko, line 255: Have you considered introducing `open` property into `d2l-insights-custom-toast-message` (and `_saveError` into this class)? This way you could
```JS
<d2l-insights-custom-toast-message ?open="${this._saveError}">
...
if (cardSelectionList.isInvalidSystemAccessValue()) {
  this._saveError= true
  return
}
```
I think this way is more natural for the Lit-Element framework.

test/components/custom-toast-message.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+			const el = await fixture(html`<d2l-insights-custom-toast-message></d2l-insights-custom-toast-message>`);
+			await expect(el).to.be.accessible();
+		});
+	});
+
+	describe('render', () => {
+		it('should render the invalid system last accessed value toast message as expected', async() => {
+			const el = await fixture(html`<d2l-insights-custom-toast-message></d2l-insights-custom-toast-message>`);
+			await new Promise(resolve => setTimeout(resolve, 200)); // allow fetch to run
```
Vitalii-Misechko, line 24: `nit` wrong comment

test/components/custom-toast-message.test.js
```diff
@@ -0,0 +1,33 @@
...
+			expect(el.shadowRoot.querySelector('d2l-alert-toast').innerHTML).to.deep.equal('');
```
Vitalii-Misechko, line 25: `nit` should it be `.to.equal` instead?

components/custom-toast-message.js
rohitvinnakota: My thinking was we can add additional toast messages as needed as functions and reuse the same component to conditionally render those messages; essentially calling those functions to trigger the toast message.

Vitalii-Misechko: sounds reasonable

Vitalii-Misechko (approved):

components/card-selection-list.js
```diff
@@ -235,13 +235,13 @@ class CardSelectionList extends RtlMixin(Localizer(LitElement)) {
 
 	_handleThresholdFieldChange(event) {
 		const newValue = Number(event.target.value);
-		if (isNaN(newValue) || newValue < lastSysAccessThresholdMinDays || newValue > lastSysAccessThresholdMaxDays) {
-			return;
-		}
-
 		this.lastAccessThresholdDays = Math.floor(newValue);
```
bemailloux, line 238: So what happens if there's an invalid input and the value is NaN? Does the input field show NaN?

(I think that's fine, but I also think the convention is for an input field to correct itself if there is an invalid input, which is something you've removed here.)

bemailloux (commented):
Is there a corresponding server-side validation PR?

Also, what happens if we tried to save server-side and that failed? it doesn't look like there's error handling for that case.

components/card-selection-list.js
rohitvinnakota: When you click out, it makes the input field blank. I think it was the same functionality before.

rohitvinnakota:
> Is there a corresponding server-side validation PR?
> 
> Also, what happens if we tried to save server-side and that failed? it doesn't look like there's error handling for that case.

Hmm yeah. I could add a quick check to return a failed response if we pass in an invalid value, but that should not happen with the client side validation.

johngwilkinson:
> > Is there a corresponding server-side validation PR?
> > Also, what happens if we tried to save server-side and that failed? it doesn't look like there's error handling for that case.
> 
> Hmm yeah. I could add a quick check to return a failed response if we pass in an invalid value, but that should not happen with the client side validation.

We do need to show an error if the server returns 404 or 500 or the like, too - let's say the user's internet connection has dropped.

components/custom-toast-message.js
```diff
@@ -0,0 +1,25 @@
...
+class CustomToastMessage extends RtlMixin(Localizer(LitElement)) {
+
+	render() {
+		return html`
+			<d2l-alert-toast type="critical"></d2l-alert-toast>
+		`;
+	}
+
+	systemLastAccessError() {
+		this.shadowRoot.querySelector('d2l-alert-toast').innerHTML = this.localize('settings:invalidSystemAccessValueToast');
```
bemailloux, line 15: You shouldn't use innerHTML to insert text like this: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#Security_considerations

Instead `text` and `open` can just become properties in this class (basically what Vitalii said).

rohitvinnakota: I added `text ` as a prop, but added a function for open since that way, the message renders more than once.

bemailloux (approved):

PR #226, opened 2020-12-16
Small fixes

bemailloux:
A couple of little things I noticed
- print button wasn't supposed to exist yet
- in user drill name should be `lastName, firstName`. It was `first, last`

FYI @rohitvinnakota @MykolaGalian @anhill-D2L @johngwilkinson


bemailloux:
So this is unfortunate. If we don't hardcode the course access timestamps, then visual diff fails every time. But if we do hardcode them, then visual-diff for the course access card will fail occasionally.

The solution is probably to pass a "current time" parameter to the course access card. I won't implement that here.

Vitalii-Misechko:
> in user drill name should be lastName, firstName. It was first, last

Could you confirm this with Kevin? LMS profile shows `first, last` and [design](https://xd.adobe.com/view/2b998c7a-800e-472b-b41d-d2e14658bff7-7eca/) form [the story](https://rally1.rallydev.com/#/detail/userstory/449082079828?fdp=true) specifies this order as well

Vitalii-Misechko (approved):

NicholasHallman:
> So this is unfortunate. If we don't hardcode the course access timestamps, then visual diff fails every time. But if we do hardcode them, then visual-diff for the course access card will fail occasionally.
> 
> image
> 
> The solution is probably to pass a "current time" parameter to the course access card. I won't implement that here.

We can also hard code the days in fake-lms to > 14 days away so they stop drifting up the chart

NicholasHallman (approved):

bemailloux:
> We can also hard code the days in fake-lms to > 14 days away so they stop drifting up the chart

We could do that but it would reduce the effectiveness of the test slightly.

bemailloux:
> Could you confirm this with Kevin? LMS profile shows `first, last` and [design](https://xd.adobe.com/view/2b998c7a-800e-472b-b41d-d2e14658bff7-7eca/) form [the story](https://rally1.rallydev.com/#/detail/userstory/449082079828?fdp=true) specifies this order as well

Changed from `First, Last` to `First Last`

Vitalii-Misechko (approved):

NicholasHallman:
Another possible solution to the drifting vis-diffs is to override Date.now() so it always returns a fixed date in the demo mode

johngwilkinson:
> Another possible solution to the drifting vis-diffs is to override Date.now() so it always returns a fixed date in the demo mode

https://sinonjs.org/releases/v1.17.6/fake-timers/ possibly? Haven't tried it in this sort of test before.

PR #229, opened 2020-12-17
Us122289+DE41776 user drill table errors

bemailloux:
Shows error messages for various possible states.

## Quality notes
* Testing (in progress)
  * [x] Unit tests for new code 
  * [x] Responsive
  * [x] Accessibility
  * [x] Test cases
    * (Regression) It shows skeletons when loading
    * (Regression) Both tables show data if there is data to show
    * When user was not loaded due to invalid user id, displays "return to engagement dashboard" message
    * When user was not loaded due to lack of permissions, displays "return to engagement dashboard" message
    * When user was not loaded due to reload after specific user + filter combo, displays "return to engagement dashboard" message
    * Return to engagement dashboard button ~~returns to home view~~ hard resets to default view
    * When filters were applied such that no courses are in view, it hides both tables and displays "No data" message
    * When filters were applied such that no active courses are in view, it displays "no active courses" message; inactive still displays data
    * When filters were applied such that no inactive courses are in view, it displays "no inactive courses" message; active still displays data
* Security, docs, perf, devops: N/A
* UX:
  * [ ] Demo to Kevin and/or Megan

## Screenshots
image
image
image
image

FYI @johngwilkinson @Vitalii-Misechko @MykolaGalian @anhill-D2L

### DIFF
--- a/components/user-drill-view.js
+++ b/components/user-drill-view.js
@@ -1,13 +1,18 @@
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/button/button.js';
-import './user-drill-courses-table.js';
 import 'd2l-users/components/d2l-profile-image';
+
+import './user-drill-courses-table.js';
+import './message-container';
+
 import { bodySmallStyles, heading2Styles, heading3Styles } from '@brightspace-ui/core/components/typography/styles.js';
+import { computed, decorate } from 'mobx';
 import { css, html } from 'lit-element/lit-element.js';
 import { createComposeEmailPopup } from './email-integration';
 import { ExportData } from '../model/exportData';
 import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
+import { RECORD } from '../consts';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 import { until } from 'lit-html/directives/until';
 
@@ -20,7 +25,7 @@ import { until } from 'lit-html/directives/until';
 class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 	static get properties() {
 		return {
-			data: { type: Object, attribute: {} },
+			data: { type: Object, attribute: false },
 			user: { type: Object, attribute: false },
 			isDemo: { type: Boolean, attribute: 'demo' },
 			isStudentSuccessSys: { type: Boolean, attribute: false },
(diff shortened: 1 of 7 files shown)


johngwilkinson (approved):

components/user-drill-view.js
```diff
@@ -183,15 +200,33 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 		);
 	}
 
+	// @computed
+	get _userRecords() {
+		return this.data.records.filter(r => r[RECORD.USER_ID] === this.user.userId);
+	}
+
 	render() {
+		if (!this.skeleton && !this.user.userId) {
```
bemailloux, line 209: I tried to pass in `null` to this component, but apparently lit-analyzer doesn't like it if you pass `null` in as an object. I think that's ridiculous but couldn't find a way around that error, so I've had to compromise by checking the userId instead.

NicholasHallman: I think undefined works, it treats it like the the property was excluded

NicholasHallman (approved):

components/user-drill-view.js
bemailloux: I tried that as well. I think it would have worked if the property was also reflected in an attribute but it's not, so lit-analyzer complains (whether or not I use `ifDefined` too)

bemailloux:
Just waiting for some possibly updated wording from Kevin

johngwilkinson:
You're getting the same sauce errors as me. Stuff like
```
[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[LastAccessCard.update()]'', TypeError: Cannot read property 'filter' of undefined
TypeError: Cannot read property 'filter' of undefined
    at HTMLElement.get _cardValue (http://localhost:9876/base/components/last-access-card.js:87:4)
```

johngwilkinson:
> You're getting the same sauce errors as me. Stuff like
> 
> ```
> [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[LastAccessCard.update()]'', TypeError: Cannot read property 'filter' of undefined
> TypeError: Cannot read property 'filter' of undefined
>     at HTMLElement.get _cardValue (http://localhost:9876/base/components/last-access-card.js:87:4)
> ```

Actually, I think these errors might not be causing the build to fail - might just be noise. Would be worth cleaning up if so (after the milestone is done).

bemailloux:
> Actually, I think these errors might not be causing the build to fail - might just be noise. Would be worth cleaning up if so (after the milestone is done).

I've noticed that Safari in particular seems to be crashing more than usual. Maybe related, or maybe not...?

johngwilkinson:
> > Actually, I think these errors might not be causing the build to fail - might just be noise. Would be worth cleaning up if so (after the milestone is done).
> 
> I've noticed that Safari in particular seems to be crashing more than usual. Maybe related, or maybe not...?

I found a bunch of those errors in my local karma run too. Could be making it flakier, but I suspect it's unrelated.

rohitvinnakota (approved):

PR #231, opened 2020-12-17
US123012: off-by-one error when sorting user-drill courses table

johngwilkinson:
Quality Notes
Before: on an org with no predicted grades, sorting on time-in-content did nothing because it was looking at current grade (one column over) instead (which happened to be null in my test data).
After: sorting works on all columns
Before: export always included predicted grade (even if not enabled)
After: export doesn't include predicted grade if not enabled (PM said either was fine, but this was equally easy along with the other changes)
Still testing:
- [x] render and sort of inactive courses
- [x] render, sort, and export when predicted grades are enabled

### DIFF
--- a/components/user-drill-courses-table.js
+++ b/components/user-drill-courses-table.js
@@ -10,21 +10,13 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { RECORD } from '../consts';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
-export const ACTIVE_TABLE_COLUMNS = {
+const TABLE_COLUMNS = {
 	COURSE_NAME: 0,
 	CURRENT_GRADE: 1,
 	PREDICTED_GRADE: 2,
 	TIME_IN_CONTENT: 3,
 	DISCUSSION_ACTIVITY: 4,
 	COURSE_LAST_ACCESS: 5
-};
-
-export const INACTIVE_TABLE_COLUMNS = {
-	COURSE_NAME: 0,
-	CURRENT_GRADE: 1,
-	TIME_IN_CONTENT: 2,
-	DISCUSSION_ACTIVITY: 3,
-	COURSE_LAST_ACCESS: 4
 };
 
 const numberFormatOptions = { maximumFractionDigits: 2 };
@@ -70,7 +62,7 @@ class UserDrillCoursesTable extends SkeletonMixin(Localizer(MobxLitElement)) {
 	constructor() {
 		super();
 		this._sortOrder = 'desc';
-		this._sortColumn = ACTIVE_TABLE_COLUMNS.COURSE_NAME;
+		this._sortColumn = TABLE_COLUMNS.COURSE_NAME;
 		this._currentPage = 1;
 		this._pageSize = DEFAULT_PAGE_SIZE;
 	}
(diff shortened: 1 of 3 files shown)


NicholasHallman (approved):
Looks good, and much cleaner thanks!

test/components/user-drill-courses-table.test.js
```diff
@@ -359,7 +359,8 @@ describe('d2l-insights-user-drill-courses-table', () => {
 				await new Promise(resolve => setTimeout(resolve, 200));
 				await el.updateComplete;
 				expect(el.dataForExport[0]).to.deep.equal(
-					[ 'Course 101 (Id: 101)', '80 %', 'No predicted grade', '110', 3, 3, 3, getLocalDateTime(1607979700000), true]);
+					[ 'Course 101 (Id: 101)', '80 %', '110', 3, 3, 3, getLocalDateTime(1607979700000), true]);
+				expect(el.headersForExport).to.deep.equal(['Course Name', 'Grade', 'Time in Content (mins)', 'Threads', 'Reads', 'Replies', 'Course Last Access', 'Is Active Course']);
```
bemailloux, line 363: Oops, I had written a unit test for the [`sorting / isS3 = true` case](https://github.com/Brightspace/insights-engagement-dashboard/blob/US123012/courseTableSort/test/components/user-drill-courses-table.test.js#L133) but not for the `isS3 = false` case. That test should be added.

components/user-drill-courses-table.js
```diff
@@ -10,7 +10,7 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { RECORD } from '../consts';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
-export const ACTIVE_TABLE_COLUMNS = {
+const TABLE_COLUMNS = {
```
bemailloux, line 13: Oh cool - so the inactive table pretends it has Predicted Grades but just doesn't show the column. That's way cleaner.

test/components/user-drill-courses-table.test.js
johngwilkinson: I'll get it in another PR - it's a little finicky.

johngwilkinson:
Some strange errors in the sauce tests - might be something more than flakiness.

bemailloux (approved):

test/components/user-drill-courses-table.test.js
johngwilkinson: Since I'm still trying to get a build and got these tests worked out, I've added them here.

test/components/user-drill-courses-table.test.js
```diff
@@ -207,18 +218,28 @@ describe('d2l-insights-user-drill-courses-table', () => {
...
-				['Course name', undefined],
-				['Final Grade', numsWithTextSort],
-				['Total Time in Content (mins)', numsWithTextSort],
-				['Discussion Activity', undefined],
-				['Course Last Access', datesWithTextSort]
+				['Course name', undefined, 0],
+				['Final Grade', numsWithTextSort, 1],
+				['Total Time in Content (mins)', numsWithTextSort, 3],
+				['Discussion Activity', undefined, 4],
+				['Course Last Access', datesWithTextSort, 5]
```
bemailloux, line 238: Looks like the added numbers don't do anything here.

johngwilkinson: removed, thanks

bemailloux (approved):

PR #232, opened 2020-12-17
Move the Table Scroll Buttons Down

NicholasHallman:
fixed scrolling

FYI: @anhill-D2L @MykolaGalian @rohitvinnakota @Vitalii-Misechko @hyehayes


components/table.js
```diff
@@ -167,6 +167,7 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 			}
 
 			d2l-scroll-wrapper {
+				--d2l-scroll-wrapper-action-visible_-_top: 150px;
```
johngwilkinson, line 170: Does this also move it down when the table is near the bottom of the screen? I don't think we want that.

bemailloux (commented):
Good idea! What happens if you have a small table though (e.g. with 1 row)?

components/table.js
NicholasHallman: the button still stops at the bottom, because of it's sticky property

NicholasHallman:
> Good idea! What happens if you have a small table though (e.g. with 1 row)?

It sticks to the bottom, https://visual-diff.d2l.dev/screenshots/insights-engagement-dashboard/user-drill/2020-12-17-20.47.03.657-1c777afe4e59927255b9ea97f309ba89362f7217-report.html

components/table.js
johngwilkinson: But it's moved down relative to where it's supposed to be - I don't know if that's acceptable or not.

bemailloux:
> It sticks to the bottom, https://visual-diff.d2l.dev/screenshots/insights-engagement-dashboard/user-drill/2020-12-17-20.47.03.657-1c777afe4e59927255b9ea97f309ba89362f7217-report.html

Looks a little awkward but better than having it overlap the header 👍

bemailloux (approved):

components/table.js
NicholasHallman: I see what you're saying, a better approach would be to change the top of the d2l-sticky-element component, but they don't provide a css property to edit...

johngwilkinson: Can you run it by Kevin and see which he prefers?

PR #258, opened 2020-12-23
Skeletons do not appear for summary cards Fix

Vitalii-Misechko:
This PR fixes skeleton loading and adds Visual diffs tests.

FYI @anhill-D2L @rohitvinnakota @hyehayes @johngwilkinson @MykolaGalian 

* Functional Testing 
  * [x] Unit tests
  * [x] Test cases
* Security, devops, perf: N/A
* UX and docs
  * [x] No demo needed

The issue
image


test/visual-diff/engagement-dashboard.visual-diff.js
```diff
@@ -3,7 +3,8 @@ const VisualDiff = require('@brightspace-ui/visual-diff');
 
 describe('engagement-dashboard', () => {
 
-	const visualDiff = new VisualDiff('engagement-dashboard', __dirname);
+	// default value is 0.1 but VisualDiff overrides it to 0
+	const visualDiff = new VisualDiff('engagement-dashboard', __dirname, { tolerance: 0.05 });
```
Vitalii-Misechko, line 7: The default value is described here https://www.npmjs.com/package/pixelmatch

bemailloux: Looks like threshold is used in the colour comparison between 2 pixels. So if things are slightly the wrong colour the diff won't detect that, which is fine.

(I was worried initially that it would stop detecting if pixels were moved around on the page, but it should continue to detect that.)

model/fake-lms.js
```diff
@@ -1,3 +1,21 @@
+function parseHash() {
+	if (!window.location.hash) return {};
+
+	const hash = window.location.hash.substring(1);
+
+	return hash.split('&')
+		.map(s => ({ key: s.split('=')[0], value: s.split('=')[1] }))
+		.reduce((acc, val) => {
```
bemailloux, line 8: `Totally optional / just for fun` You could simplify this to `reduce((acc, val) => { ...acc, ...val }, {})`

bemailloux (approved):
LGTM!

model/fake-lms.js
Vitalii-Misechko: Looks neat!

test/visual-diff/engagement-dashboard.visual-diff.js
Vitalii-Misechko: It sounds to me that it's better to decouple these two sets of tests. Don't want to lower the threshold for the main dashboard.

Vitalii-Misechko:
Hooray! Finally, tests passed! 🕺 🚀

MykolaGalian (approved):

PR #304, opened 2021-01-05
US123100 [Engagement][UserDrill] Access Trend

Vitalii-Misechko:
[US123100](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F462188718528&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement][UserDrill] Access Trend

PR depends on #306 

FYI @anhill-D2L @rohitvinnakota @hyehayes @johngwilkinson @NicholasHallman

* Outside this PR
   * axis X tickers localization. Probably we would want to do localization for all three charts in a separate story.
   * applying the course filter when clicking on area rather than a point of area's series (not sure if we need this)

* Functional Testing 
  * [x] Accessibility.
  * [x] Responsive
  * [x] Add toggle to hide chart
  * [x] Test cases
    * The chart shows spinner when loading data
    * The chart applies the course filter when a user clicks on a series
    * The chart clear filter when a user makes a second click on a filtered series.
      
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo
 
### Picture 1
 
image

### Loading

image


MykolaGalian (approved):

components/access-trend-card.js
```diff
@@ -0,0 +1,273 @@
...
+			selectedCourses: { type: Object, attribute: false }
+		};
+	}
+
+	constructor() {
+		super();
+		this.data = {
+			_data: {
+				serverData: {
+					orgUnits: []
```
bemailloux, line 25: I'm not sure if you have to initialize every subfield of this object. `this.data = {};` might be fine - but I'd check with one of the Daves to make sure

components/access-trend-card.js
```diff
@@ -0,0 +1,273 @@
...
+
+			series: this._series
+		};
+	}
+
+	get _trendData() {
+		const courses = [{
+			orgUnitId: 1,
+			data: [
+				{ x: Date.UTC(2020, 1, 1), y: 43934 },
```
bemailloux, line 215: The course access max value will probably be 7 (see the [Course Access spike](https://rally1.rallydev.com/#/23525809118d/workviews?detail=%2Fuserstory%2F462269641028&view=349baf7a-c63a-4fa4-a43b-2bb09274aad7) for details. So you might want to decrease the course access counts in the test data to match reality.

bemailloux (commented):
What does it look like when it's loading?

components/access-trend-card.js
Vitalii-Misechko: Good point. I've simplified initialization in the constructor by introducing a guard condition for the first `render` run which happens before properties being assigned.
```JS
	get _series() {
		if (!this.data._data) return [];

...
```

components/access-trend-card.js
Vitalii-Misechko: Thanks! Update screenshot in the details section of the PR.

Vitalii-Misechko:
> What does it look like when it's loading?

Good question! Added a screenshot below and to the details section of the PR

image

components/access-trend-card.js
```diff
@@ -0,0 +1,269 @@
...
+	_orgUnitName(orgUnitId) {
+		const orgUnit = this._serverData.orgUnits.find(unit => unit[ORG_UNIT.ID] === orgUnitId);
+
+		return orgUnit ? orgUnit[ORG_UNIT.NAME] : '';
+	}
+
+	get _series() {
+		if (!this.data._data) return [];
+
+		const colors = UserTrendColorsIterator();
```
NicholasHallman, line 258: this will only return 40 colors, I recommend

`UserTrendColorsIterator(0,1,this._trendData.length)`

in case the user has more than 40 courses

NicholasHallman (approved):

components/access-trend-card.js
Vitalii-Misechko: Cool! Thanks for comment

Vitalii-Misechko: I'll fix colors in the next PR

PR #306, opened 2021-01-05
User Drill Trend Legend and Interactions US123294

NicholasHallman:
Implementation Notes 

The legend doesn't use HighCharts because I wanted more control over how the legend would layout in responsive views like tablets mobile etc.

 The filtering uses the existing Category filters and it has some side effects. It appears in the "applied filters" component and is applied to the general engagement dashboard and carries over into other user views. I'm going to talk with Kevin and Megan about the intended behavior to see if we want to reset it when the user leaves the learner view.  

If the other tables use this same course filter then filter interactions are very simple

Quality Notes

Functional Testing
  - Unit Tests
    - [x] Test the filters apply
    - [x] Test the filters toggle on click
  - Vis diff
    - [x] Update the Goldens for the user-drill view
  - Manual Testing
    - [x] Legend shows both active and inactive courses
    - [x] Legend item is opaque when active
    - [x] Legend item is translucent when inactive
    - [x] Legend items are filtered from the table when active 
    - [x] Legend excludes courses filtered out by the org unit tree filter
    - [x] Legend is keyboard accessible ( narrator reads out course names and tabbable )   
   
Performance 
  - Optimizations
     - [x] Event delegation (Many possible event handlers if all 40 possible courses are visible. One for keyboard and mouse)  

Dev-Ops - N/A  

Cost - N/A  

Documentation - N/A  

UX
  - Adds a legend underneath of the trend tables (which don't exist yet)
  - Demoed legend course layout to Kevin
  - Added the colors specified in the design


johngwilkinson:
Interesting approach. I thought the legend would only affect the time-series charts, not the table or summary cards (because that's what it would do without extra work if it was just a HC legend on a single chart) - but "Clicking on a course in the legend filters it out of the card and dashboard view" is kind of ambiguous (what card? was this text maybe copied from somewhere?). Definitely need to clarify that with Megan and Kevin, as well as the question about affecting the ou filter/main page (I suspect it shouldn't).

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+
+		const isUsersRecord = record => record[RECORD.USER_ID] === this.user.userId;
+		const recordOrgUnitId = record => record[RECORD.ORG_UNIT_ID];
+		// get a unique set of orgId's then get the name of those org units.
+		return Array.from(
+			new Set(this.data
+				.records
+				.filter(isUsersRecord)
+				.map(recordOrgUnitId)))
+			.map(this._orgUnitInfo.bind(this));
```
johngwilkinson, line 101: Is there a reason to prefer `bind` to an arrow function? My impression was that since es6, we don't need `bind` anymore. A quick google suggests arrow functions are preferred for being safer, clearer, and faster: https://stackoverflow.com/questions/42117911/lambda-functions-vs-bind-memory-and-performance

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+		];
+	}
+
+	get serverData() {
+		return this.data._data.serverData;
+	}
+
+	get courses() {
+
+		const isUsersRecord = record => record[RECORD.USER_ID] === this.user.userId;
```
johngwilkinson, line 93: You could just use `data.recordsByUser.get(this.user.this.userId)` I think? It is `computed` in `FilteredData`, so it shouldn't incur any performance overhead - in fact, since it is used elsewhere already, it should be a tiny bit faster.

components/user-drill-view.js
```diff
@@ -23,6 +24,34 @@ import { until } from 'lit-html/directives/until';
 export const numberFormatOptions = { maximumFractionDigits: 2 };
 const demoDate = 1608000000000; //for unit test
 
+class SelectedCourses {
```
johngwilkinson, line 27: Seems like this class belongs with the legend component.

consts.js
```diff
@@ -29,3 +29,55 @@ export const ORG_UNIT = {
...
+	'#A7FAE7',
+	'#C7FDFF',
+	'#E9E6FF',
+	'#6900A0',
+	'#990056',
+	'#FFE2DB',
+	'#FFAB61',
+];
+
+export function* UserTrendColorsIterator(start = 0, end = 40, step = 1) {
```
johngwilkinson, line 76: Could you use `USER_TREND_COLORS.values()` instead? It's an iterable: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/values#Iteration_using_.next

components/user-drill-view.js
NicholasHallman: My thinking was we would pass it into the trend charts so they will update when the courses change and use it as a record filter

johngwilkinson: Yes, but I'd prefer to see it in the legend component with `export` and imported here. Maybe I'm wrong - just seems like this is legend stuff and this file doesn't need to know about its innards.

components/courses-legend.js
NicholasHallman: I don't think I will ever understand why calling a class method through a callback inside of the class makes `this` unreferenced I was trying to write good OOP haha. I agree though which is why I had written the above two simple functions as lambdas. I also didn't know bind was slow thanks for the reference!

components/user-drill-view.js
NicholasHallman: Ooo I see sound good

components/courses-legend.js
johngwilkinson: Because you're passing a function reference, not calling it. Arrow functions almost always do the right thing, so you don't have to worry about undefined `this`es: `.map(x => this._orgUnitInfo(x))`

components/user-drill-view.js
```diff
@@ -23,6 +24,34 @@ import { until } from 'lit-html/directives/until';
...
+
+	has(value) {
+		return this.selected.has(value);
+	}
+}
+
+decorate(SelectedCourses, {
+	selected: observable,
+	toggle: action,
+});
```
Vitalii-Misechko, line 53: should we store selected courses state in browser's URL?

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+		};
+	}
+
+	_renderCourse(course, color) {
+		const containerStyles = classMap({
+			'd2l-insights-user-course-legend-item': true,
+			'd2l-insights-user-course-legend-item-filtered': this.selectedCourses.has(course.orgUnitId)
+		});
+		return html`
+		<div tabindex="0" ouid="${course.orgUnitId}" class="${containerStyles}">
```
Vitalii-Misechko, line 118: `div` doesn't have `ouid` property. Is it safe to use it? Should we use `data-`-property instead?

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+			<div class="d2l-insights-user-course-legend-color" style="background-color: ${color};"></div>
```
Vitalii-Misechko, line 119: `jfyi` Lit-element [offers](https://lit-element.polymer-project.org/guide/styles#dynamic-classes-and-styles) `styleMap` for dynamic styles. The current implementation looks good to me though.

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+			<p class="d2l-insights-user-sourse-legend-name">${course.name}</p>
+		</div>
+		`;
+	}
+
+	_renderCourses() {
+		const colors = UserTrendColorsIterator();
+		return this.courses.map(course => this._renderCourse(course, colors.next().value));
```
Vitalii-Misechko, line 127: `edge case`: What happens if a user has more than 40 courses?

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	//EVENTS
+
+	_handleInteraction(e) {
+
+		// interaction out of scope, but this will get ya started
+
+		if (e.target.parentElement === null) return;
+		const orgUnitId = e.target.getAttribute('ouid') | e.target.parentElement.getAttribute('ouid');
```
Vitalii-Misechko, line 137: Should we prefer `logical or` instead of `bitwise or` here? It seems to me the code takes `ouid` from the target and if the attribute does not exist then takes value from the parent's element.

test/visual-diff/summary-cards.visual-diff.js
```diff
@@ -3,7 +3,7 @@ const VisualDiff = require('@brightspace-ui/visual-diff');
 
 describe('summary cards', () => {
 
-	const visualDiff = new VisualDiff('summary-cards', __dirname);
+	const visualDiff = new VisualDiff('summary-cards', __dirname, { tolerance: 0.05 });
```
Vitalii-Misechko, line 6: `just curious` Does the change decrease notifications?

components/courses-legend.js
```diff
@@ -0,0 +1,159 @@
...
+					.d2l-insights-user-course-legend-item:nth-child(4n) {
+						margin-right: 10px;
+					}
+				}
+			`
+		];
+	}
+
+	get serverData() {
+		return this.data._data.serverData;
```
Vitalii-Misechko, line 88: `thought` accessing another's object private field doesn't look good. Should we make `this.data._data` public (remove underscore)?

consts.js
Vitalii-Misechko: @NicholasHallman It seems colors out of order. Here is what I'm getting localy

image

components/user-drill-view.js
```diff
@@ -23,6 +24,34 @@ import { until } from 'lit-html/directives/until';
...
+		}
+		else {
+			this.selected.add(value);
+		}
+	}
+
+	has(value) {
+		return this.selected.has(value);
+	}
+}
```
Vitalii-Misechko, line 48: it seems that this class requires `size` getter to implement no course selected case. If no course selected then show all courses

Vitalii-Misechko: Here is how I use it in [my PR](https://github.com/Brightspace/insights-engagement-dashboard/pull/304/files#diff-bb09fb33688c671275aaae22a3d0a2b436f51f5956f356b936771cb09c5e380eR255-R263)

```JS
.filter(course => this.selectedCourses.has(course.orgUnitId) || this.selectedCourses.size === 0);
```

components/courses-legend.js
NicholasHallman: Yeah I understand why but my intuition is always confused. Class methods only exist when they are members of an instanced class, so `this` in a class method should always be the instance of the class, since it doesn't exist otherwise. In JS you can decouple the method from the class by referencing it which to me is forever confusing but I understand that to the computer a class method and a function are the same thing and it doesn't care about by understanding of abstractions. But it still makes me sad haha

consts.js
NicholasHallman: Hmm that looks right to me? I'll try merging a new branch with your cards and this legend and see if it all makes sense

components/user-drill-view.js
NicholasHallman: On one hand, we don't want it to stay the same from user to user, if you leave a user-drill and then enter a new one the legend filter should reset. On the other hand if I'm in a user drill with a course filter and want to share it then it should be saved in the url state. So I'm looking for the solution to these two scenarios atm

components/courses-legend.js
NicholasHallman: Ah yes I should make it data-ouid thank you

components/courses-legend.js
NicholasHallman: Good to know thanks!

components/courses-legend.js
NicholasHallman: Ah, I will ask... My guess is we put up a message and hide all of the trends but I don't know for sure.

components/courses-legend.js
NicholasHallman: It should be `||` but funny enough and unintentionally this seems to work! I think when `e.target.getAttribute('ouid')` is `null` and we use the `|` operation it gets coerced into 0 and then `0 | anything == anything` so it works but I don't like it

test/visual-diff/summary-cards.visual-diff.js
NicholasHallman: so far yes

components/courses-legend.js
NicholasHallman: Megan says we should start re-using the colors so I'll make the generator a little more useful

components/courses-legend.js
johngwilkinson: Or add a getter for serverData - might be more readable

consts.js
Vitalii-Misechko: I think that it's hard to differentiate courses when they have one base color. Have a look at the design screenshot

image

NicholasHallman: Yeah we should bring this up with kevin, maybe they were added to the table wrong

components/courses-legend.js
Vitalii-Misechko: I believe this part is outside the scope of this PR )

consts.js
NicholasHallman: Kevin says the hue ordering is unintentional and is going to update the rally ticket with better groupings

components/user-drill-view.js
```diff
@@ -366,6 +376,16 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 			></d2l-summary-cards-container>
 
 			<div class="d2l-insights-user-drill-view-content">
+			${ this.isDemo ? html`
```
NicholasHallman, line 379: Is this enough to keep the trends off prod but in tests? or should we do something else?

components/user-drill-view.js
```diff
@@ -366,6 +376,16 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+				<d2l-insights-courses-legend
+					.data="${this.data}"
+					.user="${this.user}"
+					.selectedCourses="${this.selectedCourses}"
+					?skeleton="${this.skeleton}"
+				></d2l-insights-courses-legend>
+			` : nothing }
+
+				</d2l-insights-courses-legend>
```
johngwilkinson, line 388: this line shouldn't be here, should it?

components/user-drill-view.js
johngwilkinson: Manual testing should confirm, but I expect so, yes.

NicholasHallman: I tested with local lms and it didn't show. Not sure if that qualifies

Vitalii-Misechko (approved):

MykolaGalian (approved):

PR #336, opened 2021-01-13
US123374 [Engagement][UserDrill] Access Trend Interaction

Vitalii-Misechko:
[US123374](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F464899990920&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement][UserDrill] Access Trend Interaction

FYI @anhill-D2L @rohitvinnakota @hyehayes  @bemailloux @MykolaGalian

* Functional Testing 
  * [x] Accessibility
     * A user can navigate to each series with keyboard.
     * Screenreader announces each series as " `<Course name>`, series 1 of 3 with 8 data points.`"
     * A point is read as "2. Monday, Feb 10, 2020, Date,  3 Course Access Count. Course 1.". "2. `<number of point of the series>` `<values of axis X>`, Date, `<value of axis Y>` Course Access Count  `<series name>`"
     * Series activation via a keyboard is not announced. (Will be addressed in a separate story for all charts?)
  * [x] Responsive
  * [x] Test cases
    * Click on a series marks it selected. All other courses are greyed.
    * Legend reflects selected courses properly
    * Click on a selected series unselect it.
    * Once all series are deselected then the course filter is cleared. All series deselected = show all series
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo

Course-Access-Over-Time-With-Lines


Vitalii-Misechko:
Working on a test that verifies the color of a selected course.

components/access-trend-card.js
```diff
@@ -83,10 +83,69 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const p1 = twoPoints[0];
+		const p2 = twoPoints[1];
+		const m = (p2.y - p1.y) / (p2.x - p1.x);
+		const b = p1.y - m * p1.x;
+
+		// slope-intercept form of a line eqution
+		return (x) => m * x + b;
+	}
+
+	_getClickedAreaByXY(xVal, yVal) {
```
johngwilkinson, line 122: Checking: the solution here doesn't work? https://www.highcharts.com/forum/viewtopic.php?t=39786

johngwilkinson: nvm,  see you tried it

components/access-trend-card.js
```diff
@@ -83,10 +83,69 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 	_toggleFilter(orgUnitId) {
 		this.selectedCourses.toggle(orgUnitId);
 	}
 
+	_twoClosestPoints(xValue, points) {
+		// it assumes that points are ordered by elem.x value
+		return points.reduce((result, point) => {
+			if (result.length === 2) return result;
+
+			if (point.x > xValue) {
```
johngwilkinson, line 103: This looks like it returns the two points _following_ the given x value - but don't we want the largest point with a smaller x value and the smallest point with a larger x value, i.e. the two points _surrounding_ the given x value?

johngwilkinson:
Are unit tests (for point selection) coming?

components/access-trend-card.js
johngwilkinson: One other thing to check: area charts stack the area, right? So, if series 1 has value 2 and series two has value 3, the point for series 2 is drawn at 2  + 3 = 5? Is that accounted for somewhere here?

components/access-trend-card.js
Vitalii-Misechko: It's good that you've asked. This nudged me to restest these two options. 

The current option has appeared because it fixes a minor case when the chart dynamically adds or removes course/series. I found this case in an early prototype of the solution.

The final implementation does not require hiding/unhiding series of the chart. And we settled with selection behavior that allows showing markers while hovering the mouse cursor over a series area. Thus, I think, this workaround can be removed in favor of the standard highcharts behavior.

components/access-trend-card.js
Vitalii-Misechko: Went in favor of default behavior. Have a look [comment here](https://github.com/Brightspace/insights-engagement-dashboard/pull/336#discussion_r556736926)

components/access-trend-card.js
```diff
@@ -255,15 +273,16 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	get _series() {
 		if (!this.data._data) return [];
 
-		const colors = UserTrendColorsIterator();
+		const selected = (course) => this.selectedCourses.has(course.orgUnitId) || this.selectedCourses.size === 0;
+		const colors = Array.from({ length: this._trendData.length }, function() { return this.next().value; }, UserTrendColorsIterator(0, 1, this._trendData.length));
```
johngwilkinson, line 277: I believe you can just do `Array.from(UserTrendColorsIterator(0, 1, this._trendData.length))`

johngwilkinson (approved):

components/access-trend-card.js
Vitalii-Misechko: > This looks like it returns the two points following the given x value - but don't we want the largest point with a smaller x value and the smallest point with a larger x value, i.e. the two points surrounding the given x value?

It takes surrounding points. The first point is overridden each cycle until the code finds a bigger x value. 

Imagine you have [1 3 5 7]. A user clicked at 4. 

Step num |result|point|Comment
-|-|-|-
1 |[1]|1|
2|[3]|3|result is overridden
3|[3,5]|5|the code found point from the right 
4|[3,5]|7| do nothing we already find all points

components/access-trend-card.js
Vitalii-Misechko: Thanks. Looks shorter

Vitalii-Misechko:
> Are unit tests (for point selection) coming?

Sure, they are. I'm going to pass initialized `selectedCourses` and check if the colors of the chart are correct

components/access-trend-card.js
Vitalii-Misechko: > One other thing to check: area charts stack the area, right? So, if series 1 has value 2 and series two has value 3, the point for series 2 is drawn at 2 + 3 = 5? Is that accounted for somewhere here
Nope. I believe, highcharts just uses appropriate scale.

components/access-trend-card.js
NicholasHallman: Anybody down to go golfing? 😂

```javascript
[...UserTrendColorsIterator(0, 1, this._trendData.length)]
```

NicholasHallman (approved):

PR #341, opened 2021-01-13
Fix - All Selected Filters Alters Behavior

NicholasHallman:
Implementation Notes:

This is fairly simple. The chart tells it's filter what "all" looks like, as in, if all of my categories were selected what items would be in that resulting array. 

then anytime we add something to the filter it checks the intersection of the selected and all sets, and if the intersection is the same as all then it's full, and we reset the selected set back to nothing. ez pz.

~~We do this on firstUpdated because we don't necessarily need the data in the arrays, we just need the categories the data~~ ~~slots into. So we can let the filters know what "all" looks like during the skeleton.~~ 
I'm dumb next commit will add it to the filters constructor

**Before:**

neJHyF8f2K

**After:**

3VNW52a3Hi

Testing
  - Manual tests include
    - Going through all of the category filters, except for time in content vs grade which doesn't toggle, and making sure that the selectedCategories is empty once all items are selected (there are visual cues, like the pie chart resetting to an infixed position).
  - Unit tests
    - [x] discussion activity should reset to none selected when all are selected
    - [x] current grade should reset to none selected when all are selected
    - [x] course access should reset to none selected when all are selected
    - [x] user drill legend should reset to none selected when all are selected   
    
Performance
  - hypothetically this is a performance improvement. We only filter when some items are selected now. 
  
All Other - N/A


johngwilkinson:
I think I missed the background - did Kevin ask for this? This changes filter behaviour so filters reset whenever all items are selected. I'm not sure that's good: I can't select a, then b, then c, then deselect b (assuming a, b, and c are all the categories). Maybe it's a better workflow, but maybe it's not as good. Is there a case where it's necessary?

I was going to comment on "firstUpdated" because it will not pick up changes to the categories list, but I see you're planning to change that. Doing it in `render` is usually the right thing, since that gets called when something changes.

NicholasHallman:
> I think I missed the background - did Kevin ask for this? This changes filter behaviour so filters reset whenever all items are selected. I'm not sure that's good: I can't select a, then b, then c, then deselect b (assuming a, b, and c are all the categories). Maybe it's a better workflow, but maybe it's not as good. Is there a case where it's necessary?

I asked Megan 
> "Do we want to reset the selected items back to "none selected" when you select everything?" 

> "yes we do"

When I was talking with Kevin about this a few days ago he noted that it was confusing, and especially so for screen readers.

> I was going to comment on "firstUpdated" because it will not pick up changes to the categories list, but I see you're planning to change that. Doing it in `render` is usually the right thing, since that gets called when something changes.

The categories for the engagement dashboard cards don't change so I'm going to add them to their respective filters constructor.

For the course legend however, the categories change anytime you change the org filter. So that one will be done on render / updated

components/course-last-access-card.js
```diff
@@ -54,6 +55,9 @@ export class CourseLastAccessFilter extends CategoryFilter {
 			'caf'
 		);
 		this._urlState = new UrlState(this);
+		this.all = new Set(
+			[...DATA_BUCKETS].map((_, i) => i)
```
johngwilkinson, line 59: how about `new Set(DATA_BUCKETS.keys())`?

components/current-final-grade-card.js
```diff
@@ -32,6 +34,8 @@ export class CurrentFinalGradesFilter extends CategoryFilter {
 			'cgf'
 		);
 		this._urlState = new UrlState(this);
+
+		this.all = new Set(CATEGORIES.slice(0, 10));
```
johngwilkinson, line 38: Is the `slice` just to make a copy for code safety? You can just do `slice()` (no parameters) for that - although we more commonly do `[...CATEGORIES]`.

model/categoryFilter.js
```diff
@@ -17,12 +18,29 @@ export class CategoryFilter {
 		if (!isApplied) this.selectedCategories.clear();
 	}
 
+	set all(allValues) {
```
johngwilkinson, line 21: `all` should be decorated with `computed`, which makes this setter an action, so that it doesn't trigger more than one update (might be true in one or two other places in this PR too)

components/current-final-grade-card.js
NicholasHallman: the slice is because we only want categories 0 - 9, 10 isn't interactable but it still appears on the x-axis of the chart

components/course-last-access-card.js
NicholasHallman: TIL!

model/categoryFilter.js
```diff
@@ -62,5 +63,6 @@ decorate(CategoryFilter, {
 	selectCategory: action,
 	setCategories: action,
 	toggleCategory: action,
+	setAll: action,
```
johngwilkinson, line 66: `_all` should be an `observable`, I just realized, since what we render depends on its value. It might not matter in practice, since generally it is changes to `selectedCategories` that trigger redraws - but there could be edge cases.

johngwilkinson (approved):

components/course-last-access-card.js
```diff
@@ -55,7 +55,7 @@ export class CourseLastAccessFilter extends CategoryFilter {
 			'caf'
 		);
 		this._urlState = new UrlState(this);
-		this.setAll(new Set(DATA_BUCKETS.keys()));
+		this._all = new Set(DATA_BUCKETS.keys());
```
johngwilkinson, line 58: you probably want a constructor on the base class and `super(new Set(...))`

johngwilkinson (approved):

johngwilkinson (approved):

components/courses-legend.js
```diff
@@ -24,9 +25,19 @@ export class SelectedCourses {
 		}
 		else {
 			this.selected.add(value);
+			if (this._all && this.isAllSelected()) {
```
Vitalii-Misechko, line 28: `minor` I'd suggest moving check `this._all` directly into `isAllSelected` function. This way no one forgets to check if _all for undefined

components/courses-legend.js
```diff
@@ -171,6 +192,12 @@ class CoursesLegend extends SkeletonMixin(Localizer(MobxLitElement)) {
 		).map(orgUnitInfo);
 	}
 
+	//LIFECYCLE
+
+	updated() {
+		this.selectedCourses.setAll(new Set(this.courses.map(course => course.orgUnitId)));
```
Vitalii-Misechko, line 198: `thought` I'm glad that this doesn't cause an infinite cycle.  We basically trigger a new update here by changing MobX observable. Mobx and Lit-Element somehow handle this correctly.

model/categoryFilter.js
```diff
@@ -17,12 +18,30 @@ export class CategoryFilter {
 		if (!isApplied) this.selectedCategories.clear();
 	}
 
+	//@action
+	setAll(allValues) {
+		this._all = allValues;
+		if (this._all && this.isAllCategoriesSelected()) {
```
Vitalii-Misechko, line 24: `minor` I'd suggest moving check `this._all` directly into isAllCategoriesSelected function. This way no one forgets to check if `_all` for undefined

test/components/course-last-access-card.test.js
```diff
@@ -54,6 +54,20 @@ describe('d2l-insights-course-last-access-card', () => {
...
+			filter.toggleCategory(0);
+			filter.toggleCategory(1);
+			filter.toggleCategory(2);
+			filter.toggleCategory(3);
+			filter.toggleCategory(4);
+			filter.toggleCategory(5);
+			filter.toggleCategory(6);
+			expect([...filter.selectedCategories]).to.eql([]);
+		});
+	});
```
Vitalii-Misechko, line 69: Nice test!

Vitalii-Misechko (approved):
LGTM!

components/courses-legend.js
NicholasHallman: because courses is computed I think, since courses doesn't change it doesn't re-render... maybe?? But yeah I was worried about adding _all as an observable for that reason. must be the lit-mobx adobe thingy?

PR #358, opened 2021-01-15
R20.21.02 + hide alert unless demo

NicholasHallman:
Implementation notes

This is the release cut with an extra commit to hide the alert message for prd.


Vitalii-Misechko (approved):
Short and sweet!

johngwilkinson (approved):

johngwilkinson:
I believe the release branch will get deleted when you merge, so you might want to restore it so we don't forget to include this commit.

PR #365, opened 2021-01-18
Migrate Tests from Travis to GHActions

NicholasHallman:
Dragon Slayer. Migrate the travis jobs to github actions

- [x] Linting
- [x] Sauce Labs

Once merged we can disable travis on this repo

FYI @anhill-D2L @MykolaGalian @bemailloux  @hyehayes

### DIFF
--- a/test/engagement-dashboard.test.js
+++ b/test/engagement-dashboard.test.js
@@ -2,12 +2,12 @@ import '../engagement-dashboard.js';
 import { expect, fixture, html } from '@open-wc/testing';
 import { LastAccessFilter } from '../components/last-access-card';
 import { runConstructor } from '@brightspace-ui/core/tools/constructor-test-helper.js';
+import { trySelect } from './tools.js';
 
 describe('d2l-insights-engagement-dashboard', () => {
 
 	describe('accessibility', () => {
-		it('should pass all axe tests', async function() {
-			this.timeout(7000);
+		it('should pass all axe tests', async() => {
 
 			const el = await fixture(html`<d2l-insights-engagement-dashboard
 				course-access-card courses-col discussions-card discussions-col
@@ -15,13 +15,10 @@ describe('d2l-insights-engagement-dashboard', () => {
 				system-access-card tic-col tic-grades-card
  				demo
  			></d2l-insights-engagement-dashboard>`);
-			// need for this delay might be tied to the mock data async loading in engagement-dashboard.js
-			await new Promise(resolve => setTimeout(resolve, 1500));
-
 			// close the default view dialog that shows up. It causes browsers on OSX to assign aria-attributes and
 			// roles to buttons in the background that are not normally allowed
-			const defaultViewDialog = el.shadowRoot.querySelector('d2l-insights-default-view-popup');
-			defaultViewDialog.opened = false;
+			const defaultViewDialog = trySelect(el.shadowRoot, 'd2l-insights-default-view-popup');
+			if (defaultViewDialog !== null) defaultViewDialog.opened = false;
 			// wait for the dialog closing animation to finish
 			await new Promise(resolve => setTimeout(resolve, 500));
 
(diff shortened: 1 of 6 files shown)


johngwilkinson (approved):

Vitalii-Misechko (approved):

NicholasHallman:
@johngwilkinson @Vitalii-Misechko Can you review the changes I made to the tests to make them consistent? specifically the new tooling trySelect and trySelectAll? Do you think these are good approaches to solving safari's variable parent child render times?

test/tools.js
```diff
@@ -0,0 +1,27 @@
+
+// these trySelect tools are useful with safari which can take a variable amount of time to load in a
+// fixture. This solves the "children aren't taken into account of await updateComplete" problem by working
+// like the selenium wait to exist functions.
+export const trySelectAll = async(elm, query, attempts = 50) => {
```
johngwilkinson, line 5: These look fine to me if they work

test/model/data.test.js
```diff
@@ -91,7 +92,7 @@ describe('Data', () => {
 			// trigger a truncated reload and allow recordProvider to resolve
 			sut._selectorFilters.role = new RoleSelectorFilter({ serverData: { selectedRolesIds: null, isRecordsTruncated: true } });
 			sut.selectedRoleIds = [mockRoleIds.student, TRUNCATE_IF_THIS_ROLE_IS_PRESENT];
-			await new Promise(resolve => setTimeout(resolve, 0));
+			await new Promise(resolve => setTimeout(resolve, 1000));
```
Vitalii-Misechko, line 95: Just curious why did you change 0 to 1000 here?

test/engagement-dashboard.test.js
```diff
@@ -132,26 +129,20 @@ describe('d2l-insights-engagement-dashboard', () => {
...
+					?grade-col="${expectedList.includes('grade-col')}"
+					?last-access-col="${expectedList.includes('last-access-col')}"
+					?tic-col="${expectedList.includes('tic-col')}"
+					demo
+				></d2l-insights-engagement-dashboard>`);
 
-				const usersTable = el.shadowRoot.querySelector('d2l-insights-users-table');
-				const innerTable = usersTable.shadowRoot.querySelector('d2l-insights-table');
+				const usersTable = await trySelect(el.shadowRoot, 'd2l-insights-users-table', 80);
+				const innerTable = await trySelect(usersTable.shadowRoot, 'd2l-insights-table');
```
Vitalii-Misechko, line 145: @NicholasHallman.

Cool! I like these new functions! 👍 

These changes nudged me to make a quick search and I found that it seems [a known problem regarding awaiting when all children of a container complete their update](https://github.com/Polymer/lit-element/issues/155). I guess that we might want to try the proposed code in the future.
```JS
updateComplete() {
  return Promise.all([super.updateComplete(), this.someChildElement.updateComplete()];
}
```

Vitalii-Misechko (approved):
LGTM!

test/model/data.test.js
NicholasHallman: Oh yes I'll change that back thanks!

test/engagement-dashboard.test.js
NicholasHallman: I also stumbled on that but I don't think the Promise.All solution will work for us. I think we will need to do a promise chain. something like

```javascript
function appendChildAndWait(children, el, query) {
	const child = el.querySelector(query);
	children.append(child);
	return child.updateComplete();
}

function updateComplete() {
	const children = [];
	return super.updateComplete()
		.then(() => appendChildAndWait(children, this.shadowRoot, 'child-component'))
		.then(() => appendChildAndWait(children, children[children.length - 1], 'childs-child'))
		.then(() => appendChildAndWait(children, children[children.length - 1], 'childs-childs-child'));
}
```

because before the parent finishes updating the child may not exist yet. So we would need to write unique chains for each test. The variable time method felt a little more general but I agree this is the right way to do it

NicholasHallman: Or as a simple abstraction

```javascript
function setupQueryPromiseChain(queries, component) {
	component.updateComplete = async() => {
		await super.updateComplete();
		let child = this.shadowRoot.querySelector(queries[0]);
		for (let i = 1; i < queries.length; i++) {
			await child.updateComplete();
			child = child.querySelector(queries[i]);
		}
		await child.updateComplete();
	};
}
```

Vitalii-Misechko: Reliable for tests `updateComplete` doesn't look as simple as I thought from the start. Thanks for comments.
It just occurs to me that a component in a chain can report that it completed rendering while it just started a fetch operation (the component in loading mode). So, it seems that  `updateComplete` should be aware of its children and therfore, placed directly in components.

PR #370, opened 2021-01-22
Refactor Table Sort as a Mixin, supply sort functions for ease of use

NicholasHallman:
A little cleanup so we don't need to reuse the same functions everywhere. The mixin adds the needed table properties so the table updates when it detects a sort change. The sorting functions are defined as an array in the constructor which is much easier to read. Finally the sorts have been named so they can be easily added. 

Functional Tests
  - Behavior is maintained, all tests passing.
  - Manual tests side by side with master are the same

All others N/A

### DIFF
--- a/components/users-table.js
+++ b/components/users-table.js
@@ -3,6 +3,7 @@ import '@brightspace-ui-labs/pagination/pagination';
 import '@brightspace-ui/core/components/inputs/input-text';
 import { action, computed, decorate, observable, reaction } from 'mobx';
 import { css, html } from 'lit-element';
+import { defaultSort, SortMixin, userNameSort } from '../model/sorts.js';
 import { formatNumber, formatPercent } from '@brightspace-ui/intl';
 import { RECORD, USER } from '../consts';
 import { COLUMN_TYPES } from './table';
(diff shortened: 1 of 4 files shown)


model/sorts.js
```diff
@@ -0,0 +1,66 @@
...
+export const courseNameSort = (chosenOrder, column) =>
+	(course1, course2) => {
+		const courseId1 = course1[column].toLowerCase();
+		const courseId2 = course2[column].toLowerCase();
+		return (courseId1 > courseId2 ? chosenOrder[0] :
+			courseId1 < courseId2 ? chosenOrder[1] :
+				chosenOrder[2]);
+	};
+
+const reverseSorts = []; // add sorts here to revers them
```
bemailloux, line 34: I'm surprised tests pass without anything being added here. Wasn't there one column in the original users table where the sorting _was_ reversed?

components/users-table.js
```diff
@@ -107,6 +106,16 @@ class UsersTable extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
+		this.sorts = [
+			undefined, // we don't sort the checkboxes
+			userNameSort,
+			defaultSort,
+			defaultSort,
+			defaultSort,
+			defaultSort,
+			defaultSort
+		];
```
bemailloux, line 117: `Just a thought` IMO sorting information should go with `get columnInfo`, but that's tricky because that getter now returns the visible columns only.

bemailloux (approved):
LGTM!

model/sorts.js
NicholasHallman: Ah yep it's because they are being sorted in reverse in the function, not because of the given order. Fixed in latest commit

bemailloux: Ah got it. Took a bit of squinting / remembering how sorting works to figure it out but now I understand why you didn't need to change any tests. 👍

MykolaGalian (approved):

PR #377, opened 2021-01-29
US124027 Grades data

MykolaGalian:
[US124027](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F483691472192&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement][UserDrill] Add data to Grades Trend

Screenshot from local LMS test:

Functional:
Local LMS work as expected with test data (with new POST API https://github.com/Brightspace/lms/pull/6143),
also with user-drill page reload.

Accessibility: n/a
Security: n/a
Devops: n/a
UI: approved by Kevin

@Vitalii-Misechko @anhill-D2L @bemailloux


components/grades-trend-card.js
```diff
@@ -200,40 +203,12 @@ class GradesTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-				[Date.UTC(2020, 1, 7), 30],
-				[Date.UTC(2020, 1, 14), 25],
-				[Date.UTC(2020, 1, 21), 40],
-				[Date.UTC(2020, 1, 28), 55],
-				[Date.UTC(2020, 2, 4), 25]
-			]
-		}];
-
-		return courses;
+		return toJS(this.userData.courseGrades).map(grades => {
```
johngwilkinson, line 206: you shouldn't need `toJS`

components/user-drill-view.js
```diff
@@ -336,6 +335,17 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return numCourses < 10;
 	}
 
+	get _userData() {
+		if (!this.__userData) {
+
+			this.__userData = new UserData({
+				userServerData: fetchUserData // TODO: demo data
```
johngwilkinson, line 342: remove demo data comment?

model/userData.js
```diff
@@ -0,0 +1,37 @@
...
+		};
+	}
+
+	async loadData(orgUnitIds, userId) {
+		this.isLoading = true;
+		try {
+			this.userData  = await this.userServerData(orgUnitIds, userId);
+			this.isLoading = false;
+		} catch (ignored) {
+			console.log('error');
```
johngwilkinson, line 23: this should pop up an error message, the same as when the main data load fails

model/userData.js
```diff
@@ -0,0 +1,37 @@
+import { computed, decorate, observable } from 'mobx';
+
+/**
+ * User Data from the server
+ */
+export class UserData {
+	constructor({ userServerData }) {
+		this.userServerData = userServerData;
```
johngwilkinson, line 8: `fetchUserData` is a better name (the verb indicates that this is a function, not a data object)

components/user-drill-view.js
```diff
@@ -165,6 +167,14 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 		`];
 	}
 
+	firstUpdated() {
+		this._userData.loadData(this.userOrgUnitIds, this.user.userId);
```
johngwilkinson, line 171: This won't fetch data again if the org units change - I think that can happen if the user adjusts the ou or semester filter while on this page, but I'm not sure. Possible fix would be to move this into `render`.

model/userData.js
```diff
@@ -17,10 +18,12 @@ export class UserData {
...
 		this.isLoading = true;
 		try {
-			this.userData  = await this.userServerData(orgUnitIds, userId);
+			this.userData  = await this.fetchUserData(orgUnitIds, userId);
 			this.isLoading = false;
+			this.isQueryError = false;
 		} catch (ignored) {
-			console.log('error');
+			this.isQueryError = true;
+			console.log('user-query-error');
```
johngwilkinson, line 26: remove `console.log`

johngwilkinson (commented):
Looks good. Be sure to check with Kevin on the appearance and working of the error message.

components/user-drill-view.js
```diff
@@ -329,14 +322,49 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+
+			this.__userData = new UserData({
+				fetchUserData : this.isDemo ? fetchDemoUserData : fetchUserData
+			});
+		}
+
+		return this.__userData;
+	}
+
+	get allCourses() {
```
johngwilkinson, line 342: `computed`?

components/user-drill-view.js
```diff
@@ -329,14 +322,49 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		return this.data.orgUnitTree.selected.length !== 0 ? allSelectedCourses : this.allCourses;
+	}
+
+	get userIdFromUrl() {
+		const urlParams = new URLSearchParams(window.location.search);
+		return Number(urlParams.get('v').split(',')[1]);
+	}
+
 	render() {
+		if (this.filteredOrgUnitIds !== this.userOrgUnitIds) {
```
johngwilkinson, line 361: Possible future optimization (not this PR, I think): if `userData` were filtered by the courses visible in `data`, then we would not need to reload it on every ou/semester filter change. Similar to how `data` works, we would only need to reload if we need data for an ou that we don't already have client-side.

johngwilkinson (approved):

NicholasHallman:
I see a lot of extra space in the discussion activities section of the user-table. Do you think you can fix it in this PR or should we make it a separate defect? 
extra space

NicholasHallman (approved):

MykolaGalian:
> I see a lot of extra space in the discussion activities section of the user-table. Do you think you can fix it in this PR or should we make it a separate defect?
> extra space

In next PR

PR #396, opened 2021-02-08
Reload userData from server

MykolaGalian:
[US124027](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F483691472192&sharedViewId=328666453148&fdp=true?fdp=true): [Engagement][UserDrill] Add data to Grades Trend

Optimization - only need to reload userData if we need grades data for course Ids that we don't already have on the client-side.
Fixed additional spacing in the table.

Functional tests:
Local LMS work as expected with test data.

Accessibility: n/a
Security: n/a
Devops: n/a
UI: n/a

@Vitalii-Misechko @anhill-D2L @bemailloux


components/grades-trend-card.js
```diff
@@ -206,12 +206,27 @@ class GradesTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+				return {
+					orgUnitId: grades.courseId,
+					data: grades.gradesData.map(item => [item.date, item.grade * 100])
+				};
+			});
+	}
+
+	get _filteredOrgUnitIds() {
+		const allSelectedCourses = this.data.orgUnitTree.allSelectedCourses.sort((a, b) => a - b);
+		return this.data.orgUnitTree.selected.length !== 0 ? allSelectedCourses : this.allCourses;
```
johngwilkinson, line 221: `.filter(...includes(...))` is basically O(n*m). You can improve that to O(n) by returning a Set here and adjusting the calling code accordingly.

components/grades-trend-card.js
```diff
@@ -206,12 +206,27 @@ class GradesTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	get _trendData() {
-		return [...this.userData.courseGrades].map(grades => {
-			return {
-				orgUnitId: grades.courseId,
-				data: grades.gradesData.map(item => [item.date, item.grade * 100])
-			};
-		});
+		return [...this.userData.courseGrades]
```
johngwilkinson, line 209: In future, we'll have similar code for each trend card, so it will make sense to move it to the user data class when that comes up. This is good for now, though.

components/user-drill-view.js
```diff
@@ -347,14 +347,14 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-	get userOrgUnitIds() {
+	get _filteredOrgUnitIds() {
 		const allSelectedCourses = this.data.orgUnitTree.allSelectedCourses.sort((a, b) => a - b);
 		return this.data.orgUnitTree.selected.length !== 0 ? allSelectedCourses : this.allCourses;
 	}
 
 	render() {
-		if (this.filteredOrgUnitIds !== this.userOrgUnitIds) {
-			this.filteredOrgUnitIds = this.userOrgUnitIds;
+		if (this.filteredOrgUnitIds.length < this._filteredOrgUnitIds.length) {
```
johngwilkinson, line 356: Comparing length isn't sufficient because the list could stay the same size or get shorter but include different org units. The condition is "some id in the new list is not in the old list", like here: https://github.com/Brightspace/insights-engagement-dashboard/blob/master/model/selectorFilters.js#L125

components/grades-trend-card.js
```diff
@@ -206,12 +206,19 @@ class GradesTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			.map(grades => {
+				return {
+					orgUnitId: grades.courseId,
+					data: grades.gradesData.map(item => [item.date, item.grade * 100])
+				};
+			});
+	}
+
+	get _filteredOrgUnitIds() {
+		const allSelectedCourses = this.data.orgUnitTree.allSelectedCourses.sort((a, b) => a - b);
```
johngwilkinson, line 220: no need to sort, I think

johngwilkinson (commented):
looks good but for a unit test and the one comment

MykolaGalian:
fixed additional spacing in the table

components/table.js
```diff
@@ -163,7 +163,7 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 				padding-right: 30px;
 			}
 			.d2l-insights-table-table td:not(.d2l-insights-table-cell-first):not(td>d2l-icon) {
-				min-width: 130px;
+				min-width: 30px;
```
johngwilkinson, line 166: Is this a fix for the discussions column issue that Kevin pointed out yesterday, or is this for some other reason?

johngwilkinson (approved):

components/table.js
MykolaGalian: Yes this is the column problem raised by Kevin

NicholasHallman: Does this affect the responsiveness of the table at all?

MykolaGalian: Judging by the vis-diff tests, there have been changes by a couple of pixels

NicholasHallman (approved):

PR #397, opened 2021-02-08
Remove puppeteer and visual-diff from dev dependencies

bemailloux:
They're not needed since the visual diff CI job installs its own dependencies, and they slow down `npm install`ing quite a bit. Plus puppeteer is about to get updated and that might cause sync issues if we don't keep up.

A better option is just to install locally when needed as described in the readme.

More context: https://d2l.slack.com/archives/C0PHG3QB0/p1612808089086800

FYI @anhill-D2L @Vitalii-Misechko @johngwilkinson

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,12 +118,15 @@ Note: Place all visual tests in the ./test/visual-diff folder or they will be ig
 To run diff tests locally
 
 ```
+# install puppeteer and visual-diff locally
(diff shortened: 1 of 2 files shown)


NicholasHallman (approved):
Awesome!

MykolaGalian (approved):

PR #406, opened 2021-02-10
US124027: hook up user drill to metron bff

johngwilkinson:
Quality Notes
- functional: gets expected (canned) data from BFF (which calls LMS, after fix in related PR)
- security: n/a
- perf: n/a
- devops: n/a
- ux/docs: n/a

FYI @MykolaGalian @Vitalii-Misechko


engagement-dashboard.js
```diff
@@ -261,6 +263,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 				.isStudentSuccessSys="${this._serverData.serverData.isStudentSuccessSys}"
 				org-unit-id="${this.orgUnitId}"
 				.viewState="${this._viewState}"
+				.metronEndpoint="${this.metronEndpoint}"
```
bemailloux, line 266: This should be `metron-endpoint` instead of `.metronEndpoint`. Unless you want to make it `attribute: false`

model/lms.js
```diff
@@ -16,7 +16,7 @@ const dataEndpoint = '/d2l/api/ap/unstable/insights/data/engagement';
 const relevantChildrenEndpoint = orgUnitId => `/d2l/api/ap/unstable/insights/data/orgunits/${orgUnitId}/children`;
 const ouSearchEndpoint = '/d2l/api/ap/unstable/insights/data/orgunits';
 const saveSettingsEndpoint = '/d2l/api/ap/unstable/insights/mysettings/engagement';
-const userDrillDataEndpoint = '/d2l/api/ap/unstable/insights/data/engagementUserDrill';
+const userDrillDataEndpoint = '/unstable/insights/data/userdrill';
```
bemailloux, line 19: Is `unstable` still necessary? I thought that was a Valence thing.

engagement-dashboard.js
johngwilkinson: the `.` accesses the property, not the attribute

model/lms.js
johngwilkinson: It's in the bdp-query path. We could change it, but I thought keeping the convention of "unstable" = "not public" might be useful.

engagement-dashboard.js
bemailloux: Right, but that means in `user-drill-view.js` having metronEndpoint as an attribute isn't necessary. So 
```
metronEndpoint: { type: String, attribute: 'metron-endpoint' }
``` 
can be changed to 
```
metronEndpoint: { type: String, attribute: false }
``` 
over there.

johngwilkinson: It could. I don't see a strong reason either way, but being able to set the endpoint in plain html is potentially nice and basically free.

johngwilkinson:
Ready for review. Does it send review request emails when I change from "draft" to "ready". I don't know.

bemailloux:
> Ready for review. Does it send review request emails when I change from "draft" to "ready". I don't know.

It does not

bemailloux (approved):

johngwilkinson:
Updated to actually hook up the data to all three trend cards (one was already done). This is pretty repetitive, but I think refactoring can wait until this is merged.

My belief is that this will be all we need to provide the user trends for this release - given that we finish hooking up the three metrics server-side.

NicholasHallman (approved):

johngwilkinson:
Visual diff change is just due to fixing "Content View Over Time" to be "Content Views Over Time"

johngwilkinson:
Okay, one more vis diff merge, because we had a bad golden
image

johngwilkinson:
What are the odds of a good vis diff run? I think I've gone at least ten times now without a single one where every capture is correct. :(

johngwilkinson:
Okay, that last run looked correct. I've merged it, and now just need to get another build that matches...

PR #413, opened 2021-02-11
Us123463/bff polling

johngwilkinson:
- [ ] merge after metrics are available in the bff

Quality Notes
- functional (e2e)
  - load dashboard, adjust filters; confirm that polling is happening on metron api configured in LMS
  - reload; confirm that response comes directly
- security
  - d2lfetch always gets *:*:* tokens, which is not ideal, but standard for d2l
- perf
  - not a big change, but polling interval means data arrives either right away, after 250ms, after 750ms, etc.
- devops: requires up-to-date, web-facing (e.g. dev-tunnel-ed) LMS with metron pointing at up-to-date BDP stack
- ux/docs: n/a

FYI @MykolaGalian @NicholasHallman


model/lms.js
```diff
@@ -12,7 +12,7 @@ export function restore() {
```
bemailloux, line 12: This file probably should be called something else now

test/model/lms.test.js
```diff
@@ -316,27 +316,51 @@ describe('Lms', () => {
...
 		});
+
+		it('should poll for data', async() => {
+			const dataUri = 'begin:https://data.example.com/unstable/insights/data/engagement';
+			const expected = { the: 'data' };
+			fetchMock.post('path:/d2l/lp/auth/oauth2/token', {
+				'expires_at': Date.now() + 100000,
+				'access_token': 't'
+			});
+			fetchMock.get(dataUri, 202, { repeat: 2 });
```
bemailloux, line 340: neat!

bemailloux (approved):

model/lms.js
johngwilkinson: Agreed: https://github.com/Brightspace/insights-engagement-dashboard/pull/416

model/lms.js
```diff
@@ -36,7 +38,17 @@ export async function fetchData({ roleIds = [], semesterIds = [], orgUnitIds = [
 		url.searchParams.set('selectedOrgUnitIdsCsv', orgUnitIds.join(','));
 	}
 	url.searchParams.set('defaultView', defaultView ? 'true' : 'false');
-	const response = await fetch(url.toString());
+	const uri = url.toString();
+
+	let response = await d2lfetch.fetch(uri, { headers: { 'cache-control': 'no-store' } });
+	let waitMs = 0;
+	while (response.status === 202) {
+		if (waitMs < 2000) waitMs += 250;
```
Vitalii-Misechko, line 46: `thoughts`: Should we add some random part into 250 to avoid overcrowding after a sudden Aws lambda outage? I've heard about the idea of a random component in backoff algorithms but have never used it.

Vitalii-Misechko (approved):
LGTM!

model/lms.js
johngwilkinson: Good question! The usual reasons for randomizing don't apply here because this isn't a retry due to an outage: there's no reason why a number of clients would all get a 202 at the same time, unless they sent the first request at the same time, which would only happen if the end users all loaded the page at the same time.

PR #444, opened 2021-02-25
Use intl library to get decimal and thousands separator symbols

bemailloux:
Originally we tried to have these symbols "translated", but the translators didn't understand what we wanted to do, and did nothing. (Except in zh where they used the _wrong_ symbol.)

Solution: use the intl library instead.

Quality notes:
* Testing: see screenshots
* Devops: this will be going into 20.21.03 as well
* Perf, security, UX, docs: N/A

Screenshots (demo page)
* English
image

* German
image

* French
image

* Spanish
image

* Spanish (Mexico) - same symbols as English
image

* Chinese
image

@NicholasHallman @anhill-D2L

### DIFF
--- a/components/chart/chart.js
+++ b/components/chart/chart.js
@@ -6,6 +6,7 @@ import '../overlay';
 
 import { css, html, LitElement } from 'lit-element';
 import { getDateTimeDescriptor } from '@brightspace-ui/intl/lib/dateTime';
+import { getNumberDescriptor } from '@brightspace-ui/intl/lib/number';
 import { Localizer } from '../../locales/localizer';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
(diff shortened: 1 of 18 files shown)


johngwilkinson (approved):

components/chart/chart.js
```diff
@@ -151,6 +152,9 @@ class Chart extends SkeletonMixin(Localizer(LitElement)) {
 					screenReaderSection: {
 						// fixes axe error: Landmarks must have a unique role or role/label/title
 						beforeRegionLabel: ''
+					},
+					zoom: {
+						resetZoomButton: this.localize('chart:resetZoom')
```
bemailloux, line 157: I noticed one more small thing: the aria-label for the zoom button wasn't localized. This fixes it

anhill-D2L (approved):

PR #451, opened 2021-02-25
Fix of a view of empty trend chart when there is no data for it

Vitalii-Misechko:
Fix of a view of empty trend chart when there is no data for it

* Functional Testing 
  * An empty trend chart has ticks for y-axis even there is no data
  * A trend chart shows correct ticks when trend data is provided
  * Filters correctly hides/shows trend data
  
* Security, devops, perf: N/A
* UX and docs
  * [x] Demo
  
FYI @anhill-D2L @hyehayes @johngwilkinson @NicholasHallman @bemailloux @MykolaGalian 

### Before

image

### After

image

### Demo

Ticks-Demo-For-Empty-NonEmpty-chart


components/access-trend-card.js
```diff
@@ -272,8 +289,18 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+
+	_onChartLoad(event) {
+		this._chart = event.detail;
+	}
+
 	get _series() {
-		if (!this.data._data) return [];
+		if (this._emptyData) return [0, 0];
+
+		if (this._chart) this._chart.redraw(); // update tickPositions for the chart with data
```
johngwilkinson, line 302: Why is this necessary, and could it go in `render` instead of in a getter? Seems dangerous to rely on a side-effect here.

bemailloux (commented):
What happens if you load the chart with data initially, then change the filters such that the chart has no data? Does it show the empty grid in that case?

components/access-trend-card.js
Vitalii-Misechko: Good point!  Removed, it works without calling `redraw`

Vitalii-Misechko:
> What happens if you load the chart with data initially, then change the filters such that the chart has no data? Does it show the empty grid in that case?

Unfortunately, it doesn't. It shows grid with a previous series. It seem one more bug (

image

Looking into it

Vitalii-Misechko:
> > What happens if you load the chart with data initially, then change the filters such that the chart has no data? Does it show the empty grid in that case?
> 
> Unfortunately, it doesn't. It shows grid with a previous series. It seem one more bug (
> 
> image
> 
> Looking into it

Fixed

components/access-trend-card.js
```diff
@@ -272,8 +289,16 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return orgUnit ? orgUnit[ORG_UNIT.NAME] : '';
 	}
 
+	get _emptyData() {
+		return !this.data._data || this._trendData.length === 0;
+	}
+
+	_onChartLoad(event) {
+		this._chart = event.detail;
```
johngwilkinson, line 297: is `this._chart` still used anywhere?

bemailloux (approved):

components/content-views-card.js
```diff
@@ -248,8 +254,16 @@ class ContentViewsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return orgUnit ? orgUnit[ORG_UNIT.NAME] : '';
 	}
 
+	get _emptyData() {
+		return !this.data._data || !this.userData.contentViews || this._trendData.length === 0;
+	}
+
+	_onChartLoad(event) {
+		this._chart = event.detail;
```
johngwilkinson, line 262: used?

johngwilkinson (commented):
At least one of the new goldens in this PR is bad (empty cards). I also notice the y-axis ticks look very close together in the new ones - is that intentional?

components/access-trend-card.js
Vitalii-Misechko: nope, can be deleted

Vitalii-Misechko:
> At least one of the new goldens in this PR is bad (empty cards). 
I'll take a look. I considered this empty card as some kind of flakyness.

> I also notice the y-axis ticks look very close together in the new ones - is that intentional?

Removed `intervalTick: 1`, it is supposed to prevent fractional ticks on Course Access and Content View trend charts. But, it's better to discuss all pros and cons of the solution with Kevin first.

components/content-views-card.js
Vitalii-Misechko: removed

johngwilkinson (approved):

Vitalii-Misechko:
> At least one of the new goldens in this PR is bad (empty cards).

I've just gone through golden snapshots and I haven't found a bad card. Either the golden is fixed or I overlooked it.

Could you point me out if the bad snapshot exists?

johngwilkinson:
> > At least one of the new goldens in this PR is bad (empty cards).
> 
> I've just gone through golden snapshots and I haven't found a bad card. Either the golden is fixed or I overlooked it.
> 
> Could you point me out if the bad snapshot exists?

The ones you just merged look fine. They are different than the ones that were here before.

Vitalii-Misechko:
I'll port these changes to the master tomorrow if there are no objections )

johngwilkinson:
In general: for flaky tests, we shouldn't merge the wrong goldens, because then our tests are trying to prove that our code produces those results.

Vitalii-Misechko:
> In general: for flaky tests, we shouldn't merge the wrong goldens, because then our tests are trying to prove that our code produces those results.

I agree with that. I just didn't figure out what to do with such tests. Re-run or close visual diff PR and re-run it?

johngwilkinson:
> > In general: for flaky tests, we shouldn't merge the wrong goldens, because then our tests are trying to prove that our code produces those results.
> 
> I agree with that. I just didn't figure out what to do with such tests. Re-run or close visual diff PR and re-run it?

Just re-run the github jobs, I think.

PR #467, opened 2021-03-02
US123807: link to settings from user drill

johngwilkinson:
Quality Notes

- functional
  - test using demo page (had to run `npm install BrightspaceUILabs/multi-select#semver:^4` for it to work)
  - visit settings page from main page
  - visit settings page from drill page
  - navigate back using browser back button (both cases)
  - navigate back using immersive nav (both cases)
    - note that the nav link goes back to the main page regardless - changing this may be out of scope
- security, perf, devops: n/a
- ux: will demo to Kevin later with other changes


NicholasHallman (approved):

package.json
```diff
@@ -45,6 +45,7 @@
   "dependencies": {
     "@adobe/lit-mobx": "0.0.x",
     "@brightspace-ui-labs/facet-filter-sort": "^4",
+    "@brightspace-ui-labs/multi-select": "github:BrightspaceUILabs/multi-select#semver:^4",
```
bemailloux, line 48: What is this needed for?

johngwilkinson: I'm not sure why, but the dev server/demo page couldn't find this reference otherwise. It's a dependency of facet-filter-sort. Maybe github refs don't work in dev server when nested.

bemailloux: Maybe it could be because `node_modules` in either engagement-dashboard or BSI was out of date?

johngwilkinson: I'm talking about the demo page - no BSI involved. And I ran `npm ci`, and also `npm i`.

bemailloux (approved):

PR #469, opened 2021-03-03
US123807: user drill settings

johngwilkinson:
Need to get some decisions and assets, but sharing now in case it is relevant to other work in this story.

FYI @bemailloux @Vitalii-Misechko @NicholasHallman

### DIFF
--- a/components/user-drill-view.js
+++ b/components/user-drill-view.js
@@ -47,7 +47,11 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 			isStudentSuccessSys: { type: Boolean, attribute: false },
 			orgUnitId: { type: Number, attribute: 'org-unit-id' },
 			viewState: { type: Object, attribute: false },
-			metronEndpoint: { type: String, attribute: 'metron-endpoint' }
+			metronEndpoint: { type: String, attribute: 'metron-endpoint' },
+			showAverageGradeSummaryCard: { type: Boolean, attribute: 'average-grade-summary-card', reflect: true },
+			showContentViewsTrendCard: { type: Boolean, attribute: 'content-views-trend-card', reflect: true },
+			showCourseAccessTrendCard: { type: Boolean, attribute: 'course-access-trend-card', reflect: true },
+			showGradesTrendCard: { type: Boolean, attribute: 'grades-trend-card', reflect: true }
 		};
 	}
 
--- a/engagement-dashboard.js
+++ b/engagement-dashboard.js
@@ -72,7 +72,11 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 			showTicCol: { type: Boolean, attribute: 'tic-col', reflect: true },
 			showTicGradesCard: { type: Boolean, attribute: 'tic-grades-card', reflect: true },
 			lastAccessThresholdDays: { type: Number, attribute: 'last-access-threshold-days', reflect: true },
-			includeRoles: { type: String, attribute: 'include-roles', reflect: true }
+			includeRoles: { type: String, attribute: 'include-roles', reflect: true },
+			showAverageGradeSummaryCard: { type: Boolean, attribute: 'average-grade-summary-card', reflect: true },
+			showContentViewsTrendCard: { type: Boolean, attribute: 'content-views-trend-card', reflect: true },
+			showCourseAccessTrendCard: { type: Boolean, attribute: 'course-access-trend-card', reflect: true },
+			showGradesTrendCard: { type: Boolean, attribute: 'grades-trend-card', reflect: true }
 		};
 	}
 
@@ -102,6 +106,10 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		this.showTicGradesCard = false;
 		this.lastAccessThresholdDays = 14;
 		this.includeRoles = '';
+		this.showAverageGradeSummaryCard = false;
+		this.showContentViewsTrendCard = false;
+		this.showCourseAccessTrendCard = false;
+		this.showGradesTrendCard = false;
 
 		this._viewState = new ViewState({});
 		// if current view is not provided in url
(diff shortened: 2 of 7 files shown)


engagement-dashboard.js
```diff
@@ -265,6 +273,10 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 				.viewState="${this._viewState}"
 				.metronEndpoint="${this.metronEndpoint}"
 				@d2l-insights-user-drill-view-back="${this._backToHomeHandler}"
+				?average-grade-summary-card="${this.showAverageGradeSummaryCard}"
+				?grades-trend-card="${this.showGradesTrendCard}"
+				?course-access-trend-card="${this.showCourseAccessTrendCard}"
+				?content-views-trend-card="${this.showContentViewsTrendCard}"
```
bemailloux, line 279: I like your names better than mine. They're more concise and aren't necessarily tied to userDrill. I'm going to change my server-side PR to match.

components/user-drill-view.js
```diff
@@ -444,60 +461,79 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			.data="${this.data}"
+			.user="${this.user}"
+			.userData="${this._userData}"
+			.selectedCourses="${this.selectedCourses}"
+			?demo="${this.isDemo}"
+		></d2l-insights-access-trend-card>`;
+	}
+
+	get _legend() {
+		if (!this.showGradesTrendCard && !this.showContentViewsTrendCard && !this.showCourseAccessTrendCard) return '';
```
NicholasHallman, line 528: `nit` I think we should use the `nothing` directive from lit-html instead of ''

locales/en.js
```diff
@@ -184,6 +186,11 @@ export default {
 	"settings:invalidSystemAccessValueToast": "Your settings could not be saved. System Access thresholds need to be between 1 and 30.",
 	"settings:serverSideErrorToast": "Something went wrong. Your settings could not be saved.",
 
+	"settings:avgGradeSummaryDescription": "The Average Grade Summary card presents the current average grade for the student across courses that are filtered on.",
+	"settings:gradesOverTimeDescription": "The Grades Over Time card shows how the learner's grades have changed over time.",
+	"settings:accessOverTimeDescription": "The Course Access Over Time card shows how the learner’s course access frequency has changed over time.",
```
bemailloux, line 191: Huh, this line has a different apostrophe than the usual one.

components/svg/average-grade-summary.svg.js
```diff
@@ -0,0 +1,31 @@
...
+						<rect width="293" height="150" rx="15" stroke="none"/>
+						<rect x="0.5" y="0.5" width="292" height="149" rx="14.5" fill="none"/>
+					</g>
+					<rect id="Rectangle_384"  width="248" height="107" transform="translate(9 28)" fill="#fff" opacity="0.003"/>
+					<text id="Discussion_Activity"  transform="translate(9 9)" fill="#494c4e" font-size="14" font-family="Lato-Bold, Lato" font-weight="700" letter-spacing="0.014em"><tspan x="0" y="14">${this.localize('averageGradeSummaryCard:averageGrade')}</tspan></text>
+					<text id="_21"  transform="translate(9 66.6)" fill="#6c6c6c" font-size="20" font-family="Lato-Bold, Lato" font-weight="700" letter-spacing="0.01em"><tspan x="10.018" y="20">${gradeText}</tspan></text>
+					<text id="learners_did_not_par"  transform="translate(102.352 60.74)" fill="#494c4e" font-size="14" font-family="Lato-Regular, Lato" letter-spacing="0.014em"><tspan x="0" y="14">${this.localize('averageGradeSummaryCard:averageGradeTextLine1')}</tspan><tspan x="0" y="32">${this.localize('averageGradeSummaryCard:averageGradeTextLine2')}</tspan></text>
```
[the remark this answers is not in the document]
johngwilkinson: Will test, thanks

components/svg/course-access-trend.svg.js
```diff
@@ -0,0 +1,69 @@
...
+						</g>
+						<g id="X_axis"  transform="translate(24 26.174)" opacity="0">
+							<text id="_100"  transform="translate(6.188 5)" fill="#72777a" font-size="5" font-family="Lato-Regular, Lato" letter-spacing="0.014em"><tspan x="0" y="0">800</tspan></text>
+							<text id="_60"  transform="translate(6.168 23.214)" fill="#72777a" font-size="5" font-family="Lato-Regular, Lato" letter-spacing="0.014em"><tspan x="0" y="0">600</tspan></text>
+							<text id="_40"  transform="translate(5.727 61.42)" fill="#72777a" font-size="5" font-family="Lato-Regular, Lato" letter-spacing="0.014em"><tspan x="0" y="0">200</tspan></text>
+							<text id="_0"  transform="translate(13.237 81.078)" fill="#72777a" font-size="6" font-family="Lato-Regular, Lato" letter-spacing="0.014em"><tspan x="0" y="0">0</tspan></text>
+							<text id="Current_Final_Grade-2"  transform="translate(4 54.626) rotate(-90)" fill="#494c4e" font-size="4" font-family="Lato-Bold, Lato" font-weight="700" letter-spacing="0.014em"><tspan x="0" y="0">Access Counts</tspan></text>
```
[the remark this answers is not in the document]
johngwilkinson: None of this text actually shows up, and these are hidden from screen-readers. I guess I could go in and clean it out, but I don't want to mess anything up with manual editing.

components/user-drill-courses-table.js
```diff
@@ -236,6 +251,11 @@ class UserDrillCoursesTable extends SortMixin(SkeletonMixin(Localizer(MobxLitEle
 					headerText: this.localize('inactiveCoursesTable:grade'),
 					columnType: COLUMN_TYPES.NORMAL_TEXT
 				},
+				{
+					// placeholder: predicted grade: will never be included in inactive table
```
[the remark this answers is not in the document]
johngwilkinson: I believe so, because this column does get rendered during export (to match the active table).

locales/en.js
johngwilkinson: Must have copied it from the design. Will fix.

components/user-drill-view.js
johngwilkinson: What's your thinking on this? The lit-html docs say that `''` is almost always what you want - the only exception they mention is if you want to render nothing at all so as not to override default content in a slot.

bemailloux (approved):

components/user-drill-view.js
NicholasHallman: On rereading I don't think it matters. I think the original discussion went something along the lines of `nothing` is better than `''` because with `''` lit renders a node but with `nothing` it just skips it. When you have as few nodes as we do the difference is probably meaningless

NicholasHallman (approved):

PR #472, opened 2021-03-03
Toast Alerts and Axe Description | US121167

NicholasHallman:
TODO
- [x] localization strings
- [x] testing

[WIP]

Quality Notes

Testing
  - [x] Clicking on any chart or selector creates an alert
  - [x] Only two alerts are visible at a time (third alert disappears after 40 ms)
  - [x] All alerts have a axe description (except for selector alerts)
  - [x] Groups of ranges are reduced in axe description for grades and time in content  

Security - N/A
Performance - N/A
UX 
  - [x] Demoed to Kevin. 
  - [x] Demo during standup
  
  FYI: @anhill-D2L @hyehayes @Vitalii-Misechko


components/alert-data-update.js
```diff
@@ -0,0 +1,186 @@
...
+		this._finished = new Promise(res => this._defer = res);
+		this.closing = false;
+	}
+
+	get message() {
+
+		if (!this._isRendered) {
+			this._isRendered = true;
+			this._killTimeout = setTimeout(() => {
+				this._defer(); // resolve the promise, removing the event from the queue
```
johngwilkinson, line 28: maybe rename this so the comment isn't needed - `this._remove()` or something like that

components/applied-filters.js
```diff
@@ -70,22 +71,33 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+					label-text="${this.localize('appliedFilters:labelText')}">
+				</d2l-labs-applied-filters>`
+			: html``
+		}
 		`;
 	}
 
 	_filterChangeHandler(event) {
+
+		// const filter = this.data.filters.find(f => ({
```
johngwilkinson, line 86: remove commented code

components/course-last-access-card.js
```diff
@@ -225,6 +227,40 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		if (categories.length === 0) return this.localize('alert:axeNotFiltering', chartName);
+
+		const pairs = categories.sort().reverse().reduce((acc, cur) => {
+			// if we can find the cur in a pair then we have a chain
+			const desc = DATA_DESCRIPTIONS[cur];
+			if (typeof(desc) === 'string') {
+				acc.push([desc]);
+				return acc;
+			}
+			const pair = acc.find((pair) => pair[1] === desc[0]);
```
johngwilkinson, line 250: since you've sorted above, wouldn't the match always be at the end of the array?

components/course-last-access-card.js
```diff
@@ -225,6 +227,40 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			DATA_DESCRIPTIONS[1] = this.localize('courseLastAccessCard:accessibilityMoreThanFourteenDaysAgo');
+			DATA_DESCRIPTIONS[6] = this.localize('courseLastAccessCard:accessibilityLessThanOne');
+		}
+		// bin the ranges of numbers together
+		const categories = ([...this.filter.selectedCategories]);
+
+		const chartName = { chartName : this.localize('courseLastAccessCard:courseAccess') };
+		if (categories.length === 0) return this.localize('alert:axeNotFiltering', chartName);
+
+		const pairs = categories.sort().reverse().reduce((acc, cur) => {
```
johngwilkinson, line 243: this looks like a good candidate for a named method so its purpose is easier to see

components/course-last-access-card.js
```diff
@@ -225,6 +227,45 @@ class CourseLastAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			return acc;
+		}, []);
+	}
+
+	getAxeDescription() {
+
+		if (DATA_DESCRIPTIONS[0] === '') {
+			DATA_DESCRIPTIONS[0] = this.localize('courseLastAccessCard:never');
+			DATA_DESCRIPTIONS[1] = this.localize('courseLastAccessCard:accessibilityMoreThanFourteenDaysAgo');
+			DATA_DESCRIPTIONS[6] = this.localize('courseLastAccessCard:accessibilityLessThanOne');
```
bemailloux, line 255: Seems weird to be modifying a variable which should be (in spirit, if not literally) constant. But I also understand that you don't have access to the localization until `render`.

Can this go somewhere else that's only run once, like `firstUpdated` maybe? Or do something else, like putting strings and arrays in different arrays?

locales/en.js
```diff
@@ -236,4 +243,12 @@ export default {
 	"chart:loading": "Loading...",
 	"chart:resetZoom": "Reset zoom",
 	"chart:resetZoomTitle": "Reset zoom level 1:1",
+
+	"alert:updatedFilter" : "{chartName} filter updated.",
+	"alert:axeDescriptionRange" : "Viewing learners with {chartName} in these categories ",
+	"alert:axeDescription" : "Viewing learners with ",
+	"alert:axeNotFiltering" : "Stopped Filtering by {chartName}.",
+	// this "to" that, and this "to" that <- translate the word "to" in this context
```
bemailloux, line 251: It's possible the translation software might choke on a comment in the middle of the json. Hopefully not

bemailloux (commented):
What does a screen reader say when multiple alerts are displayed at the same time? Does it cut off whatever it was saying to say the new thing, or does it say everything sequentially?

components/access-trend-card.js
```diff
@@ -80,6 +81,10 @@ class AccessTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return this.localize('accessTrendCard:yAxisTitle');
 	}
 
+	get axeDescription() {
+		return this.parentElement.parentElement.querySelector('d2l-insights-courses-legend').getAxeDescription();
```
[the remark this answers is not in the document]
NicholasHallman: Yeah I will create a context object the child can tap into instead

components/course-last-access-card.js
NicholasHallman: I could put the names of the localizations into the array and then call localize later

components/alert-data-update.js
```diff
@@ -0,0 +1,186 @@
...
+		return this._message;
+	}
+
+	get description() {
+		return this._description;
+	}
+
+	get id() {
+		if (this._id === '') {
+			this._id = new Array(10).fill(0).map(() =>  Math.round(Math.random() * 9)).join('');
```
bemailloux, line 41: Maybe we should just import a guid library or something?

components/alert-data-update.js
```diff
@@ -0,0 +1,186 @@
...
+		];
+	}
+
+	constructor() {
+		super();
+		this.dataEvents = {};
+		this.lastFilters = [];
+	}
+
+	_render_event(e) {
```
bemailloux, line 160: snake_case => camelCase

components/content-views-card.js
```diff
@@ -97,6 +99,10 @@ class ContentViewsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 		}
 	}
 
+	get axeDescription() {
+		return context.get('course-legend').getAxeDescription();
```
bemailloux, line 103: Nice :)

bemailloux (approved):

components/alert-data-update.js
NicholasHallman: These id's are only for timing the animation and only three of them are needed at a time. I was thinking of adding uuidv4 but it seems overkill when the impact of an id collision here means that the animations are a little loopy for 200 milliseconds

components/alert-data-update.js
NicholasHallman: too much lua in my spare time 😅

NicholasHallman:
@johngwilkinson Yeah that's a good point. I'm trying to mimic this [popular react pattern](https://reactjs.org/docs/context.html)  (which data would probably also benefit from since we pass it down nearly every component) 

> Can you instead put a getAxeDescription on SelectedCourses and inject that object directly into the components that need it (via property)?

Yes... but that means copying the courses from `data` into `SelectedCourses` and updating them anytime we call the server. We would need to do this because `SelectedCourses` doesn't store the names, just the id's and so it can't create the description on it's own.

components/courses-legend.js
```diff
@@ -8,6 +8,39 @@ import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 import { UrlState } from '../model/urlState';
 
+export class CoursesHelper {
```
NicholasHallman, line 11: I realized that all of the components had access to the same data, so migrating the few function I needed from the legend into it's own helper was the key code reuse pattern here

johngwilkinson (approved):

PR #477, opened 2021-03-05
Concept of code design that depends on requested-data

Vitalii-Misechko:
I just wanted to share with you an idea of where we could put our code that fetches data for any MobxLitElement-descendent. I come up with this idea accidentally when realized that the current implementation has many "nested" render loops. I mean I observe that render-function calls a fetch which in turn calls render. The proposed design allows decoupling code that causes rendering from the code that just executes rendering preferably without side effects.

FYI @anhill-D2L @hyehayes @johngwilkinson @NicholasHallman @bemailloux @MykolaGalian

### DIFF
--- a/components/user-drill-view.js
+++ b/components/user-drill-view.js
@@ -11,7 +11,7 @@ import './summary-card';
 import './access-trend-card';
 
 import { bodySmallStyles, heading2Styles, heading3Styles } from '@brightspace-ui/core/components/typography/styles.js';
-import { computed, decorate } from 'mobx';
+import { computed, decorate, reaction } from 'mobx';
 import { css, html } from 'lit-element/lit-element.js';
 import { RECORD, USER } from '../consts';
 import { createComposeEmailPopup } from './email-integration';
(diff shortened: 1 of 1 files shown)


components/user-drill-view.js
```diff
@@ -367,14 +367,38 @@ class UserDrill extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return false;
 	}
 
-	render() {
-		if (this.newFilteredOrgUnitIds.some(id => !this.lastFilteredOrgUnitIds.includes(id))) {
-			this.lastFilteredOrgUnitIds = this.newFilteredOrgUnitIds;
-			if (this.lastFilteredOrgUnitIds.length !== 0) {
+	firstUpdated() {
```
johngwilkinson, line 370: I think I don't understand the purpose of this. It doesn't seem like it should be necessary:
- in lit-element, property updates during `render` do not cause a re-render: https://lit-element.polymer-project.org/guide/lifecycle#render
- in mobx, observables modified in an `action` cause at most one run of reactions

Vitalii-Misechko: > I think I don't understand the purpose of this. It doesn't seem like it should be necessary:
> * in lit-element, property updates during render do not cause a re-render: https://lit-element.polymer-project.org/guide/lifecycle#render

I agree that the codebase is okay without this proposal and Lit-element works great. My initial concern is related to the MobX of `UserDrill`-class. 

> * in mobx, observables modified in an action cause at most one run of reactions

As I mentioned in PR description I have observed nested loops(additional calls of render function) while rendering `UserDrill`-class. It turns out that `Data.loadData` is not an action. It causes two renderings. The first execution is caused by this code `this.onServerDataReload(data)` and the second one when `Data.loadData`-function finish its work.

It turns out [that ](https://doc.ebichu.cc/mobx/best/actions.html)

> The action wrapper / decorator only affects the currently running function, not functions that are scheduled (but not invoked) by the current function! This means that if you have a setTimeout, promise. then or async construction, and in that callback, some more state is changed, those callbacks should be wrapped in action as well! 

Maybe it is sufficient to just properly declare `Data.loadData` and `UserData.loadData` to avoid these additional renderings. But at the time of creating this PR I didn't know what was the root cause of them. On top of that, I thought this is a good idea to decouple the rendering and side effects which it caused.

johngwilkinson: I do think the best approach is to rely on correct annotations as much as possible. It seems that the side-effect of loading data is already decoupled because it runs asynchronously.

PR #489, opened 2021-03-10
Content View Histogram

NicholasHallman:
Won't be merged until the next sprint.

content  #view

Implementation Notes

Binning
The binning technique is pretty simple. We find the max value in the data and then find the smallest bin grouping it fits into. A max value of 15 fits into the 50 bin group, a value of 700 fits into the 1000 bin group etc.
These are the available bin schems.

0-10-20-30-40-50
0-20-40-60-80-100
0-40-80-120-160-200
0-100-200-300-400-500
0-200-400-600-800-1000

Binning outliers.

Occasionally you may have a situation where many learners visit content rarely but the instructor or otherwise visits regularly skewing the data. To avoid all of the learners getting grouped into the bottom bin and the teacher at the top we implement the following.

Find the [IQR whisker](https://en.wikipedia.org/wiki/Interquartile_range) of the data, giving us an upper bound to filter out the outliers.
Find the new max value (of the data minus the outliers) to pick a bin.
Reintroduce the outliers in a `greater than` bin

Quality Notes

Functional Testing
  - [x] A users total content view is binned into a bar on chart (manual)
  - [x] The chart responds to other charts filters (manual)
  - [x] The tooltip reads the number of users in the row (manual)
  - [x] Screen reader knows the title of the chart and the description of the selected bar (manual, windows narrator)
  - [x] Null data is excluded, doesn't crash chart (manual)
  - [x] Constructs (auto)
  - [x] Accessible (auto)
  - [x] Bin by the various different binning configurations (auto)
  - [x] Make sure that values > 1000 get binned into the grater than bin (auto)
  - [x] Don't use the outlier data to pick the bins (auto)
  
Performance
  - Iterates through the user records. Should not be a problem.
 
UX
  - [x] Kevin Demo
  - [ ] Standup Demo
  
Security
  - N/A
  
Dev-Ops
  - N/A

Cost
  - N/A
  
Documentation
  - N/A 
  
 FYI @Vitalii-Misechko @johngwilkinson @hyehayes


model/fake-dataApiClient.js
```diff
@@ -18,24 +18,24 @@ function getDelayFromUrlHash() {
...
+			[1, 200, 700, 1, 71, 4000, null, 0, 0, 0, null, 4],
+			[1, 100, 700, 1, 81, 1000, null, 3, 2, 1, null, 12],
+			[2, 100, 700, 1, 91, 1200, null, 1, 33, 1, null, 3],
+			[2, 300, 500, 0, 9, 7200, null, 1, 3, 5, null, 22],
+			[2, 300, 700, 0, 3, 0, 289298332, 0, 0, 0, null, 55],
+			[2, 400, 700, 0, 100, 7200, Date.now() - 432000001, 0, 0, 0, 0.9, 2],
+			[2, 500, 700, 0, 88, 4000, null, 4, 4, 1, null, 7],
+			[8, 200, 700, 0, null, 0, null, 55, 2, 3, null, 9],
+			[6, 600, 700, 0, 95, 2000, Date.now() - 8560938122, 0, 0, 0, null, 12],
+			[1, 400, 700, 1, 75, 2000, null, 2, 1, 4, null, 44]
```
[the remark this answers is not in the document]
NicholasHallman: Good catch! I put the data in the wrong place first time around. Forgot to delete it.

anhill-D2L:
Where would a user be placed if they have more than 200 content views?

NicholasHallman:
> Where would a user be placed if they have more than 200 content views?

I sent Kevin a message about this. I have a feeling that that is one of the cases where the binning would change on the fly

bemailloux (commented):
Looks good so far, just waiting for dynamic bins implementation.

test/components/content-view-histogram.test.js
```diff
@@ -21,4 +27,113 @@ describe('content-view-histogram', () => {
 			await expect(el).to.be.accessible();
 		});
 	});
+
+	describe('binning', () => {
```
anhill-D2L, line 31: Does anything weird happen with the binning if you have only one, or zero users that have content views in a course?

NicholasHallman: It breaks!

components/content-view-histogram.js
```diff
@@ -0,0 +1,292 @@
...
+				acc.push(`> ${cur[1]}`);
+			} else {
+				acc.push(`${cur[0]}-${cur[1] + 1}`);
+			}
+			return acc;
+		}, []).concat(['0']);
+	}
+
+	get _chartDataBuckets() {
+		// buckets go from top to bottom, 200 views to 0
```
anhill-D2L, line 160: I think this comment is outdated now

components/content-view-histogram.js
```diff
@@ -0,0 +1,294 @@
...
+		if (values.length === 0) upperBin = 50;
+		const range = upperBin / 5;
+		const bins = [];
+
+		for (let i = 1; i <= 5; i++) {
+			bins.push([range * i, range * (i - 1)]);
+		}
+
+		if (this.courseAccessOutliers.length !== 0 || largestAccess > 1000) {
+			bins.push([Number.POSITIVE_INFINITY, upperBin]);
```
anhill-D2L, line 91: Do we need a different tooltip for this case?

NicholasHallman: Yes it does! How could I forget

anhill-D2L (approved):

bemailloux:
> Binning outliers.
> Occasionally you may have a situation where many learners visit content rarely but the instructor or otherwise visits regularly skewing the data. To avoid all of the learners getting grouped into the bottom bin and the teacher at the top we implement the following.
> Find the IQR whisker of the data, giving us an upper bound to filter out the outliers.
> Find the new max value (of the data minus the outliers) to pick a bin.
> Reintroduce the outliers in a greater than bin

This is smart! Do you have a screenshot of this case occurring? I'd like to see what it looks like.

components/content-view-histogram.js
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+		return bins.reverse();
+	}
+
+	_median(a) {
+		const result = a[Math.floor((a.length - 1) / 2)];
+		return result;
+	}
```
bemailloux, line 104: `Just a thought` Maybe the stats stuff should go into a utils library? It's broadly applicable and I don't think it needs to be taking up space in this class.

components/content-view-histogram.js
```diff
@@ -0,0 +1,298 @@
...
+
+	get courseAccesses() {
+		return this.data
+			.withoutFilter(filterId)
+			.users.map(record => record[USER.TOTAL_COURSE_ACCESS]).sort((a, b) => a - b)
```
bemailloux, line 109: `courseAccesses` is called in a couple of different places, each time redoing the sort. Should the value be memoized somewhere? Or maybe mobx-`computed`?

components/content-view-histogram.js
```diff
@@ -0,0 +1,298 @@
...
+			const totalCount = record[USER.TOTAL_COURSE_ACCESS];
+			if (totalCount === 0) return this.bins.length;
+			return this.bins.findIndex(bin => totalCount <= bin[0] && totalCount > bin[1]);
+		};
+
+		const buckets = [...this.dataBuckets];
+
+		return this.data
+			.withoutFilter(filterId)
+			.users
```
bemailloux, line 178: Looks like you could use `courseAccesses` here instead of redoing `withoutFilter().users`?

test/components/content-view-histogram.test.js
bemailloux: ^ Looks like there should also be a test case for this but I don't see one.

components/content-view-histogram.js
```diff
@@ -0,0 +1,298 @@
...
+		return html`
+		<div class="d2l-insights-content-views-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this._cardTitle}</div>
+		<d2l-labs-chart
+			.options="${this.chartOptions}"
+			?skeleton="${this.skeleton}"
+		></d2l-labs-chart>`;
+	}
+}
+
+customElements.define('d2l-labs-content-view-histogram', ContentViewHistogram);
```
Vitalii-Misechko, line 298: Should we use `d2l-insights-content-view-card` as a tag to stay consistent with the names of other cards?

NicholasHallman: Yeah I was going to do that but we already have a content-views-card unfortunately

NicholasHallman:
> ^ Looks like there should also be a test case for this but I don't see one.
@bemailloux sorry I'm not sure I understand what you mean

NicholasHallman:
@bemailloux  Yes I do! so here is an example with no outlier.

no outlier 0-100

And here is an example where the largest point is 200, but all of the other points are < 100

with outlier 0-100

bemailloux:
> > ^ Looks like there should also be a test case for this but I don't see one.
> > @bemailloux sorry I'm not sure I understand what you mean

Sorry I meant unit test

components/content-view-histogram.js
bemailloux: Also, should `filter` happen before `sort`? That might make the sort a tiny bit faster

bemailloux:
A ">1000" bucket will appear if there are a few outliers with more than 1000 content views. But does a ">1000" bucket appear if there are a significant number or a majority of users with more than 1000 content views?

bemailloux:
A ">1000" bucket will appear if there are a few outliers with more than 1000 content views. But does a ">1000" bucket appear if there are a significant number or a majority of users with more than 1000 content views?

NicholasHallman:
> A ">1000" bucket will appear if there are a few outliers with more than 1000 content views. But does a ">1000" bucket appear if there are a significant number or a majority of users with more than 1000 content views?

The largest binning system goes up to 1000, so if all of our users somehow had content views > 1000 then yes, they all go in the top bin. If need be I can add a bin that goes up to 2000 or some other increment but I felt that 1000 was a good max

test/components/content-view-histogram.test.js
NicholasHallman: There is kind of. The test case for "show bins in range 50 with no data" was the fix for bad chart behavior with no data. One user worked fine before and so didn't need any change

bemailloux (approved):
LGTM!

PR #492, opened 2021-03-11
bugfix: tic midpoints

johngwilkinson:
Quality notes: quadrant toggling and tooltips work now; accessibility works to (tested with Narrator)


johngwilkinson:
It would be good to have tests for this stuff to avoid regressions - I'll think about another PR for that

bemailloux (approved):
Glad you caught this now, instead of just before cert done. 

[For retro] Maybe we should do some manual testing during the cert period before translations show up? That'll give us time to fix bugs.

NicholasHallman (approved):

PR #495, opened 2021-03-11
Us123809/users table perf

johngwilkinson:
Qualiity Notes
- functional
  - tested w. demo page: load and display canned data; sorting; export
  - tested e2e: load and display draco data; sorting; email popup populates correctly (given permissions)
- perf
  - Before: 3.6s on preprocessing and formatting user table data (for 50k users in 50k courses of canned data)
  - After: 300-400ms
  - Note that export still takes 10-15s (with no UI indication of work) due to date formatting. Using new Date(ts).toISOString() takes it down to a couple of seconds, for example, so we consider making a change to the export date format
- ux/devops/security: no change


johngwilkinson:
Note on what worked here:
- biggest improvement by far was doing date formatting only for the page we're actually rendering (e.g. 20 of 50k users)
- localization was also significant: every call to `localize` parses the resulting string as an AST for template substitution - even if it's not a template - so this is something to watch for in tight loops

Also note: other PRs incoming for other perf improvements, but this one has the biggest effect

bemailloux (approved):

NicholasHallman (approved):

PR #510, opened 2021-03-16
Content View Histogram Interactions | US123709

NicholasHallman:
Implementation Notes

This PR Includes
 - Color changes on bar on click
 - Toggle bars on click
 - Toggling all bars to "on" resets the filter
 - Toast alerts
 - Axe description
 - Bins are based on all data, not visible data, to make filtering possible
 - URLState 
 
Filtering for this chart is a little strange. Because this chart uses user records to display the total content views the filter converts the filter from a user record filter to a record filter. It does this by

1. finding the user record that maps to a record using the userDictionary
2. finding which bin that user fits into
3. filtering based on if that bin is selected or not

which is O(n). nice! Thanks @johngwilkinson for pointing out the userDictionary

Quality Notes

Functional Testing
 - [x] Merging of ranges for axe descriptions
 - [x] Filtering 
 - [x] Colors from filters

Performance - N/A

Security - N/A

Dev-Ops - N/A

Cost - N/A

UX
 - [ ] Demo during standup
 
Accessibility resources / checklists:
 - [x] Accessible descriptions from toast popups
 - [x] Accessible descriptions when filter is cleared

FYI: @hyehayes @anhill-D2L @Vitalii-Misechko


model/filteredData.js
```diff
@@ -50,7 +50,7 @@ export class FilteredData {
 
 	get records() {
 		const appliedFilters = this.filters.filter(f => f.isApplied);
-		return this._data.records.filter(r => appliedFilters.every(f => f.filter(r, this.userDictionary)));
+		return this._data.records.filter(r => appliedFilters.every(f => f.filter(r, this.userDictionary, this._data.serverData.users)));
```
johngwilkinson, line 53: No need for this that I can see: `this.userDictionary` has all the same data, and it's already in a dictionary for O(1) lookup by user id.

components/content-view-histogram.js
```diff
@@ -111,16 +200,36 @@ class ContentViewHistogram extends SkeletonMixin(Localizer(MobxLitElement)) {
 			.sort((aRecord, bRecord) => aRecord[USER.TOTAL_COURSE_ACCESS] - bRecord[USER.TOTAL_COURSE_ACCESS]);
 	}
 
-	get courseAccesses() {
+	get serverData() {
+		return this.data._data.serverData
```
johngwilkinson, line 204: Please don't reference private properties (`_`) from outside the owner. There are ways to get the users directly from `this.data`.

components/content-view-histogram.js
```diff
@@ -258,6 +430,12 @@ class ContentViewHistogram extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	render() {
+
+		// the filter depends on the bin scheme so we need to update it.
+		this.filter.binScheme = this.bins[0];
+		console.log(this.filter._binScheme)
```
johngwilkinson, line 436: remove logging

components/content-view-histogram.js
```diff
@@ -1,14 +1,100 @@
...
 import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
-import { USER } from '../consts';
+import { RECORD, USER } from '../consts';
+import { UrlState } from '../model/urlState';
+
 const filterId = 'd2l-insights-content-view-histogram';
 
+let userBinCache = {}
```
johngwilkinson, line 16: likely better to use a `Map`

components/content-view-histogram.js
```diff
@@ -1,14 +1,100 @@
...
+			undefined // set all later
+		);
+
+		this._binScheme = 50;
+
+
+		this._urlState = new UrlState(this);
+	}
+
+	set binScheme(scheme) {
```
johngwilkinson, line 51: suggestion: if this took `bins`, it could handle pulling out the relevant one _and_ calling `setAll` and clearing the bin cache - then you don't need to override `setAll`, I think.

components/content-view-histogram.js
```diff
@@ -1,14 +1,100 @@
...
+		if (scheme[0] !== Number.POSITIVE_INFINITY) {
+			this._binScheme = scheme[0];
+		} else {
+			this._binScheme = scheme[1];
+		}
+		userBinCache = {}
+	}
+
+	get bins() {
```
johngwilkinson, line 60: You can likely improve perf by making `binScheme` `observable` and making this property `computed`.

components/content-view-histogram.js
NicholasHallman: yes but those user records are filtered, I need the unfiltered records so the y axis of the chart stays consistent. I can also get them from data but I would need to add a .withoutFilter() for each filter I think...

johngwilkinson: > yes but those user records are filtered, I need the unfiltered records so the y axis of the chart stays consistent. I can also get them from data but I would need to add a .withoutFilter() for each filter I think...

I don't think they're filtered:
```
get userDictionary() {
		return this._data.userDictionary;
	}
```
which would be
```
this.userDictionary = new Map(newServerData.users.map(user => [user[USER.ID], user]));
```

NicholasHallman: Oh I thought you meant data.users which I was already using for filtered results. userDictionary looks very helpful for this thank you :)

components/content-view-histogram.js
```diff
@@ -268,10 +423,13 @@ class ContentViewHistogram extends SkeletonMixin(Localizer(MobxLitElement)) {
 }
 
 decorate(ContentViewHistogram, {
-	sortedUserRecords: computed,
-	courseAccessWithoutOutliers: computed,
+	filteredSortedUserRecords: computed,
+	//sortedUserRecords: computed,
```
NicholasHallman, line 427: having some trouble with mobx. It doesn't seem to track changes to userDictionary but it tracks fine to users...

johngwilkinson: `userDictionary` only changes on a server reload, right?

NicholasHallman: It does, but it's not re-computing when the loading is finished. I cheated this by adding this.data.users to the function, it works now

johngwilkinson: Oh, I see: `userDictionary` is not an `observable`. The best thing to do, I think, would be to make it a getter that returns a user map based on `serverData` (not `_serverData`, because it needs to reference the proxy), and decorate that getter with `computed`.

NicholasHallman: Works great thank you!

components/content-view-histogram.js
```diff
@@ -1,14 +1,77 @@
...
+
+	set bins(bins) {
+		this._bins = bins;
+		const scheme = bins[0];
+		if (scheme[0] !== Number.POSITIVE_INFINITY) {
+			this._binScheme = scheme[0];
+		} else {
+			this._binScheme = scheme[1];
+		}
+		super._all = new Set((new Array(bins.length + 1).fill(0).map((v, i) => i)));
```
bemailloux, line 52: Why are we setting `_all` on `super` and not `this` here?

components/content-view-histogram.js
```diff
@@ -1,14 +1,77 @@
...
 const filterId = 'd2l-insights-content-view-histogram';
 
+export class ContentViewHistogramFilter extends CategoryFilter {
+	constructor() {
+		const filterFunc = (record, userDictionary) => {
+			let recordBin = 0;
+			const userId = record[RECORD.USER_ID];
+			const userRecord = userDictionary.get(userId);
+			const views = userRecord[USER.TOTAL_COURSE_ACCESS];
+			if (views === 0) return this.bins.length;
```
bemailloux, line 23: I don't get this. If views === 0 shouldn't we be returning false?

components/content-view-histogram.js
```diff
@@ -1,14 +1,77 @@
...
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
-import { USER } from '../consts';
+import { UrlState } from '../model/urlState';
+
 const filterId = 'd2l-insights-content-view-histogram';
 
+export class ContentViewHistogramFilter extends CategoryFilter {
+	constructor() {
+		const filterFunc = (record, userDictionary) => {
+			let recordBin = 0;
```
bemailloux, line 19: This line isn't necessary, you can just declare `const recordBin = this.bins.findIndex(...);` lower down

components/content-view-histogram.js
```diff
@@ -1,14 +1,77 @@
...
+		};
+		super(
+			filterId,
+			'contentViewHistogram:title',
+			filterFunc,
+			'cvhf',
+			undefined // set all later
+		);
+
+		this._binScheme = 50;
```
bemailloux, line 35: Is _binScheme used for anything in this class?

components/content-view-histogram.js
NicholasHallman: Because _all is only used by the base class. 
I think `super_all !== this._all` but I could be wrong

components/content-view-histogram.js
NicholasHallman: If the user has no views then they are in the bottom bin (the 0 bin) which is always the last bin in the array. We don't store a bin for zero because all bin schemes have a 0 bin and we know where it is.

components/content-view-histogram.js
NicholasHallman: Ah yep thank you! This was from earlier when the recordBin was calculated between two if statements so I needed the variable to exist before the if.

components/content-view-histogram.js
NicholasHallman: Not anymore :) It was being used so the filter could determine the binning system. But no we just give it the whole bin.

components/content-view-histogram.js
bemailloux: But this is the filter function. Shouldn't it be returning a boolean?

NicholasHallman: OH lol this is what recordBin is for

bemailloux: That makes more sense :) :)

bemailloux (approved):

components/content-view-histogram.js
```diff
@@ -1,14 +1,72 @@
...
+			this.selectedCategories.clear();
+			return;
+		}
+		const categories = value.split(',').map(category => Number(category));
+		this.setCategories(categories);
+	}
+}
+
+decorate(ContentViewHistogramFilter, {
+	setAll: action
```
johngwilkinson, line 67: this is covered in super now, isn't it?

components/content-view-histogram.js
```diff
@@ -1,14 +1,72 @@
...
+			filterFunc,
+			'cvhf',
+			undefined // set all later
+		);
+
+		this._urlState = new UrlState(this);
+	}
+
+	get bins() {
+		return this._bins;
```
johngwilkinson, line 42: should `_bins` be initialized to something in the constructor?

also, should it be `observable`? if it's not, setting `filter.bins` might not trigger a re-render. In practice it might because of the change to `super._all`, but that's harder to reason about.

components/content-view-histogram.js
```diff
@@ -1,14 +1,72 @@
...
+	}
+
+	set bins(bins) {
+		this._bins = bins;
+		super._all = new Set((new Array(bins.length + 1).fill(0).map((v, i) => i)));
```
johngwilkinson, line 47: shouldn't this call `this.setAll`?

NicholasHallman:
Lots of changes in the last commit so let me take a minute to talk through the problem and how I solved it.

My initial assumption was that the chart would load, watch the data and wait for the userDictionary to be populated and then give the filter the binning scheme so that it could filter the data. This works great until you refresh the page.

The problem is that the filters run BEFORE the charts do (duh) because we want to filter the data before we draw it. So they filter was trying the bin the data using the default binning scheme, and then once the chart updated it it was too late. All the other charts were now showing incorrectly filtered records.

The solution

Fixing this meant that the filter or the chart needed to be able to determine the binning scheme independently using only the userDictionary as that is all the filter has access to. First I made the userDictionary an observable and created a binning utility that would listen to the dictionary and recalculate the binning scheme when it changed. Then each class (the chart and the filter) get an instance of the binning utility singleton and use it to filter or display the data. This way the filter of the chart can give the bin utility the observable userDictionary depending on who's first.

I could have kept that userDictionary as a calculated value but the issue was that we were passing unique map objects to the filter so it had no way of easily figuring out if the data had changed or not, which meant the bin utility was constantly re-evaluating the bin scheme unnecesarely. The chart got around this problem by using data as an observable.

bemailloux (approved):

components/content-view-histogram.js
```diff
@@ -61,11 +120,11 @@ export class ContentViewHistogramFilter extends CategoryFilter {
 		const categories = value.split(',').map(category => Number(category));
 		this.setCategories(categories);
 	}
-}
 
-decorate(ContentViewHistogramFilter, {
-	setAll: action
-});
+	setAll(all) {
```
johngwilkinson, line 124: I don't think this does anything: if you don't override it, `this.setAll` is `super.setAll`.

model/data.js
```diff
@@ -95,6 +97,9 @@ export class Data {
 		this._serverData = newServerData;
 		this._serverDataProxy++;
 		this._selectorFilters.semester.selected = this.serverData.selectedSemestersIds || [];
+
+		this._userDictionary.clear();
+		newServerData.users.map(user => this._userDictionary.set(user[USER.ID], user));
```
johngwilkinson, line 102: `map` is for producing an array; you want `.forEach` I guess. If you're interested, I think if you make `_userDictionary` in the bin util `observable`, you'll be able to replace `_userDictionary` here instead of modifying it - generally that's safer - and still have `bins` update as needed.

johngwilkinson: Probably not a huge deal, but `_userDictionary` will perform better if it's `computed` rather than `observable` because the mobx overhead will be just on the getter, not the dictionary and each of its entries.

NicholasHallman: Ah yeah good catch on the map I'll switch it to forEach. I can't use computed because when I pass the userDictionary into the filter it looses it's 'computed' ness because I'm not calling the getter anymore to access it. To make it work I needed to either give the filter `data` first (which we didn't have any support for) or make the userDictionary an observable

NicholasHallman: I'm going to try your second suggestion to replace the variable instead of modify it thank you

johngwilkinson: Looking at this more, I think you are running into something I hit a while back. Take a look at this comment in engagement-dasbhoard.js:
```
                        // There are row filters - which look at each record individually when deciding to include or
			// exclude it - and there are aggregate filters, which also look at all records selected by
			// other filters (e.g. to include a record if it has an above-average value in some field).
			// Aggregate filters are potentially ambiguous if there are more than one and each depends
			// on the results of the other: we avoid this by building them on specific sets of filters.
```
What exactly are the semantics of filtering by bin here? Do the possible bins change depending on which data is selected by other filters? If so, this is an "aggregate filter", and it should be handled the way the ticvgrades filter is: by having its constructor invoked in engagement dashboard with the filtered data object from all the other filters. 

That will remove some ambiguity about what counts as an update, and get mobx working well because you'll have a reference to (filtered) `data`, whence you can get `data.userDictionary` as a computed as needed. I believe the util class and the filter class can collapse, and the filter can be used by the web component to get data from, like ticvgrades does.

Right now, there's this code smell of a) a singleton (which is a kind of global variable) b) which gets what you'd think would be a private property (userDictionary) set by other code some time after construction; and c) we're losing the connection between this property value and the object that owns it (`data`). The above would make all that go away.

NicholasHallman: > What exactly are the semantics of filtering by bin here?  

I'm not sure I 100% understand what you mean but the steps are
- get the bin scheme
- look at the records user and find the total content view from it's user record
- figure out which of the bins it fits into
- check if that bin is selected

> Do the possible bins change depending on which data is selected by other filters  

Nope, the bins should stay static unless more data is pulled from the server. So the bins are chosen by looking at the largest value in all of the available data that is not an outlier. Changing the org unit pulls more data, I think, and so it will update the binning scheme.

So this isn't like the ticvgrades cause the bin axis doesn't change. So unless I'm wrong I don't think it's an aggregates filter

> it should be handled the way the ticvgrades filter is: by having its constructor invoked in engagement dashboard with the filtered data object from all the other filters.

If I can do this with unfiltered data then this sounds like a great solution

johngwilkinson: Ah, okay, if the other filters don't affect the bins, then you're right, it's just a row filter. (Hopefully that's intended - for example, if most of the bins have been filtered out, they still appear but are empty, meaning you can't use filtering to get better bin resolution - I think that's probably fine for this data).

> If I can do this with unfiltered data then this sounds like a great solution

Yes, you could use the "call constructor with filteredData in engagement-dashboard.js" pattern regardless in this case, I think.

model/data.js
```diff
@@ -110,6 +109,28 @@ export class Data {
 		return this._selectorFilters.role.selected;
 	}
 
+	get userDictionary() {
+		this._serverDataProxy;
+		return new Map(this._serverData.users.map(user => [user[USER.ID], user]));
+	}
+
+	get userEnrollmentDictionary() {
```
NicholasHallman, line 117: This new computed value lets the bin scheme change when you select different org units. This way if you filter out certain courses you can get a more granular view of the data in the remaining courses.

components/content-view-histogram.js
```diff
@@ -78,31 +173,18 @@ class ContentViewHistogram extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		for (let i = 1; i <= 5; i++) {
-			bins.push([range * i, range * (i - 1)]);
-		}
-
-		if (this.courseAccessOutliers.length !== 0 || largestAccess > peaks[peaks.length - 1]) {
-			bins.push([Number.POSITIVE_INFINITY, upperBin]);
-		}
-
-		return bins.reverse();
+		this.filter.setAll(new Set(new Array(this.filter.bins.length + 1).fill(0).map((v, i) => i)));
```
johngwilkinson, line 182: is this line needed here? it's also in the getter for `filter.bins`

engagement-dashboard.js
```diff
@@ -550,7 +551,8 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 				.withFilter(new CurrentFinalGradesFilter())
 				.withFilter(new DiscussionActivityFilter());
 
-			this.__data = rowFilteredData.withFilter(new TimeInContentVsGradeFilter(rowFilteredData));
+			this.__data = rowFilteredData.withFilter(new TimeInContentVsGradeFilter(rowFilteredData))
+				.withFilter(new ContentViewHistogramFilter(rowFilteredData));
```
johngwilkinson, line 555: Looks like changes in the tic filter won't affect the cv filter, so maybe it would be better to build another intermediate value with the cv filter first, then pass that to the tic filter. That way filtering on content views will affect the averages used in the tic filter.

johngwilkinson (approved):

PR #515, opened 2021-03-19
removing Polymer style mixins

dlockhart:
In order to migrate table to Lit, we need to ditch these Polymer style mixins. I've [modified scroll-wrapper](https://github.com/BrightspaceUI/table/pull/228) so that whenever the `show-actions` attribute is present (which it is in table's usage), the dashed overflow borders appear automatically.

So no need to set these from here anymore.


components/table.js
```diff
@@ -167,26 +167,8 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
...
-				};
-				--d2l-scroll-wrapper-left: {
-					border-left: dashed 1px black;
-				};
-				--d2l-scroll-wrapper-right: {
-					border-right: none;
-				};
-
 				--d2l-scroll-wrapper-border-color: var(--d2l-color-galena);
 				--d2l-scroll-wrapper-background-color: var(--d2l-color-sylvite);
```
dlockhart, line 171: These two variables are still needed but hopefully I can remove the need for them in a future change.

NicholasHallman (approved):

bemailloux (approved):

PR #521, opened 2021-03-25
remove overrides for scroll-wrapper action colours

dlockhart:
We're [changing scroll-wrapper](https://github.com/BrightspaceUI/table/pull/234/files) so that the action button colours are no longer configurable -- they just use nice defaults that should work for everyone.

So setting these variables is no longer necessary.


dlockhart:
Closing as the visual diff tests don't work from a fork -- will reopen form a proper branch.

PR #522, opened 2021-03-25
removing unnecessary overrides

dlockhart:
We're [changing scroll-wrapper](https://github.com/BrightspaceUI/table/pull/234/files) so that the action button colours are no longer configurable -- they just use nice defaults that should work for everyone.

So setting these variables is no longer necessary.


NicholasHallman (approved):
Looks great in the vis diff thanks!

dlockhart:
> Looks great in the vis diff thanks!

Haha, really? The border/background disappeared in the visual diff because the table change hasn't merged yet. :) So I'll wait for that to happen, then re-run the visual diff and things should look a lot better.

PR #532, opened 2021-04-13
US123549: stub learner dashboard entrance

johngwilkinson:
Quality Notes

- functional
  - click on Insights Portal tile for Learner Engagement Dashboard: loads as expected (image below)
    - hit f5; reloads user selector page
  - click on "Back to Insights Portal": returns to insights portal
  - click on tile for Engagement Dashboard; goes to regular engagement dashboard
- perf
  - does not try to load dashboard data when hitting user selector
- security, ux/docs, devops: n/a

image

FYI @Vitalii-Misechko @anhill-D2L

### DIFF
--- a/locales/en.js
+++ b/locales/en.js
@@ -2,7 +2,6 @@
 
 export default {
 	"dashboard:title": "Engagement Dashboard",
-	"dashboard:learner:title": "Learner Engagement Dashboard",
 	"dashboard:userView:title": "Learner Engagement Dashboard",
 	"dashboard:backToInsightsPortal": "Back to Insights Portal",
 	"dashboard:backToEngagementDashboard": "Back to Engagement Dashboard",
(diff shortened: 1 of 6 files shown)


locales/en.js
```diff
@@ -2,6 +2,7 @@
 
 export default {
 	"dashboard:title": "Engagement Dashboard",
+	"dashboard:learner:title": "Learner Engagement Dashboard",
 	"dashboard:userView:title": "Learner Engagement Dashboard",
```
bemailloux, line 6: Should this line and the above be merged into one? I don't think the two pages should ever have different names.

bemailloux (approved):

locales/en.js
johngwilkinson: Good catch! I'll do that.

NicholasHallman (approved):
✔

PR #533, opened 2021-04-19
Detailed Filter Tag

NicholasHallman:
## Implementation Notes

Rewrote the template for the applied-filters component. The generic d2l component doesn't allow for updating the title of the filter items. The redone filter tag template has the following keyboard controls

 - Tab into to select first tag
 - Arrow keys to navigate tags
 - Space / Enter to delete a tag, disabling the filter

Tag css states are recreated as well. Including hover and focus

Small quality of life improvment the hitbox on the x is the entire right square of the tag and is now easier to click.

Some new features for filters
 - Descriptive titles | Used to show additional details in the filter tag | OPTIONAL used for showing ranges etc
 - axeDescription | adds a label to the tag for screen readers | OPTIONAL if excluded label is title. 

with range

## Quality Notes

Testing
  - [x] If a descriptive title is available use it
  - [x] Checked controlls with and without the screen reader active
  - [x] Screen reader reads the label on focus 
  - [x] No change to existing tags (existing tests work)
  - [x] Grades filter will group continuouse grades into a single range
  - [x] Grades filter will hide the range if mutliple non continuouse ranges are selected  
 
Docs - N/A

UX
  - Small change, Instead of hiding the filters if they overflow the tags wrap around.
  - Need to demo
 
DevOps - N/A

Cost - N/A

Performance - N/A


components/time-in-content-vs-grade-card.js
```diff
@@ -233,22 +233,25 @@ class TimeInContentVsGradeCard extends SkeletonMixin(Localizer(MobxLitElement))
 		const chartName = { chartName : this.localize('timeInContentVsGradeCard:timeInContentVsGrade') };
 		if (!this.filter.isApplied) return this.localize('alert:axeNotFiltering', chartName);
 
-		let description = this.localize('alert:axeDescription');
+		const curriedLocalizer = (term) => (options) => this.localize(term, options);
+		const localizerOptions = curriedLocalizer('alert:axeDescription');
```
Vitalii-Misechko, line 237: `til, minor:` `curriedLocalizer`-declaration looks nice, however, the code can omit currying function due to we know the first argument at the moment of declartion. Therefore, we could write something simple
```JS
const curriedLocalizer = (options) => this.localize('alert:axeDescription', options);
```

components/applied-filters.js
```diff
@@ -21,77 +23,199 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-					for="d2l-insights-applied-filters-dropdown"
-					label-text="${this.localize('appliedFilters:labelText')}">
-				</d2l-labs-applied-filters>`
-		: html`` }
-		`;
+			${!this.skeleton ?
+				html`
+				<div class="d2l-insights-tag-container">
+					<span class="d2l-insights-applied-filters-title">${this.localize('appliedFilters:labelText')}</span>
+					${repeat(filters, (item) => `${item.id}:${item.isApplied}`, (item) => (item.isApplied ? html`
```
Vitalii-Misechko, line 172: `curious` Why do you use `repeat` directive here instead of `map`?

components/applied-filters.js
```diff
@@ -21,77 +23,199 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+						<div @focus="${this._handleFocus}" tabindex="${isFirst.get() ? 0 : -1}" class="d2l-insights-tag-item" data-filter-id="${item.id}" aria-label="${item.description}" role="button">
```
Vitalii-Misechko, line 173: `minor` It seems that you can use `index` argument instead of `isFirst`-object by declaring `index`-argument 
```JS
${repeat(filters, (item) => `${item.id}:${item.isApplied}`, (item, index) => (item.isApplied ? html`
```

components/applied-filters.js
```diff
@@ -21,77 +23,199 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 			this.localize('alert:axeNotFiltering', { chartName: this.localize(filter.title) })
 		);
-		this.data.getFilter(event.detail.menuItemKey).isApplied = event.detail.selected;
+		filter.isApplied = false;
+
+		if (this.eventFollowedKeyEvent) {
+			// if we removed a filter using the keyboard we need to
+			// wait for this component to rerender and then
+			// focus on the new tag in this position.
+			this.eventFollowedKeyEvent = false;
```
Vitalii-Misechko, line 211: Thank you for these comments

Vitalii-Misechko (approved):
LGTM

components/applied-filters.js
NicholasHallman: Repeat will manage elements being added and removed from the array automatically for you. Instead of needing to remove and redraw all items to the dom it will updated only items that have changed.

components/applied-filters.js
NicholasHallman: I can't use the index property because the filters array contains filters that aren't applied so the index for applied items might be 

[3, 5, 6, 8] for example

components/applied-filters.js
```diff
@@ -21,77 +23,199 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+				description: f.axeDescription ? f.axeDescription(localizer, 'appliedFilters:axeDescriptionCategories') : title
+			};
+		});
 
 		if (filters.filter(f => f.isApplied).length < 1) {
-			return  html``;
+			return nothing;
 		}
 
 		filters.push({
```
johngwilkinson, line 147: Looks like this workaround can go now: the clear all button can be rendered directly in the template, I think.

components/applied-filters.js
```diff
@@ -4,10 +4,12 @@ import '@brightspace-ui-labs/facet-filter-sort/components/filter-dropdown/filter
 import '@brightspace-ui-labs/facet-filter-sort/components/filter-dropdown/filter-dropdown-category.js';
```
johngwilkinson, line 4: These imports are no longer used, right?

components/applied-filters.js
```diff
@@ -21,77 +23,199 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			// if we removed a filter using the keyboard we need to
+			// wait for this component to rerender and then
+			// focus on the new tag in this position.
+			this.eventFollowedKeyEvent = false;
+			setTimeout(() => {
+				const tags = this.shadowRoot.querySelectorAll('.d2l-insights-tag-item');
+
+				if (tags.length === 0) return;
+				tags[this.currentFocus].focus();
+			}, 10);
```
johngwilkinson, line 217: 10ms might not be long enough, e.g. on a laggy mobile device. Can you not just set a property for which tag should be focused then do the focusing during `render()`?

components/time-in-content-vs-grade-card.js
```diff
@@ -233,22 +233,24 @@ class TimeInContentVsGradeCard extends SkeletonMixin(Localizer(MobxLitElement))
 		const chartName = { chartName : this.localize('timeInContentVsGradeCard:timeInContentVsGrade') };
 		if (!this.filter.isApplied) return this.localize('alert:axeNotFiltering', chartName);
 
-		let description = this.localize('alert:axeDescription');
+		const curriedLocalizer = (options) => this.localize('alert:axeDescription', options);
```
johngwilkinson, line 236: Unless I'm misreading, this could be inlined, which would be slightly easier to read.

johngwilkinson (commented):
Don't forget right-to-left testing

components/applied-filters.js
NicholasHallman: I'm a little confused. Should we create a clear all tag seperately from the loop in the template and then have another decision to hide it if there are no active filters? Wouldn't it do the same thing?

components/applied-filters.js
NicholasHallman: render() won't work because it will be before the tag has been removed. But it works much better in updated()!

components/time-in-content-vs-grade-card.js
NicholasHallman: Yeah I got curried away

components/applied-filters.js
johngwilkinson: I think what you said would be better, yes. I'm just asking for a small refactor because it's only written this way (generate a fake extra filter) in order to shoehorn it into the component we used to be using.

components/applied-filters.js
johngwilkinson: Updated() is definitely better. I was thinking render could do it while drawing the tags - but I guess there's no lit-html way to express "has focus".

components/time-in-content-vs-grade-card.js
johngwilkinson: 🍛 🥴

components/applied-filters.js
```diff
@@ -123,6 +119,19 @@ class AppliedFilters extends SkeletonMixin(Localizer(MobxLitElement)) {
 		this.currentFocus = indexOfTag;
 	}
 
+	updated() {
+		if (this.eventFollowedKeyEvent) {
```
johngwilkinson, line 123: Just looking for more simplifying just in case: do we need this check here, or could this set the focus based on `this.currentFocus` after every `render()`?

NicholasHallman: If we remove this check then the couse will also cause the focus to change to the next tag if we click the x with a mouse which is not standard behaviour? Usually mouse itneractions don't trigger a focus on a different element but it makes sense with keyboard navigation to keep your place in the list. I can try focusing always but thinking about it I want to say it will be odd?

johngwilkinson: Nah, I don't think we should change the behaviour. I thought maybe this component would always know which element should have focus, but if not then this looks good.

johngwilkinson (approved):

PR #535, opened 2021-04-21
US126962 [Engagement] Learner Engagement Dashboard Page. Search

Vitalii-Misechko:
[US126962](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F521050018628&sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&fdp=true?fdp=true): [Engagement] Learner Engagement Dashboard Page

This PR includes only `search` input and its behavior. It was made to make early integration with John's branch

* Functional Testing 
  * [x] Browsers (Chrome, FF, Edge)
  * [x] Accessibility. The search can be fired with a keyboard
  * [x] Test cases
     * The page shows skeleton list of 10 items when a user hits the search button
* Security, devops, perf: N/A
* UX and docs
  ~* [ ] Demo to Kevin~ Will demo when I finish all components of the story
  
 FYI @NicholasHallman @anhill-D2L 
  
  
image

PR #537, opened 2021-04-22
US126962 [Engagement] Learner Engagement Dashboard Page. Header

Vitalii-Misechko:
[US126962](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F521050018628&sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&fdp=true?fdp=true): [Engagement] Learner Engagement Dashboard Page

Related PR: https://github.com/Brightspace/insights-engagement-dashboard/pull/535

This PR includes only the header of the user list. It was made to make early integration with John's branch

* Functional Testing 
  * [x] Accessibility. Screen reader tests will be made in the next story
  * [x] Test cases
     * The page header can be focused with keyboard
     * The page header reflects the current sort order
     * The current sort order change leads to data reload
     * The current sort order can be changed with mouse or keyboard
* Security, devops, perf: N/A
* UX and docs
  ~* [ ] Demo to Kevin~ Will demo when I finish all components of the story
  
 FYI @NicholasHallman @anhill-D2L 
  
  
image


bemailloux (approved):
Looks fine. An alternative approach which might reduce duplication would be to use the `table` component, but maybe that component doesn't fit exactly into your use case.

Vitalii-Misechko:
> Looks fine. An alternative approach which might reduce duplication would be to use the `table` component, but maybe that component doesn't fit exactly into your use case.

Good point! I had a similar thought, although after discussion with @johngwilkinson and some analysis of additional changes for the `table` component this alternative looked a bit better.

components/user-selector.js
```diff
@@ -60,12 +97,13 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 		this.viewState = null;
 
 		this._tokenPromise = this._tokenPromise.bind(this);
+		this._sortedAscending = false;
```
johngwilkinson, line 100: Something to clarify: is the user supposed to be able to sort by First Name _or_ Last Name? That was my assumption, but maybe I'm wrong.

Vitalii-Misechko: Good question! Kevin has responded to me that the common behaviour in the LMS is that users are able to do either.
It seems that I need to improve the current solution

Vitalii-Misechko: I've just added the ability to sort on lastname and firstname

components/user-selector.js
```diff
@@ -97,13 +111,14 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 		this.viewState = null;
 
 		this._tokenPromise = this._tokenPromise.bind(this);
+		this._sortColumn = SORT_COLUMN.FIRST_NAME;
```
johngwilkinson, line 114: Just checking: should the default be first name or last name?

johngwilkinson (approved):

components/user-selector.js
Vitalii-Misechko: Good catch! `last name` should be the first.

PR #553, opened 2021-04-28
Exclude fixed rules of axe-core 4.2.0

Vitalii-Misechko:
`dequelabs/axe-core` released `4.2.0` 2 days ago. This version fixes some axe rules which causes CI failures on this repo. The PR temporarily excludes these fixed rules to unblock CI

FYI @NicholasHallman


bemailloux:
Do you know how much effort might be required to fix these issues?

johngwilkinson (approved):
I haven't checked, but I wonder if you have the same viz diff failure I got: a couple of pixels in the slash below the users table.

Vitalii-Misechko:
> Do you know how much effort might be required to fix these issues?

No, I don't. But I couldn't find a quick solution for the first error. So, I assume It'll take some time. I hope to have more detailed info by tomorrows standup

bemailloux (approved):

Vitalii-Misechko:
> I haven't checked, but I wonder if you have the same viz diff failure I got: a couple of pixels in the slash below the users table.

The vis-diff states that there is `Difference (3.5px)` but I actually can't see it. I'm merging vis PR

NicholasHallman:
> > I haven't checked, but I wonder if you have the same viz diff failure I got: a couple of pixels in the slash below the users table.
> 
> The vis-diff states that there is `Difference (3.5px)` but I actually can't see it. I'm merging vis PR

Probably due too this thread https://d2l.slack.com/archives/C0PHG3QB0/p1619016333167900

Vitalii-Misechko:
I had to exclude the flaky `should render the users profile` test. It fails in Sauce Labs environment.

PR #556, opened 2021-04-29
Add Sort, Desc, Bookmark to API

NicholasHallman:
## Implementation Notes

Added a load more button that will query the server for more users on press.

Defining some assumed behaviour here.
- Pressing the load more button takes the last user in the list and uses them as a bookmark. The new results append. the bookmark is reset to undefined.
- Pressing the search button does not use the bookmark
- Pressing any of the sorts does not use the bookmark

For sorting and sort order this pr formats the component state into the queries desired formatting. 

Note from Kevin: 

> While I don't think infinite scroll is ideal if we can not give users a mechanism to jump to a specific view that they had in the UI infinite scroll is likely the best option so that users don't loose information

So load more will append to the list and not clear it

### Testing
 - Api accepts a search options object that validates allowed search parameters

### UX
 - Need to demo to kevin and discuss if we are replacing entries or appending them when the user hits load more.

### Performance - N/A

FYI: @bemailloux @anhill-D2L @hyehayes


components/user-selector.js
```diff
@@ -148,6 +159,7 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 					${this.users.map(u => this.userListItem(u))}
 				</d2l-list>
 			</div>
+			<d2l-button @click="${this._onLoadMore}" class="d2l-insights-laod-more">Load More</d2l-button>
```
Vitalii-Misechko, line 162: We can re-use `treeSelector:loadMoreLabel`-lang term. Typo `laod` => `load`

components/user-selector.js
```diff
@@ -97,8 +96,18 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 				.d2l-insights-user-selector-header-sort-indicator {
+					cursor: default;
 					pointer-events: none;
 				}
+
+				span[role="button"] {
+					cursor: pointer;
+					user-select: none;
+				}
```
Vitalii-Misechko, line 106: nice!

components/user-selector.js
```diff
@@ -262,30 +274,58 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 	}
 
 	_onSearch(e) {
-		this._search(e.detail.value, this._sortColumn, this._sortedAscending);
+		this._search(e.detail.value);
+	}
+
+	_onLoadMore() {
+		const lastUser = this.users[this.users.length - 1];
+		this._bookmark = `${lastUser.FirstName},${lastUser.LastName},${lastUser.Id}`;
```
johngwilkinson, line 282: This should come from the API response (`users.PagingInfo.Bookmark` I think)

components/user-selector.js
```diff
@@ -148,6 +159,7 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 					${this.users.map(u => this.userListItem(u))}
 				</d2l-list>
 			</div>
+			<d2l-button @click="${this._onLoadMore}" class="d2l-insights-load-more">${this.localize('treeSelector:loadMoreLabel')}</d2l-button>
```
johngwilkinson, line 162: This button should only appear if the API response indicates more items are available (`users.PagingInfo.HasMoreItems` I think)

components/user-selector.js
```diff
@@ -10,6 +10,7 @@ import { ListItemButtonMixin } from '@brightspace-ui/core/components/list/list-i
 import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
+import { nothing } from 'lit-html';
```
johngwilkinson, line 13: heads-up: lint issue

NicholasHallman: Auto import strikes again!

test/model/dataApiClient.test.js
```diff
@@ -392,12 +392,24 @@ describe('Lms', () => {
...
+			expect(await getVisibleUsers({})).to.deep.equal(mockLmsUserResponseData);
+		});
+
+		it('should not include a params other than allowed', async() => {
+			fetchMock.get(`${usersEndpoint}?search=&desc=&bookmark=&sort=`, mockLmsUserResponseData);
+			const options = {
+				search: '',
+				desc: '',
+				bookmark: '',
+				sort: '',
```
bemailloux, line 409: Maybe this should check that the values are filled in correctly as well?

bemailloux (approved):

components/user-selector.js
```diff
@@ -262,30 +283,59 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 	_search(searchText) {
+		const oldUsers = this.users;
 		this.skeleton = true;
 		this.users = usersForSkeleton;
+		const searchOptions = {
+			search: searchText,
+			desc: !this._sortedAscending,
+			sort: this._sortColumn === SORT_COLUMN.FIRST_NAME ? 'first' : 'last',
+			bookmark: this._fromLoadMore ? this._lastBookmark : undefined
```
Vitalii-Misechko, line 302: `ni`: Turning `this._fromLoadMore` into a function parameter would reduce code a bit

components/user-selector.js
```diff
@@ -148,6 +162,7 @@ class UserSelector extends SkeletonMixin(Localizer(MobxLitElement)) {
 					${this.users.map(u => this.userListItem(u))}
 				</d2l-list>
 			</div>
+			${this._renderLoadMore()}
```
Vitalii-Misechko, line 165: `minor`: You might want to add a test that verifies that `Loat More` appears when `PagingInfo.HasMoreItems === true`

Vitalii-Misechko (approved):
LGTM!

model/dataApiClient.js
```diff
@@ -231,11 +231,14 @@ export async function saveSettings(settings) {
...
+export async function getVisibleUsers(searchOptions) {
 	const url = new URL(userListEndpoint, window.location.origin);
-	if (search) {
-		url.searchParams.set('search', search);
-	}
+
+	const allowedParams = ['search', 'sort', 'desc', 'bookmark'];
+	Object.keys(searchOptions)
+		.filter(param => allowedParams.includes(param) && searchOptions[param] !== undefined)
+		.map(param => url.searchParams.set(param, searchOptions[param]));
```
johngwilkinson, line 240: minor: this should be `forEach`, for side-effects, as opposed to `map`, for transforming an array

PR #568, opened 2021-05-12
US126962: bug fixes and ui tweaks for learner engagement

johngwilkinson:
Quality Notes
- functional (all e2e)
  - explore as auditor and admin
  - various combinations of search, sort, and load more
    - default sort is ascending; name order in list matches header order
    - skeleton behaves well during load more: no re-rendering/hiding of already loaded items, and no expansion of skeleton list during load (if there are three items in the search results, sorting them shows three items in the skeleton)
  - non-ascii name
  - Eric Cooper (the default data user with a profile pic)
    - fixed "flash" when profile pic got loaded twice in user drill
  - various sequences of navigation using both the header and the browser nav buttons
    - fixed issue: visit a couple of users using just the header to go back; then go back through them using the browser buttons
      - used to show "undefined undefined" when reloading the first user
- perf/devops/security: n/a
- ux/docs
  - sorting in selector now works like elsewhere, e.g. class list, as per discussion with Kevin

FYI @NicholasHallman 

image

image


bemailloux (approved):

Vitalii-Misechko (approved):
LGTM!

PR #570, opened 2021-05-19
fix the version for puppeteer to match the version with visual-diff

JikaiLong:
insights-engagement-dashboard is using puppeteer version 8
and visual-diff version 3.0.0 is using puppeteer version 7
and visual-diff version 4.0.0 is using puppeteer version 8
NPM install failed due to mismatch of puppeteer versions, change the version of visual-diff from 3.0.0 to 4.0.0 fixed the issue for me 

Functional Testing
NPM install works 

Performance
NA

Security
NA

Dev-Ops
NA

Cost
NA

UX
NA

FYI @anhill-D2L  @hyehayes @johngwilkinson @Vitalii-Misechko

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
-    "@brightspace-ui/visual-diff": "^3.0.0",
     "@brightspace-ui/stylelint-config": "0.1.0",
+    "@brightspace-ui/visual-diff": "^4.0.0",
     "@open-wc/testing": "^2",
     "@open-wc/testing-karma": "^4.0.1",
     "@polymer/polymer": "^3.4.1",


NicholasHallman (approved):
LGTM!

package.json
```diff
@@ -23,8 +23,8 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7",
     "@babel/plugin-proposal-class-properties": "^7.13.0",
-    "@brightspace-ui/visual-diff": "^3.0.0",
     "@brightspace-ui/stylelint-config": "0.1.0",
+    "@brightspace-ui/visual-diff": "^4.0.0",
```
bemailloux, line 27: Do we actually need to include puppeteer in devDependencies? I thought visual-diff was responsible for getting its own dependencies, and so we wouldn't need to get puppeteer ourselves.

bemailloux: Ehh :shrug:

I'd prefer to have it here than have to install globally

bemailloux (approved):

bemailloux:
@JikaiLong FYI: Sauce tests are notoriously flaky. If they've failing for no obvious reason it's likely that they will pass if you run them again.

package.json
JikaiLong: Just to double check, I tried uninstall puppeteer and ran test cases again and got the same result:
 
SUMMARY:
√ 620 tests completed
i 2 tests skipped

Looks like the package is indeed not used for now.

PR #576, opened 2021-06-01
switching to Lit version of scroll-wrapper

dlockhart:
This has been migrated to Lit so no need to use the old table one anymore.


components/table.js
```diff
@@ -183,7 +183,7 @@ class Table extends SkeletonMixin(Localizer(RtlMixin(LitElement))) {
 
 	render() {
 		return html`
-			<d2l-scroll-wrapper show-actions>
```
dlockhart, line 186: Actions are shown by default.

NicholasHallman: Awesome thank you!

NicholasHallman (approved):

PR #577, opened 2021-06-02
Add required labels to selectable d2l-list-items.

dbatiste:
This PR adds labels to selectable d2l-list-items that will be required in the near future in order to support changes that are being made to lists for selection, multi-selection, and multi-actions,

### DIFF
--- a/components/card-selection-list.js
+++ b/components/card-selection-list.js
@@ -109,7 +109,7 @@ class CardSelectionList extends RtlMixin(Localizer(LitElement)) {
 
 		return html`
 			<d2l-list id="card-selection-list" @d2l-list-selection-change="${this._handleCardSelectionListChange}">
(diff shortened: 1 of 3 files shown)


johngwilkinson (approved):
It's possible this component won't be released again in the near future - is it important to get this update into production by a certain time?

dbatiste:
> It's possible this component won't be released again in the near future - is it important to get this update into production by a certain time?

Yeah I think we'll need this updated to unblock some core changes related to how labels are handled for list-items.  It looks like this doesn't use the semantic release github action, do I think I will need to manually update the version and create the release.

bemailloux (approved):

bemailloux:
> I think I will need to manually update the version and create the release

Our release process is documented here: https://github.com/Brightspace/insights-engagement-dashboard#releasing-and-versioning

TL;DR if this is going into 20.21.7 it should be released as 1.2107.x

PR #579, opened 2021-06-14
Notify Slack Webhook on Locale Change

NicholasHallman:
Removes the labeler action and replaces it with a webhook action.

Placing a webhook into this repo's SLACK_WEBHOOK secret will push a webhook to that slack channel when
- a pull request is opened
- and there have been changes to locale

### DIFF
--- a/.github/notify.sh
+++ b/.github/notify.sh
@@ -1,2 +1,2 @@
 #!bin/bash
-curl -X POST -H 'Content-type: application/json' --data '{"text":"Translation Push Created"}' https://hooks.slack.com/services/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
+curl -X POST -H 'Content-type: application/json' --data '{"text":"Translation PR Created"}' ${WEBHOOK}
--- a/.github/notify.sh
+++ b/.github/notify.sh
@@ -1,0 +1,2 @@
+#!bin/bash
+curl -X POST -H 'Content-type: application/json' --data '{"text":"Translation Push Created"}' https://hooks.slack.com/services/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
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -1,8 +1,8 @@
 name: Draco Insights Dashboard Notify
 on:
   pull_request: ["opened"]
-  paths:
-    - 'locales/**/*'
+  paths: ['locales/**/*']
+
 jobs:
   notify:
     runs-on: ubuntu-latest
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -1,0 +1,10 @@
+name: Draco Insights Dashboard Notify
+on:
+  pull_request:
+  paths:
+    - 'locales/**/*'
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+    - run: .github/notify.sh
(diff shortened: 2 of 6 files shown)


.github/workflows/notify.yml
```diff
@@ -0,0 +1,10 @@
+name: Draco Insights Dashboard Notify
+on:
+  pull_request:
+  paths:
+    - 'locales/**/*'
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+    - run: .github/notify.sh
```
neverendingqs, line 10:
```suggestion
    - run: curl -X POST -H 'Content-type: application/json' --data '{"text":"Translation Push Created"}' ${SLACK_WEBHOOK}
      env:
        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
```

(You'll need to set `SLACK_WEBHOOK` in https://github.com/Brightspace/insights-engagement-dashboard/settings/secrets/actions)

.github/workflows/notify.yml
```diff
@@ -0,0 +1,10 @@
+name: Draco Insights Dashboard Notify
+on:
+  pull_request:
```
neverendingqs, line 3:
```suggestion
  pull_request: ["Created"]
```

I think this is what you want. Otherwise, it'll run every time someone pushes to the pull request.

.github/workflows/notify.yml
```diff
@@ -0,0 +1,10 @@
...
+  paths:
+    - 'locales/**/*'
+jobs:
+  notify:
```
neverendingqs, line 7: You may want something like the following to avoid sending a Slack message for "manual" changes.

```suggestion
  notify:
    if: ${{ github.actor == 'brightspace-bot' }}
```

* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif
* https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

.github/workflows/notify.yml
neverendingqs: I think calling a script should work too. I just did inline to show how the env var is used.

.github/workflows/notify.yml
NicholasHallman: I'll include this as a second PR to make sure it sends the push first, then add the bot condition

.github/notify.sh
```diff
@@ -0,0 +1,2 @@
+#!bin/bash
+curl -X POST -H 'Content-type: application/json' --data '{"text":"Translation PR Created"}' ${WEBHOOK}
```
bemailloux, line 2: Does webhook have any kind of auth on it?

(As in, why is it not possible for someone to just randomly guess the link, and spam stuff to it?)

NicholasHallman: The webhook has a reasonably long 24 character hash on the end of it which I assume works as the password. So you would need to know our slack orgs id, our slack channel id and guess the hash. Each webhook is only authorized to post into one channel so if somebody did guess it they would be limited to sending garbage to that single channel and not anywhere they please in our org

bemailloux (approved):

.github/notify.sh
bemailloux: Ah ok, the hash is unique to org/channel. That makes sense. I was thinking that if it was a hash only, someone could spam a message to any hash and have a non-zero chance of actually hitting something. But if they need to know the org and channel too then it's much harder :)

anhill-D2L (approved):

PR #583, opened 2021-06-15
Fix Notify and Test and Add Bot Limit

NicholasHallman:
It worked!

translation notify

Now that we know the webhook fires properly I am limiting the job to only trigger when the brightspace-bot triggers the action as was Marks suggestion.

### DIFF
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -9,6 +9,7 @@ jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
+    - uses: Brightspace/third-party-actions@actions/checkout
     - run: bash ./.github/notify.sh
       env:
         WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   notify:
+    if: ${{ github.actor == 'brightspace-bot' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/notify.yml
```diff
@@ -7,8 +7,10 @@ on:
 
 jobs:
   notify:
+    if: ${{ github.actor == 'brightspace-bot' }}
     runs-on: ubuntu-latest
     steps:
-    - run: .github/notify.sh
+    - uses: Brightspace/third-party-actions@actions/checkout
```
neverendingqs, line 13: Curious: why do we need checkout here?

NicholasHallman: we need to checkout the code into the container running the action so we can run the notify.sh script

neverendingqs: Oh, so we can run the script. Duh 😅.

Since the script is a one-liner, you can skip cloning (which reduces runtime) and just do it inline? If you need multiple lines, you can do something like this:

```yaml
- run: |
    echo "hi"
    echo "hi again"
```

Up to you. Not a big deal for me either way.

neverendingqs (approved):

bemailloux (approved):

.github/workflows/notify.yml
NicholasHallman: I'm going to keep it like this because it's annoying to test, I have to open new Pr's to see if it's working or not 😅

PR #585, opened 2021-06-16
Bump node version used in GitHub actions

margaree:
We are planning to bump the version of Puppeteer used in visual-diff and so this PR is to bump the node version to the current LTS version.


NicholasHallman (approved):
Thanks! LGTM

svanherk (approved):

PR #598, opened 2021-06-30
Opets/de44136 dynamic tree filter

opets:
changes:
 - moved orgunit mapping logic from `data.js` to `dataApiClient.js`
 
* Functional Testing
  * [x] Accessibility
  * [x] RTL
  * [x] Responsive
  * [x] Test cases
  * [x] Manul UI testing with LMS API and `Tree.isDynamic == true`
* Security, devops, perf: N/A


model/data.js
```diff
@@ -108,18 +108,11 @@ export class Data {
...
-			Type: n[ORG_UNIT.TYPE],
-			Parents: n[ORG_UNIT.PARENTS],
-			IsActive: n[ORG_UNIT.IS_ACTIVE]
-		}));
 
 		this.orgUnitTree = new Tree({
 			// add in any nodes from the most recent search (if the semester filter didn't change); otherwise
 			// the search will blink out and come back, and also drop any "load more" results
-			nodes: mappedNodes,
+			nodes: nodes,
```
bemailloux, line 115: `Nit / just FYI` This could also be written:
```
this.orgUnitTree = new Tree({
  nodes,
  leafTypes: [COURSE_OFFERING],
  ...
}
```

bemailloux (approved):

model/data.js
opets: fixed

Vitalii-Misechko:
LGTM! Will approve right after E2E tests.
Have you a chance to make E2E tests on Draco LMS? 
If not could you make E2E  tests when `Tree.isDynamic == true`?

opets:
> LGTM! Will approve right after E2E tests.
> Have you a chance to make E2E tests on Draco LMS?
> If not could you make E2E tests when `Tree.isDynamic == true`?

I tested the on the local machine with `isDynamic == true`, is this ennought?

Vitalii-Misechko (approved):
> I tested this on the local machine with isDynamic == true, is this ennought?

Sounds good!

PR #604, opened 2021-07-12
Role List Improvments

NicholasHallman:
Implementation Notes

Here we do our best to prevent the user from entering the "No roles selected state" because it is confusing. When no roles are selected we show all roles because it is nice to have data when you load into the dashboard for the first time. So once you have made some refinements the user should not be able to get back into "default" mode.

Corner Case

If the user selects a single role, say instructor, then goes into their config variables and disables the instructor role they will see no data. Entering the settings menu will have all of the roles unchecked and the save and clos ebutton enabled. Selecting any role, then unselecting back to nothing will once again disable the save and close button.

AC

1 When all roles are selected in the Role Filters, data for all roles should appear in the dashboard
  - This has always been true. No change

2 When no roles are selected, no data should appear in the dashboard
  - When you first log in all roles are selected (in db no roles selected)
  - If you select only one role, then disable it in the config no roles will appear
  - If you attempt to deselect all roles the button is disabled

3 We need to provide the user with the ability to easily select all and deselect all Role Filter options
  - Buttons added.

4 At least one role should be selected. Show an message when no filters are selected in the Settings
  - Input checkboxes are turned red and a caution icon is added to the role filter title. Saving scrolls the browser to an alert which lists out page errors.

5 For existing configurations, in instances when no roles are selected they should have all roles selected following the defect fix
 - This was always the case, no change

Screenshot (2)

settings alert

Quality Notes

Testing
- Tested on local LMS with mocked data from a local bdp-query 

Security Dev-Ops Documentation N/A

FYI: @Brightspace/draco-vis


components/dashboard-settings.js
```diff
@@ -177,8 +180,12 @@ class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
 		this.showCourseAccessTrendCard = false;
 		this.showGradesTrendCard = false;
 		this.showPredictedGradeCol = false;
+
+		this.allowSave = true;
 	}
 
+	firstUpdated() {}
```
Vitalii-Misechko, line 187: `nit` delete?

model/dataApiClient.js
```diff
@@ -29,13 +29,16 @@ function concatMetronUrl(endpoint, apiPath) {
...
 function mapOrgUnits(orgunits) {
-	return orgunits.map(orgunit => ({
-		Id: orgunit[ORG_UNIT.ID],
-		Name: orgunit[ORG_UNIT.NAME],
-		Type: orgunit[ORG_UNIT.TYPE],
-		Parents: orgunit[ORG_UNIT.PARENTS],
-		IsActive: orgunit[ORG_UNIT.IS_ACTIVE]
-	}));
+	return orgunits.map(orgunit => {
+		if (orgunit.constructor.name !== 'Array') return orgunit; // only map if in the old format
```
Vitalii-Misechko, line 33: - Proposal `if(Array.isArray(orgunit)) ...`. 

- This change will allow us to change the existing endpoint output data format, right?

Vitalii-Misechko:
Do these new buttons visible on visual-diff tests? I just don't see them

Vitalii-Misechko:
4-th point of ACs is somewhat controversial to me. `Save and Close` button is disabled so there is no need to move the mouse pointer to it. Therefore, a user might not see the tooltip. Let see what Kevin says.

Vitalii-Misechko (approved):

model/dataApiClient.js
NicholasHallman: I think so, I didn't test this with bdp-query but with a mock metron server that returns data identical to the demo data. So I can't guarentee this works for real data atm

components/role-list.js
```diff
@@ -77,6 +81,15 @@ class RoleList extends Localizer(LitElement) {
 		this._setRoleData(data);
 	}
 
+	get includeRoles() {
+		return this._includeRoles;
+	}
+
+	set includeRoles(v) {
+		console.log(v);
```
bemailloux, line 89: Remove console.log?

NicholasHallman:
> 4-th point of ACs is somewhat controversial to me. Save and Close button is disabled so there is no need to move the mouse pointer to it. Therefore, a user might not see the tooltip. Let see what Kevin says.

Good point, I can't tab to the disabled button so I can't see the tooltip using the keyboard. I've asked kevin for alternatives

components/dashboard-settings.js
```diff
@@ -177,6 +213,49 @@ class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
...
 		this.showPredictedGradeCol = false;
+		this.errors = undefined;
+		this.isAlertCollapsed = false;
+	}
+
+	_handleScroll(e) {
+		const id = e.target.getAttribute('data:id');
+		const elm = shadowHash.querySelector(`#${id}`);
+		const distance = window.pageYOffset + elm.getBoundingClientRect().top;
+		window.scrollTo({ top: distance, behavior: 'smooth' });
```
Vitalii-Misechko, line 224: nice!

components/dashboard-settings.js
```diff
@@ -177,6 +213,49 @@ class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
...
+	}
+
+	get _hasRoleListError() {
+		return this.errors ? !!this.errors.find(e => e === INVALID_ROLE_SELECTION) : false;
```
Vitalii-Misechko, line 228: `jfy` alternative `this.errors && !!this.errors.find(e => e === INVALID_ROLE_SELECTION)`

components/role-list.js
```diff
@@ -75,6 +106,18 @@ class RoleList extends Localizer(LitElement) {
...
+
+	updated(changed) {
+		if (changed.has('isError')) {
+			const checkboxes = this.shadowRoot.querySelectorAll('d2l-input-checkbox');
+			if (this.isError) {
+				checkboxes.forEach(check => check.shadowRoot.querySelector('input[type="checkbox"]').style = 'border-color: var(--d2l-color-cinnabar)');
+			} else {
+				checkboxes.forEach(check => check.shadowRoot.querySelector('input[type="checkbox"]').style = '');
+			}
+		}
```
Vitalii-Misechko, line 120: `jfy` there is a declarative alternative. We could use a checkbox CSS class that depends on `isError`. 

```JS
.some-css-is-error {
  border-color: var(--d2l-color-cinnabar);
}
const isErrorStyle = {
  'some-css-is-error': this.isError
}
			<div class="${classMap(styles)}">
				${filterData.map(item => html`<d2l-input-checkbox class="${classMap(isErrorStyle)}" value="${item.id}" @change="${this._handleSelectionChange}" ?checked="${item.selected}" >${item.displayName}</d2l-input-checkbox>`)}
			</div>
```

components/role-list.js
```diff
@@ -111,17 +187,27 @@ class RoleList extends Localizer(LitElement) {
...
 	_handleSelectionChange() {
-		this.includeRoles = this._selected;
+		// prevent mobx-lit commit from over-writing the array by reusing the proxy
+		this.includeRoles.length = 0;
+		this._selected.forEach(selected => this.includeRoles.push(selected));
+		if (this.includeRoles.length === 0) {
+			this.isError = true;
+		} else {
+			this.isError = false;
+		}
```
Vitalii-Misechko, line 210: `jfy` shorter alternative 
```JS
this.isError = this.includeRoles.length === 0
```

model/shadowHash.js
```diff
@@ -0,0 +1,42 @@
+// The browser traverses the dom tree and adds all components with an ID to a hash table.
+// However, it excludes elements in shadowRoots since they are technically separate DOM trees.
+// To make up for this limitation the shadowHash allows you to register and query for components outside
+// of the browsers tooling. Allowing for linking, scrolling etc.
+
+const CLASS = Symbol('class');
+const ID = Symbol('id');
+const ELEMENT = Symbol('element');
+
+class ShadowHash {
```
Vitalii-Misechko, line 10: Could you create a couple of unit test for this class?

components/role-list.js
NicholasHallman: Can we use one of our components style's inside another component? I thought the sahdow dom prevented it? Or can components use the styles of it's parents?

Vitalii-Misechko (approved):
LGTM

components/role-list.js
Vitalii-Misechko: No, we can't. You thought right. Now, I'm not sure if my code would work.

Vitalii-Misechko (approved):

components/dashboard-settings.js
```diff
@@ -15,10 +16,21 @@ import { heading1Styles, heading2Styles } from '@brightspace-ui/core/components/
...
  */
+
+const INVALID_SYSTEM_ACCESS = Symbol('invalid-system-access');
+const INVALID_ROLE_SELECTION = Symbol('invalid-role-selection');
+
+const ERROR_LINKS = {};
+ERROR_LINKS[INVALID_SYSTEM_ACCESS] = { term: 'settings:systemAccessError', id: 'last-access-threshold-edit' };
+ERROR_LINKS[INVALID_ROLE_SELECTION] = { term: 'settings:roleListError', id: 'role-list-items' };
+
+const IMMERSIVE_NAV_HEIGHT = 62;
```
bemailloux, line 32: This seems likely to break eventually

components/role-list.js
```diff
@@ -51,14 +54,42 @@ class RoleList extends Localizer(LitElement) {
 				min-height: 50px;
 			}
 
+			.d2l-insights-spacer {
+				margin-bottom: 10px;
+			}
+
+			.d2l-insights-invalid {
+				background-image: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjIiIGhlaWdodD0iMjIiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+CiAgPGcgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIj4KICAgIDxwYXRoIGZpbGw9IiNGRkYiIGQ9Ik0wIDBoMjJ2MjJIMHoiLz4KICAgIDxwYXRoIGQ9Ik0xOC44NjQgMTYuNDdMMTIuNjIzIDMuOTg5YTEuNzgzIDEuNzgzIDAgMDAtMy4xOTIgMEwzLjE4OSAxNi40N2ExLjc2MSAxLjc2MSAwIDAwLjA4IDEuNzNjLjMyNS41MjUuODk4Ljc5OCAxLjUxNi43OTloMTIuNDgzYy42MTggMCAxLjE5Mi0uMjczIDEuNTE2LS44LjIzNy0uMzM1LjI2NS0xLjM3LjA4LTEuNzN6IiBmaWxsPSIjQ0QyMDI2IiBmaWxsLXJ1bGU9Im5vbnplcm8iLz4KICAgIDxwYXRoIGQ9Ik0xMS4wMjcgMTcuMjY0YTEuMzM3IDEuMzM3IDAgMTEwLTIuNjc1IDEuMzM3IDEuMzM3IDAgMDEwIDIuNjc1ek0xMS45IDEyLjk4YS44OTIuODkyIDAgMDEtMS43NDcgMEw5LjI3IDguNTJhLjg5Mi44OTIgMCAwMS44NzQtMS4wNjRoMS43NjhhLjg5Mi44OTIgMCAwMS44NzQgMS4wNjVsLS44ODYgNC40NTh6IiBmaWxsPSIjRkZGIi8+CiAgPC9nPgo8L3N2Zz4K);
```
bemailloux, line 62: Could this go in a separate file (and not in b64)?

NicholasHallman: yee

components/dashboard-settings.js
NicholasHallman: We use this to make sure we scroll just a little bit above where we want to land because the immersive nav takes up part of the screen. It's purely cosemetic

bemailloux: But I mean if the immersive nav height ever changes then this won't work correctly

consts.js
```diff
@@ -79,3 +81,5 @@ export function* UserTrendColorsIterator(start = 0, step = 1, end = 40) {
 		yield USER_TREND_COLORS[i % USER_TREND_COLORS.length];
 	}
 }
+
+export const warningIconImage = css`background-image: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjIiIGhlaWdodD0iMjIiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+CiAgPGcgZmlsbD0ibm9uZSIgZmlsbC1ydWxlPSJldmVub2RkIj4KICAgIDxwYXRoIGZpbGw9IiNGRkYiIGQ9Ik0wIDBoMjJ2MjJIMHoiLz4KICAgIDxwYXRoIGQ9Ik0xOC44NjQgMTYuNDdMMTIuNjIzIDMuOTg5YTEuNzgzIDEuNzgzIDAgMDAtMy4xOTIgMEwzLjE4OSAxNi40N2ExLjc2MSAxLjc2MSAwIDAwLjA4IDEuNzNjLjMyNS41MjUuODk4Ljc5OCAxLjUxNi43OTloMTIuNDgzYy42MTggMCAxLjE5Mi0uMjczIDEuNTE2LS44LjIzNy0uMzM1LjI2NS0xLjM3LjA4LTEuNzN6IiBmaWxsPSIjQ0QyMDI2IiBmaWxsLXJ1bGU9Im5vbnplcm8iLz4KICAgIDxwYXRoIGQ9Ik0xMS4wMjcgMTcuMjY0YTEuMzM3IDEuMzM3IDAgMTEwLTIuNjc1IDEuMzM3IDEuMzM3IDAgMDEwIDIuNjc1ek0xMS45IDEyLjk4YS44OTIuODkyIDAgMDEtMS43NDcgMEw5LjI3IDguNTJhLjg5Mi44OTIgMCAwMS44NzQtMS4wNjRoMS43NjhhLjg5Mi44OTIgMCAwMS44NzQgMS4wNjVsLS44ODYgNC40NTh6IiBmaWxsPSIjRkZGIi8+CiAgPC9nPgo8L3N2Zz4K);`;
```
bemailloux, line 85: I was thinking instead you could just have it as a .svg file in the repo - that way when someone wants to change it they just drop in the new svg without worrying about converting to b64

NicholasHallman: Well ya see I don't have the svg because I stole that code from here :) https://github.com/BrightspaceUI/core/blob/d5056a572e0dcda90ee208359eef37d334200304/components/inputs/input-text.js#L202

NicholasHallman: I don't think I can move the css into another file either because the linter doesn't think that my totally legit very good css variable is allowed

bemailloux: This is the raw svg
```
<svg width="22" height="22" xmlns="http://www.w3.org/2000/svg">
    <g fill="none" fill-rule="evenodd">
        <path fill="#FFF" d="M0 0h22v22H0z"/>
        <path d="M18.864 16.47L12.623 3.989a1.783 1.783 0 00-3.192 0L3.189 16.47a1.761 1.761 0 00.08 1.73c.325.525.898.798 1.516.799h12.483c.618 0 1.192-.273 1.516-.8.237-.335.265-1.37.08-1.73z" fill="#CD2026" fill-rule="nonzero"/>

        <path d="M11.027 17.264a1.337 1.337 0 110-2.675 1.337 1.337 0 010 2.675zM11.9 12.98a.892.892 0 01-1.747 0L9.27 8.52a.892.892 0 01.874-1.064h1.768a.892.892 0 01.874 1.065l-.886 4.458z" fill="#FFF"/>
    </g>
</svg>
```

NicholasHallman: Since you can't include a literal in a css tempalte this is impossible, so I'm going to keep the other solution (seems to work for gaudi)

components/svg/warning-icon.js
```diff
@@ -0,0 +1,12 @@
...
+export const renderWarning = () => html`
+	<svg width="22" height="22" xmlns="http://www.w3.org/2000/svg">
+		<g fill="none" fill-rule="evenodd">
+			<path fill="#FFF" d="M0 0h22v22H0z"/>
+			<path d="M18.864 16.47L12.623 3.989a1.783 1.783 0 00-3.192 0L3.189 16.47a1.761 1.761 0 00.08 1.73c.325.525.898.798 1.516.799h12.483c.618 0 1.192-.273 1.516-.8.237-.335.265-1.37.08-1.73z" fill="#CD2026" fill-rule="nonzero"/>
+
+			<path d="M11.027 17.264a1.337 1.337 0 110-2.675 1.337 1.337 0 010 2.675zM11.9 12.98a.892.892 0 01-1.747 0L9.27 8.52a.892.892 0 01.874-1.064h1.768a.892.892 0 01.874 1.065l-.886 4.458z" fill="#FFF"/>
+		</g>
+	</svg>
+`;
```
bemailloux, line 12: 👍

bemailloux (approved):

PR #606, opened 2021-07-19
No Roles Error Messaging

NicholasHallman:
We now show an error message if the user's roles that they previously had selected are now excluded by the config variables.

The buton on this error message opens the settings page when pressed (reuses the same code as the settings button in the collapsable menu)

This change also observes the differences between a new user with no roles selected seeing the dashboard for the firest time; where the lms returns all results, and a user whos selected roles have been disabled in the config; in which the lms returns null.

Quality Notes

Testing
- Tested on demo
- Tested on Local LMS with mock bdp query
- Tested on Local LMS with bdp Draco stack
- [x] Should open the settings page when dashboard errors emits the appropriate event
- [x] Should display the role errors alert when the selectedRoles is null

UX
 - Wording approved by kevin
 
All others N/A

FYI: @Brightspace/draco-vis

### DIFF
--- a/components/engagement-dashboard-errors.js
+++ b/components/engagement-dashboard-errors.js
@@ -11,7 +11,7 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
 			data: { type: Object, attribute: false },
-			isNoDataReturned: { type: Boolean, attribute: 'no-data' },
+			isNoDataReturned: { type: Boolean, attribute: 'no-results' },
 			isNoRoles: { type: Boolean, attribute: 'no-roles' }
 		};
 	}
--- a/components/engagement-dashboard-errors.js
+++ b/components/engagement-dashboard-errors.js
@@ -11,7 +11,8 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
 			data: { type: Object, attribute: false },
-			isNoDataReturned: { type: Boolean, attribute: false }
+			isNoDataReturned: { type: Boolean, attribute: false },
+			isNoRoles: { attribute: false }
 		};
 	}
 
@@ -34,7 +35,12 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 		let messageType, text, linkText, href, buttonText;
 
 		// conditionally render message text and body
-		if (this._isQueryFails) {
+		if (this.isNoRoles) {
+			messageType = 'button';
+			text = this.localize('dashboard:noRolesSelected');
+			buttonText = this.localize('dashboard:goToSettings');
+
+		} else if (this._isQueryFails) {
 			messageType = 'link';
 			text = this.localize('dashboard:queryFailed');
 			linkText = this.localize('dashboard:queryFailedLink');
@@ -66,7 +72,11 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	}
 
 	_handleClick() {
-		this.dispatchEvent(new Event('d2l-insights-undo-last-filter'));
+		if (this.isNoRoles) {
+			this.dispatchEvent(new Event('d2l-insights-go-to-settings'));
+		} else {
+			this.dispatchEvent(new Event('d2l-insights-undo-last-filter'));
+		}
 	}
 }
 
--- a/components/engagement-dashboard-errors.js
+++ b/components/engagement-dashboard-errors.js
@@ -11,8 +11,8 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
 			data: { type: Object, attribute: false },
-			isNoDataReturned: { type: Boolean, attribute: false },
-			isNoRoles: { attribute: false }
+			isNoDataReturned: { type: Boolean, attribute: 'no-data' },
+			isNoRoles: { type: Boolean, attribute: 'no-roles' }
 		};
 	}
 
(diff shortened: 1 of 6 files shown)


model/data.js
```diff
@@ -96,6 +96,11 @@ export class Data {
 		};
 		try {
 			const data = await this.recordProvider(filters, this._metronEndpoint);
+			if (data.selectedRolesIds === null) {
+				this.onServerDataReload(this.serverData);
+				this.isQueryError = false;
+				return;
+			}
```
NicholasHallman, line 103: This prevents the other cards from rendering if we detect the selectedRoleIds is in the null state.

bemailloux: When is selectedRoleIds null again? Is it just when all roles have been excluded by config variables / settings, or is it also when the user first loads?

bemailloux: Gotcha. I think a comment in code is needed to understand this.

Also, just so I'm understanding fully, the intention here is: if user loads and selectedRolesIds comes back as null, then attempt to "load" the dashboard using an empty server data object. Therefore the error component loads the correct error message. Right?

engagement-dashboard.js
```diff
@@ -429,7 +429,9 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 			<d2l-insights-engagement-dashboard-errors
 				.data="${this._data}"
 				.isNoDataReturned="${this._isNoUserResults}"
-				@d2l-insights-undo-last-filter="${this._handleUndo}">
+				.isNoRoles="${this._parsedIncludeRoles.length === 0}"
```
Vitalii-Misechko, line 432: `curious` why are these not attributes like `no-data` and `no-roles`?

components/engagement-dashboard-errors.js
```diff
@@ -11,7 +11,8 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	static get properties() {
 		return {
 			data: { type: Object, attribute: false },
-			isNoDataReturned: { type: Boolean, attribute: false }
+			isNoDataReturned: { type: Boolean, attribute: false },
+			isNoRoles: { attribute: false }
```
Vitalii-Misechko, line 15: Should we add `type: Boolean` ?

components/engagement-dashboard-errors.js
```diff
@@ -66,7 +72,11 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 	}
 
 	_handleClick() {
-		this.dispatchEvent(new Event('d2l-insights-undo-last-filter'));
+		if (this.isNoRoles) {
+			this.dispatchEvent(new Event('d2l-insights-go-to-settings'));
+		} else {
+			this.dispatchEvent(new Event('d2l-insights-undo-last-filter'));
+		}
```
Vitalii-Misechko, line 79: Should we add any test that verifies the correct event?

components/engagement-dashboard-errors.js
```diff
@@ -34,7 +35,12 @@ class EngagementDashboardErrors extends Localizer(MobxLitElement) {
 		let messageType, text, linkText, href, buttonText;
 
 		// conditionally render message text and body
-		if (this._isQueryFails) {
+		if (this.isNoRoles) {
+			messageType = 'button';
+			text = this.localize('dashboard:noRolesSelected');
+			buttonText = this.localize('dashboard:goToSettings');
+
+		} else if (this._isQueryFails) {
```
Vitalii-Misechko, line 43: Should we add any render test to verify proper HTML regarding which flag is activated?

engagement-dashboard.js
NicholasHallman: No reason, I can change those

components/engagement-dashboard-errors.js
NicholasHallman: If it's not an attribute it doesn't need a type because there's no string attribute to property conversion

model/data.js
NicholasHallman: That's right, I'll add the comment to clarify

components/engagement-dashboard-errors.js
Vitalii-Misechko: thanks! TIL

engagement-dashboard.js
```diff
@@ -428,8 +428,10 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 			</div>
 			<d2l-insights-engagement-dashboard-errors
 				.data="${this._data}"
-				.isNoDataReturned="${this._isNoUserResults}"
-				@d2l-insights-undo-last-filter="${this._handleUndo}">
+				?no-data="${this._isNoUserResults}"
```
Vitalii-Misechko, line 431: @NicholasHallman `.data` and `no-data` look a bit confusing together. 
Sorry for this suggestion. 
Could you rename `no-data` to `no-results` if you don't mind?

components/engagement-dashboard-errors.js
Vitalii-Misechko: I meant to say we needed a test that checks that `_handleClick` raises the correct event type. Sorry for my ambiguous sentence in the previous post

test/engagement-dashboard.test.js
```diff
@@ -175,4 +175,14 @@ describe('d2l-insights-engagement-dashboard', () => {
 			});
 		});
 	});
+
+	describe('events', () => {
+		it('Should open the settings page on an event', async() => {
```
Vitalii-Misechko, line 180: nice!

Vitalii-Misechko (approved):
LGTM!

bemailloux (approved):
lgtm!

components/engagement-dashboard-errors.js
NicholasHallman: Ohh I see, not that the event goes to settings but that clicking the button calls the correct event

PR #616, opened 2021-08-17
Update summary card padding size 

JikaiLong:
Updated summary card padding size according to the new design:
Before:
image

After: 
image

Skeleton When onloading:
image


Vitalii-Misechko:
@JikaiLong could you provide more details on this design? 

- [ ] I mean why are these changes needed in the first place?

- [ ] Have you a chance to verify responsive design?

- [ ]  Have you verified the skeleton view of these cards?

JikaiLong:
> @JikaiLong could you provide more details on this design?
> 
> * [x]  I mean why are these changes needed in the first place?
> * [x]  Have you a chance to verify responsive design?
> * [x]   Have you verified the skeleton view of these cards?

According to Kevin: The update is to match the standard padding for cards in Daylight.
No responsive change.
skeleton view works as before.

Vitalii-Misechko (approved):
LGTM! Thank you for your answer 👍

Vitalii-Misechko:
@JikaiLong Have you noticed that _Discussion Activity_ card has a header with different padding from the left edge?

JikaiLong:
> @JikaiLong Have you noticed that _Discussion Activity_ card has a header with different padding from the left edge?
> 
> * [ ]  Discussion Activity header padding

I think the Discussion Activity is using a different template for charts, so the whole design approach could be different?

Vitalii-Misechko:
> > @JikaiLong Have you noticed that _Discussion Activity_ card has a header with different padding from the left edge?
> > 
> > * [ ]   Discussion Activity header padding
> 
> I think the Discussion Activity is using a different template for charts, 

That's correct

> so the whole design approach could be different?

I think it's different. I believe that we can just adjust `d2l-insights-discussion-activity-card-title` style and we're good

NicholasHallman:
This may have been merged prematurely. The vis-diff doesn't show Vitalli's suggested change

NicholasHallman:
@JikaiLong Why didn't the visual diff for this change get merged?

PR #624, opened 2021-08-31
US129325:  Immersive Nav for Learner Dashboard should say "Learner" instead of "User"

opets:
screens:
image
image
FYI @Brightspace/draco-vis


components/immersive-nav.js
```diff
@@ -70,7 +70,7 @@ class InsightsImmersiveNav extends Localizer(MobxLitElement) {
 			case 'home':
 				return this.localize('dashboard:backToInsightsPortal');
 			case 'user':
-				if (this.viewState.isSingleLearner) return this.localize('dashboard:backToUserSelector');
+				if (this.viewState.isSingleLearner) return this.localize('dashboard:backToLearnerSelector');
```
anhill-D2L, line 73: Will this default everyone into seeing English text?

If so, we might want to add the `backToLearnerSelector` lang term, wait for it to be translated, then start showing `backToLearnerSelector` instead of `backToUserSelector`. (This sounds like a Joe question)

locales/en.js
```diff
@@ -5,7 +5,7 @@ export default {
 	"dashboard:userView:title": "Learner Engagement Dashboard",
 	"dashboard:backToInsightsPortal": "Back to Insights Portal",
 	"dashboard:backToEngagementDashboard": "Back to Engagement Dashboard",
-	"dashboard:backToUserSelector": "Select a Different User",
+	"dashboard:backToLearnerSelector": "Select a Different Learner",
```
anhill-D2L, line 8: Alternative option would be to keep the same term, but replace User with Learner, this way people will continue to see a translated term, and it will be automatically updated when translations come in

ex. 
	"dashboard:backToUserSelector": "Select a Different Learner",

opets: I worry that lang-term will remain unchanged in this case

components/immersive-nav.js
opets: let's  do something like this:
	if(this.localize('dashboard:backToLearnerSelector'))
		return this.localize('dashboard:backToLearnerSelector');
	else
		return this.localize('dashboard:backToUserSelector');

bemailloux: @anhill-D2L If this change is deployed correctly this will not result in users seeing English text. Therefore if-statement is not required

locales/en.js
bemailloux: @opets Serge.io is able to determine if a langterm changed, so modifying an existing langterm shouldn't be a problem. Therefore I would follow @anhill-D2L's suggestion here

components/immersive-nav.js
anhill-D2L: @bemailloux Is that because we currently have a release that's being used, and showing the correct language, and the next release would then ideally include this change along with the related translations?

bemailloux: Yes. That depends on a few things (we release correctly, serge actually picks up the translation, we're able to certfix on time) but that's the normal process

locales/en.js
opets: ok, fixed

bemailloux (approved):

anhill-D2L (approved):

components/immersive-nav.js
bemailloux: I added a reminder to slack so we don't forget to release - we have forgotten before :P

PR #652, opened 2021-11-29
Update package.json

BhamelinD2L:
Removes the unused multi-select dependency.
Tested via demo pages as well as the tests to ensure there are no errors as a result of removing this dependency.


NicholasHallman (approved):

bemailloux (approved):
Demo is working even after the dep was removed. LGTM!

image

PR #658, opened 2021-12-14
switching to ESM-by-default & upgrading actions

dlockhart:
This accomplishes two things that both needed to happen together:

### 1) New Visual Diff dependencies

The main motivation for this change. By switching to the new `main` branch of the actions repo, you pick up the latest visual-diff, mocha@9 and puppeteer@12 -- this will likely case minor visual differences.

### 2) Make this project use ES Modules by default

Mocha 9 (required for the new visual-diff action) requires it, and web projects shouldn't be CommonJS anyway. I've converted the things using CommonJS things (like `require()` and `__dirname`) to their ESM equivalents.

### Other minor updates

While I was in here, I updated a few things like the latest dev-server and test-runner and linting configs.


demo/index.html
```diff
@@ -2,7 +2,6 @@
 <html lang="en" data-timezone='{"name":"Canada - Toronto","identifier":"America/Chicago"}'>
 	<head>
 		<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
-		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
```
dlockhart, line 5: Polyfill no longer needed now that legacy-Edge is gone!

test/engagement-dashboard.test.js
```diff
@@ -182,7 +182,7 @@ describe('d2l-insights-engagement-dashboard', () => {
 			const errorAlert = el.shadowRoot.querySelector('d2l-insights-engagement-dashboard-errors');
 			errorAlert.dispatchEvent(new Event('d2l-insights-go-to-settings'));
 			await el.updateComplete;
-			expect(window.location.search.includes('?v=settings')).to.be.true;
+			expect(new URLSearchParams(document.location.search).get('v')).to.equal('settings,0,');
```
dlockhart, line 185: The new test runner has other stuff on the query-string so it was `&settings` instead of `?settings`. I switched it to use the `UrlSearchParams` to future-proof it a bit more.

NicholasHallman:
Looks great! I'll keep an eye on this incase the tests fail in unexpected ways

NicholasHallman:
> While I was in here, I updated a few things like the latest dev-server and test-runner and linting configs.
Nice! Thank you

NicholasHallman (approved):

PR #669, opened 2022-01-28
US135597 [Engagement] Lit 2 migration - Phase 1

Vitalii-Misechko:
[US135597](https://rally1.rallydev.com/#/?detail=/userstory/623256115137&fdp=true): [Engagement] Lit 2 migration - Phase 1

This PR makes backwards-compatible changes to prepare all repos for a Lit 2 upgrade. 

**Non-breaking change checklist:** 

- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
  - [x] double-check by vetting `package-lock` file
- [x] audit `ifDefined` usages for null parameters

FYI @Brightspace/draco-vis


components/dashboard-settings.js
```diff
@@ -347,10 +347,14 @@ class DashboardSettings extends RtlMixin(Localizer(LitElement)) {
 	}
 
 	get _selectedRoleIds() {
+		if (!this.shadowRoot) return [];
```
NicholasHallman, line 350: Don't need this check since it's called by an event triggered by the UI. shadowRoot has to exist at this point

components/role-list.js
```diff
@@ -105,6 +105,8 @@ class RoleList extends Localizer(LitElement) {
 	}
 
 	async firstUpdated() {
+		if (!this.shadowRoot) return;
```
NicholasHallman, line 108: called after connected, shadowRoot exists

components/role-list.js
```diff
@@ -115,6 +117,8 @@ class RoleList extends Localizer(LitElement) {
 	}
 
 	async updated() {
+		if (!this.shadowRoot) return;
```
NicholasHallman, line 120: same here

components/dashboard-settings.js
Vitalii-Misechko: I agree with you. However, I'd like you to consider the following idea: If you see a check for `shadowRoot` then you know that this method/property was reviewed by a developer. This pattern is easy to review and follow

NicholasHallman:
I think we might want to decide if we are going to add shadow root checks everywhere to be safe or just where we suspect somebody could try and call a function before the connectedCallback has been run. I personally not opposed to shadowRoot checks but I think 99% of these won't ever be undefined.

IMO handle functions and lifecycle functions are all safe. Anything that another component might call could be unsafe due to construction order, but even this is unlikely since an element calling a child's functions in a lifecycle guarentees that the child is connected (but maybe not rendered)

components/dashboard-settings.js
Vitalii-Misechko: An additional benefit to the idea above is that a code can be easily moved around as soon as guard condition is kept

NicholasHallman: That's a fair point, I am cool with the checks then 👍

components/content-view-histogram.js
```diff
@@ -1,5 +1,5 @@
 import { computed, decorate, observable } from 'mobx';
-import { css, html } from 'lit-element/lit-element.js';
+import { css, html } from 'lit-element';
```
annabelsegalld2l, line 2: The import path `'lit-element/lit-element.js'` should actually be okay. `"lit-element"` works as well, just wanted to let you know that it's not a requirement to change.

annabelsegalld2l (approved):
LGTM! Thanks for jumping on this Vitalii 😄

NicholasHallman (approved):

PR #675, opened 2022-02-08
Debounce Semester Selector

NicholasHallman:
Implementation Notes

Create a declarative debounce event mixin that we can use in components in other dashoards that may suffer from this same problem.

Functional
 - Debounce
   - [x] Calling the event multiple times doesn't trigger the handler
   - [x] Calling the event with a delay multiple times does trigger the handler
 - Semester Filter
   - [ ] The debounce mixin causes the semester to delay calling the api when truncated

UX
 - [ ] demo

Documentation - N/A


test/model/debounce.test.js
```diff
@@ -0,0 +1,78 @@
...
+		const el = await fixture(html`<debounce-test
+			@d2l-click-debounce="${_handleClick}"
+		></debounce-test>`);
+		await el.updateComplete;
+
+		el.dispatchEvent(new Event('d2l-click-debounce'));
+		el.dispatchEvent(new Event('d2l-click-debounce'));
+		el.dispatchEvent(new Event('d2l-click-debounce'));
+
+		await new Promise((res) => setTimeout(res, 300));
```
Vitalii-Misechko, line 42: `nit` can be replaced with `await aTimeout(300)`

model/debounce.js
```diff
@@ -0,0 +1,39 @@
+export const EventDebounceMixin = (superclass) => class extends superclass {
```
Vitalii-Misechko, line 1: neat!

Vitalii-Misechko (approved):
LGTM!

anhill-D2L:
Is debouncing basically to prevent/minimize the issue or impact of someone spam clicking buttons in our dashboard?

components/semester-filter.js
```diff
@@ -26,6 +26,12 @@ class SemesterFilter extends Localizer(LitElement) {
 		};
 	}
 
+	static get debouncedEvents() {
+		return {
+			'd2l-insights-semester-filter-change': 500
```
anhill-D2L, line 31: What is the 500?

model/debounce.js
```diff
@@ -0,0 +1,39 @@
...
+		const entry = Object.entries(options).find(([key]) => key === event.type);
+		if (entry === undefined) {
+			this.originalDispatch(...args);
+			return;
+		}
+
+		const delay = entry[1];
+
+		if (this.activeDelays[event.type] !== undefined) {
+			clearTimeout(this.activeDelays[event.type]);
```
anhill-D2L, line 31: If I send an event every `delay-1` for a minute would anything actually happen?

test/model/debounce.test.js
```diff
@@ -0,0 +1,78 @@
...
+		let count = 0;
+		const _handleClick = () => count += 1;
+
+		const el = await fixture(html`<debounce-test
+			@d2l-click-debounce="${_handleClick}"
+		></debounce-test>`);
+		await el.updateComplete;
+
+		el.dispatchEvent(new Event('d2l-click-debounce'));
+		el.dispatchEvent(new Event('d2l-click-debounce'));
```
anhill-D2L, line 39: It might be nice having small timeout here just so I know that this works even if events don't happen at basically the same time

components/semester-filter.js
NicholasHallman: it's the delay time in milliseconds

model/debounce.js
NicholasHallman: No, it would never fire the event if you kept requesting for an event within the debounce time

NicholasHallman:
> Is debouncing basically to prevent/minimize the issue or impact of someone spam clicking buttons in our dashboard?

Yes, there is a bug on engagement dashboard where 
 - clicking on a semmester fires off a new api call
 - If you want to select multiple semmesters it fires off multiple calls
 - If one of those calls returns before you are done making selections then it will change the semmester selected list and turn off some checks
 
 This alleviates the problem a little bit by allowing you to make multiple selection in one api call.

anhill-D2L (approved):

PR #678, opened 2022-02-15
Semester Selector Race Conditions Fix

NicholasHallman:
Implementation notes

There were a couple of small bugs working together to create a bad experience

- isLoading would be false when loading was taking place  

We weren't keeping track of multiple api calls inflight so when the first call returned the skeletons would turn off, showing the data even though we were waiting for the second call.

- If data returned out of order we didn't prioritise the last call

Same as above. Keeping track of calls means we can select the last call made and update the data to its result.

- semester preSelected updated component from returned data  

The semester filter dropdown would take new lists from the api everytime it returned. Changed it to only initialize on connectedcallback when the initial data and semester list arrives. Also changed the parent component to track it's own semester list instead of relying on the child components though the shadow dom (which introduced a race condition because lit-element doesn't update components immediately)

MRp8DHKhdh


model/data.js
```diff
@@ -134,11 +138,18 @@ export class Data {
...
 
-		this.isLoading = false;
-		this._serverData = newServerData;
-		this._serverDataProxy++;
-		this._selectorFilters.semester.selected = this.serverData.selectedSemestersIds || [];
-		this._selectorFilters.user.selected = this.serverData.selectedUserId;
+		// only save the data if the last loading promise completes
+		const currentSize = this.loadingPromises.length;
+		this.loadingPromises[currentSize - 1].then(() => {
+			if (this.loadingPromises.length !== currentSize && this.loadingPromises.length !== 0) return;
```
johngwilkinson, line 144: Why the ` !== 0` check? Looks like if promise n+1 completes while promise n is waiting, the n+1 handler will zero out `loadingPromises`, causing promise n to enter this block and lose the n+1 state.

NicholasHallman: Because of the debouncing sometimes you get a situation where

click semester n - 1
call n - 1 is sent
click semester n
call n-1 is returned
call n is sent
call n is returned. length of array is 0. doesn't match

components/semester-filter.js
```diff
@@ -84,11 +81,20 @@ class SemesterFilter extends EventDebounceMixin(Localizer(LitElement)) {
...
 		 * @event d2l-insights-semester-filter-change
 		 */
-		this.dispatchEvent(new Event('d2l-insights-semester-filter-change'));
+		e.detail.itemId = Number(e.detail.itemId);
+		if (e.detail.selected) {
+			this.selected.push(e.detail.itemId);
+		} else {
+			const index = this.selected.indexOf(e.detail.itemId);
+			if (index > -1) {
+				this.selected.splice(index, 1);
```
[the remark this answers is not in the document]
johngwilkinson: Please? :)

model/data.js
johngwilkinson: I'm having trouble understanding the situation. If n doesn't get added to loadingPromises until after n-1 enters the `then` block, then n will be in the array when n completes. Can you elaborate on where the executing thread gets passed off each time?

I'm just trying to figure out a way to avoid both this situation and the situation I described above.

components/semester-filter.js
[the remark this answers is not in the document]
NicholasHallman: Sorry I have it coming hahaha

model/data.js
NicholasHallman: I think I see the problem. typing it out made it make sense

model/data.js
```diff
@@ -139,16 +139,16 @@ export class Data {
 		});
 
 		// only save the data if the last loading promise completes
-		const currentSize = this.loadingPromises.length;
-		this.loadingPromises[currentSize - 1].then(() => {
-			if (this.loadingPromises.length !== currentSize && this.loadingPromises.length !== 0) return;
+		const lastPromise = this.loadingPromises[this.loadingPromises.length - 1];
+		lastPromise.then(data => {
+			if (data !== newServerData) return;
```
johngwilkinson, line 144: This seems elegant. I do see two cases above where it might not be true (when catching an error and when `data.selectedRolesIds === null`). Not sure if that's a problem here or not.

johngwilkinson (approved):

components/semester-filter.js
```diff
@@ -84,11 +80,17 @@ class SemesterFilter extends EventDebounceMixin(Localizer(LitElement)) {
 		}));
 	}
 
-	_updateFilterSelections() {
+	async _updateFilterSelections(e) {
```
Vitalii-Misechko, line 83: `curious` why did this function become async?

Vitalii-Misechko (commented):
1. Have you considered debouncing before `recordProvider/fetchData` as an additional component to the current solution to avoid unnecessary API calls?
2. Have you considered debouncing before `loadData` as an alternative solution?
2. I guess we need a similar solution in each dashboard, right?

model/data.js
```diff
@@ -95,7 +96,10 @@ export class Data {
 			defaultView
 		};
 		try {
-			const data = await this.recordProvider(filters, this._metronEndpoint);
+			const promise = this.recordProvider(filters, this._metronEndpoint);
+			this.loadingPromises.push(promise);
+			const data = await promise;
+
```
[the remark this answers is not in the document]
NicholasHallman: I'll try this, not sure if it will work or not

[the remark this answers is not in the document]
NicholasHallman: It works, replacing with this simpler version. Good call!

components/semester-filter.js
NicholasHallman: Before I made the semester filter own the data structure that decided if checks were clicked or not I had to wait for the checks to update so I could read their new state. But I'm not doing that anymore so I'll remove the async

model/data.js
NicholasHallman: Vitalii's suggestion fixes these concerns so I'm resolving it

NicholasHallman:
Loaddata debouncing ends up being the best way to go here. it also fixes the orgunit filter which is a huge plus

NicholasHallman:
> I guess we need a similar solution in each dashboard, right?
Yep, I bet this exists on assessment, fortunatly now that we understand the problem it's a pretty easy fix

johngwilkinson:
General thought on testing, maybe not needed: I notice you're using the demo page, at least in the demo. Real LMSes are so much slower that you might hit different issues, so worth testing that way I think.

model/data.js
```diff
@@ -77,6 +78,21 @@ export class Data {
 		};
 
 		this._ouFilterDataManager = new EngagementOuFilterDataManager(this);
+
+		this.loadDataReal = this.loadData;
+		this.debounceTimeout;
+		this.loadData = new Proxy(this.loadData, {
```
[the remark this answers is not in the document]
NicholasHallman: I had copied the proxy code from the debounce mixin as a quick test, I'll switch it to a method now

johngwilkinson (approved):

Vitalii-Misechko (approved):
LGTM!

NicholasHallman:
Tested on draco lms, tricked it into "truncating" by setting the flag to true always in data for testing. Felt good there and yes it was much slower but loading persists properly and debouncing worked well

PR #686, opened 2022-02-23
fix a few recent bugs

johngwilkinson:
This fixes 3 bugs:

1. ou filter not filtered by semester filter (it should be)
2. on default view load, semesters are selected but semester filter shows "All", i.e. none appear selected
3. clear button on semester filter doesn't work (throws NPE)

Tested e2e on draco stack, but I would appreciate if reviewers would pull this branch and do their own testing too.


components/semester-filter.js
```diff
@@ -77,11 +74,16 @@ class SemesterFilter extends Localizer(LitElement) {
 		/**
 		 * @event d2l-insights-semester-filter-change
 		 */
-		e.detail.itemId = Number(e.detail.itemId);
-		if (e.detail.selected) {
-			this.selected.add(e.detail.itemId);
+		if (e.detail) {
```
anhill-D2L, line 77: Should this be ( I don't know if e could be null)
`if (e && e.detail) {`

bemailloux (commented):
Code looks LGTM. Running through some manual tests. Any thoughts on how we might be able to write integration-level tests for these dashboards?

I used to do that kind of stuff with Selenium, but that was a huge pain / time-sink.

anhill-D2L:
> Code looks LGTM. Running through some manual tests. Any thoughts on how we might be able to write integration-level tests for these dashboards?
> 
> I used to do that kind of stuff with Selenium, but that was a huge pain / time-sink.

Can visual diffs cover some of this stuff?

bemailloux (approved):
Ran through some quick tests (i.e. checking that semester filter debouncing still worked). I noticed that changing the semester filter, then quickly changing some selections in OU filter, would revert the OU filter to its previous state after it was done loading. Not sure if that's in scope or not?

engagement-dashboard

johngwilkinson:
Yeah, I did notice, too, that some race conditions between the filters could still happen, although not so easily as before. Its' still an improvement, so I'm not worried about shipping like that.

components/semester-filter.js
johngwilkinson: I don't think e can be null, although your way is probably a good idea in general.

anhill-D2L (approved):

components/semester-filter.js
```diff
@@ -34,17 +34,14 @@ class SemesterFilter extends Localizer(LitElement) {
...
-	}
-
 	async firstUpdated() {
 		this.dataProvider = this.isDemo ? fetchDemoSemesters : fetchSemesters;
 		await this._loadData();
 	}
 
 	render() {
+		// set on every render as it may be updated on each server load
+		this.selected = new Set(this.preSelected);
```
NicholasHallman, line 44: This creates a regression. Now when truncation is on and you change the semester filter the first value won't toggle. Changes made to the semester selector and data were intended to make it so that the semester filter was always in sync with whatever the last thing to come back from data was so after pre selecting it wasn't necesary. 

Connected callback check may have been too early for the data to arrive so maybe instead it should be here in render but only happen once to initialize the list

johngwilkinson: I don't fully understand, but is this something you could fix today? I believe we can still do a cert bypass.

PR #688, opened 2022-02-24
Fix Pre-Select Semester Render Problem

NicholasHallman:
Functional 
- Fix the All Selected filter message when the dashboard loads for the first time.
- Fix the deselect semester bug
- Properly initialize the selected list on first load


components/semester-filter.js
```diff
@@ -21,7 +21,8 @@ class SemesterFilter extends Localizer(LitElement) {
 			preSelected: { type: Array, attribute: false },
 			_filterData: { type: Array, attribute: false },
 			_bookmark: { type: String, attribute: false },
-			_search: { types: String, attribute: false }
+			_search: { type: String, attribute: false },
+			isLoading: { type: Boolean, attribute: 'is-loading' }
```
johngwilkinson, line 25: attribute name should be `loading`

engagement-dashboard.js
```diff
@@ -346,6 +346,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 		<d2l-insights-semester-filter
 			page-size="10000"
 			?demo="${this.isDemo}"
+			?is-default-view="${this._isLoading}"
```
johngwilkinson, line 349: shouldn't this be `?loading="${this._isLoading}"`?

johngwilkinson (commented):
Did you test on draco or the demo page?

NicholasHallman:
> Did you test on draco or the demo page?
Did a test on draco stack, still had the issue. latest commit solves the problem

components/semester-filter.js
```diff
@@ -77,15 +77,21 @@ class SemesterFilter extends Localizer(LitElement) {
...
+				const setSelected = new Set(this.selected);
+				setSelected.add(e.detail.itemId);
+				this.selected.clear();
+				[...setSelected].forEach(s => this.selected.push(s));
 			} else {
-				this.selected.delete(e.detail.itemId);
+				const setSelected = new Set(this.selected);
+				setSelected.delete(e.detail.itemId);
+				this.selected.clear();
+				[...setSelected].forEach(s => this.selected.push(s));
```
NicholasHallman, line 88: This code modifies the data.serverData.selectedSemesterIds in place so as not to replace the reference with a new array

engagement-dashboard.js
```diff
@@ -687,7 +688,7 @@ class EngagementDashboard extends Localizer(MobxLitElement) {
 	_semesterFilterChange(event) {
 		event.stopPropagation();
 		filterEventQueue.add(this.localize('alert:updatedFilter', { chartName: this.localize('semesterFilter:name') }));
-		this._serverData.selectedSemesterIds = event.detail;
+		this._serverData.selectedSemesterIds = [...this._serverData.selectedSemesterIds];
```
NicholasHallman, line 691: the setter calls reload so once the semester filter is done modifying the array we set it here to trigger the api call

johngwilkinson (approved):
Have tested, looks good!

(Found an unrelated, pre-existing bug where urlState changes won't cause data to load even if neeed)

components/semester-filter.js
```diff
@@ -77,15 +77,21 @@ class SemesterFilter extends Localizer(LitElement) {
...
+				const setSelected = new Set(this.selected);
+				setSelected.add(e.detail.itemId);
+				this.selected.splice(0);
+				[...setSelected].forEach(s => this.selected.push(s));
 			} else {
-				this.selected.delete(e.detail.itemId);
+				const setSelected = new Set(this.selected);
+				setSelected.delete(e.detail.itemId);
+				this.selected.splice(0);
+				[...setSelected].forEach(s => this.selected.push(s));
```
bemailloux, line 88: oh interesting - is this modifying all the elements in the array while avoiding reassignment (for mobx reasons I guess)?

bemailloux (approved):
LGTM!

components/semester-filter.js
NicholasHallman: Yep!

NicholasHallman: Also, do you mind pulling the branch and trying this on draco stack? John and I would like one more eyeball to make sure there isn't something we missed

bemailloux: Ok, I've completed my testing and I think we're good to go. One edge-case bug gave me the run around for a bit, but it requires a semester with no children to be selected as default, plus org unit truncation, for the bug to manifest. That's probably not an uncommon occurrence but it's also not a blocker for this PR imo.

(ref: https://d2l.slack.com/archives/C031HN77DSA/p1645718565088819)

johngwilkinson:
Sauce failed. I'll rerun it

johngwilkinson:
> Sauce failed. I'll rerun it

@NicholasHallman in the meantime, you might want to check if the tests pass locally for you - it occurs to me these changes could potentially break a test or two

johngwilkinson (approved):

PR #710, opened 2022-03-18
Ported from Lit2 phase 2 Fix. Error alert appears for a second even there are not errors

Vitalii-Misechko:
I noticed that this bug is fixed in Lit 2 Phase 2. I think we can release it earlier

* Functional Testing 
  * [x] Lit 2 Phase 1 compliant changes
  * [x] E2ETest cases
    * [x] Settings doesn't flash error message if there is no error
* Security, devops, perf: N/A


NicholasHallman (approved):
Good catch thanks for fixing this :)

bemailloux (approved):

PR #716, opened 2022-03-23
add lit as a dependency (needed for lit-mobx)

annabelsegalld2l:
lit-mobx has a peer dependency on the `"lit"` package specifically. Lit is a wrapper of lit-element and lit-html, so no code is being added. We can remove lit-element and lit-html imports once the paths are changed to reference `lit/` paths directly.


test/components/default-view-popup.test.js
```diff
@@ -87,8 +87,10 @@ describe('d2l-insights-default-view-popup', () => {
 
 		describe('render', () => {
 			it('should show the message about recently created N semesters', () => {
-				const descs = Array.from(el.shadowRoot.querySelectorAll('p.d2l-insights-default-view-popup-description'));
-				expect(descs.some(p => p.innerText.trim().endsWith('recently created 2 semesters.'))).to.be.true;
+				requestAnimationFrame(() => {
```
annabelsegalld2l, line 90: I noticed this test was also flaky on master. Wrapping the expect in a requestAnimationFrame ensures the content can fully render before we look at the inner descriptions.

Vitalii-Misechko (approved):

test/components/default-view-popup.test.js
Vitalii-Misechko: We use `aTimeout` for the same reason. Is using `requestAnimationFrame` a better approach?

annabelsegalld2l: I think either will work here. On my team we prefer the `requestAnimationFrame` because we know that we only need to wait for the next repaint. `aTimeout` is also harder to debug in the future because the timeout value can be configured.

johngwilkinson (approved):

PR #724, opened 2022-05-06
Try to fix visual-diff flake

svanherk:
Try to get the visual-diff tests a bit more stable.  Will merge the diffs then rerun a few times and see how it goes.


test/visual-diff/engagement-dashboard-skeleton.visual-diff.js
```diff
@@ -32,7 +32,7 @@ describe('engagement-dashboard-skeleton', () => {
 
 	it('Desktop', async function() {
 		const rect = await visualDiff.getRect(page, 'd2l-insights-engagement-dashboard');
-		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false, clip: rect });
```
svanherk, line 35: We've found this greatly reduces flake if your code reacts to window `resize` events.  Otherwise, the `puppeteer` `screenshot` call itself will trigger a resize.  Because we're no longer capturing beyond the viewport, I had to adjust some sizes below.

svanherk:
Ok this is ready for review.  Re-ran numerous times and the visual-diffs seem really stable.  The unit tests seem better, but still have the occasional failure.

After this merges I'll merge it into the filter PR.

johngwilkinson (approved):

PR #720, opened 2022-05-05
US134782 - Use new filter from BrightspaceUI/core

svanherk:
This replaces the Labs filter with the new one from `core`.  The new version is much more locked down - I'll call out removed functionality in comments below.  I'm also going to send an update to the existing email chain with Megan and designers to highlight the differences between the new filter and the org filter.

image

PR #726, opened 2022-05-06
Use the core filter directly in the semester filter

svanherk:
This PR is just an option... after moving to the new filter in https://github.com/Brightspace/insights-engagement-dashboard/pull/720, the `dropdown-filter` becomes pretty light because a lot of the logic is handled by the new filter or is no longer supported.  This PR removed it and just uses the new fitler directly in the semester filter, and combines test files.

If you'd rather keep them separate, you can just close this PR!


johngwilkinson:
Closing in favour of #720

